PR #19, opened 2026-05-08
fix: resolve jiti from @earendil-works/pi-coding-agent

anh-chu:
The resolveJitiImport function only looked for `@mariozechner/jiti` and `@oh-my-pi/jiti` package names. `@earendil-works/pi-coding-agent` ships jiti as the unscoped `jiti` package, causing the dashboard server to fail with "Cannot find pi's TypeScript loader (jiti)."\n\nAdded `"jiti"` as the first lookup entry so the dashboard server works with all pi distributions.


robertcsakany (commented):
I've already fixed this with a larger full update to 0.7.4

PR #49, opened 2026-05-28
fix: remove -i flag from whichViaLoginShell to prevent SIGTSTP

tpatzig:
The whichViaLoginShell() fallback spawns a shell to locate binaries not found on PATH. Using `zsh -ilc` (interactive + login) causes the shell to call tcsetpgrp() and claim the terminal's foreground process group. When the shell exits, the parent pi process is no longer in the foreground group, so the terminal driver delivers SIGTSTP — stopping the pi TUI process immediately after startup.

This manifests as `[1]+ Stopped pi` in iTerm2 (and likely other terminals) on macOS whenever the tool registry resolves a binary that isn't on PATH (e.g. `jj` when not installed), triggering the login shell fallback.

Fix: use `-lc` (login only, non-interactive). A login shell still sources ~/.zprofile for PATH resolution but does not attempt to manage the terminal's process group.


robertcsakany:
Thank for the PR, I will validate with our current design - there are other spawn canges there. If it can fit, I will merge it.

PR #71, opened 2026-06-02
feat(client): relocate per-session View menu into the composer status bar

robertcsakany:
## Summary

Moves the per-session display-preferences popover (`ChatViewMenu` — the **⚙ View** button) out of its dedicated full-width toolbar row at the top of `ChatView` and into the bottom **StatusBar**, in the `leading` cluster between the refresh button (⟳) and the `ModelSelector`. The wasted horizontal band above the first chat message is reclaimed.

Bottom bar now reads: `⟳ · ⚙ View │ [model ▾] [thinking ▾] │ <session actions>`

## What changed

- **`App.tsx`** — mount `ChatViewMenu` in the `StatusBar` `leading` slot (gated on `selectedId`); drop `displayPrefsOverride`/`onSetDisplayPrefs` from the `<ChatView>` mount.
- **`ChatView.tsx`** — delete the standalone toolbar-row `<div>`; remove the now-orphaned props, the `ChatViewMenu` import, and unused display-prefs type imports.
- **`StatusBar.test.tsx`** — new test asserting `ChatViewMenu` renders inside `status-bar`, after the refresh button and before the model selector (DOM order via `compareDocumentPosition`).
- **Spec** — add requirement *"Display-preferences menu SHALL mount in the composer status bar"* (+3 scenarios) to `chat-display-preferences`; archive the OpenSpec change.

No change to display-pref semantics, the WS protocol, persistence, or the popover internals — only the trigger button's location moves. Mobile is covered automatically (StatusBar renders on the mobile composer path too).

## Verification

- `npm test` — StatusBar suite 12/12 pass; full run 7102 passed (the single failure is an unrelated `pi-image-fit` JPEG timeout).
- `npm run build` clean; verified live in a browser on a dedicated `:8300` server: **⚙ View** renders in the status bar between refresh and model selector, and no standalone toolbar row remains at the top of chat.

Closes the OpenSpec change `relocate-view-menu-to-status-bar`.


robertcsakany:
## CodeRabbit Autofix — 1 fixed, 1 skipped

**Fixed (1):**
- 🧹 *Add an integration check for the actual relocation* — strengthened tests so the relocation is actually guarded:
  - `StatusBar.test.tsx`: assert exactly **one** `View options` trigger exists and lives in `status-bar`.
  - `ChatView.test.tsx`: assert `ChatView` renders **no** `View options` trigger (catches a duplicate `ChatViewMenu` re-appearing).

**Skipped (1) — false positive:**
- 🟠 *OpenSpec artifact in disallowed nested path* — the cited rule applies to **creating new** changes, not archiving completed ones. The repo already has 385 archived changes under `openspec/changes/archive/YYYY-MM-DD-<name>/`; that is the canonical location used by the `openspec-archive-change` flow. Moving it back to `openspec/changes/<name>/` would un-archive it. No change made.

**Commit:** `31262872` on `os/relocate-view-menu-to-status-bar`.

PR #104, opened 2026-06-13
fix(extension): resolve global prompt templates from dashboard + bridge mDNS opt-out

robertcsakany:
## Summary

Finishes the leftover Decision 3 + Issue 3 from the archived `fix-slash-dispatch-delivery` change, plus a bridge-side mDNS opt-out discovered while verifying it.

### Changes
- **`prompt-expander.ts`** — `resolveTemplate` Step 3 adds a parallel `source:"prompt"` probe so global/project/package prompt templates (e.g. `/session-summary` at `~/.pi/agent/prompts/`) expand from the dashboard instead of going to the LLM as raw slash text.
  - **Path-field fix:** real pi (0.78) `getCommands()` returns the on-disk path under `sourceInfo.path` (synthetic `SourceInfo`), **not** a top-level `c.path`. The probe now reads `c.sourceInfo?.path ?? c.path`. The prior `c.path`-only check matched `undefined` at runtime and silently fell through — it only "passed" because unit tests mocked the wrong shape. Tests updated to the real `sourceInfo` shape.
- **`bridge-context.ts`** — `hasDispatchCommand` gains an `in`-operator fallback with a guarded `typeof` for getter-backed / Proxy-hidden `dispatchCommand`. Defensive; no current pi ships `dispatchCommand`.
- **`server-auto-start.ts`** — adds a `PI_DASHBOARD_NO_MDNS` opt-out to the bridge's `autoStartServer` (mirrors the server's existing gate). Without it the bridge always runs mDNS discovery and overrides an explicit `PI_DASHBOARD_URL` via `connection.updateUrl()`, hijacking isolated sessions onto a co-located real dashboard. Required for isolated/CI runs.
- **OpenSpec** — change archived; `command-routing` spec synced (ADDED: global prompt template resolution; MODIFIED: `hasDispatchCommand` detection). Also repairs a pre-existing stray `## ADDED Requirements` header in the main spec that had hidden all requirements from the parser.

### Testing
- 59 affected unit tests pass (`prompt-expander`, `server-auto-start`, `extension-slash-command-detection`).
- `tsc --noEmit` clean for changed files.
- Full suite: 7488 pass; the only failures are pre-existing/environmental (`image-fit` jimp, `doctor-route` timing) — no server code touched here.

### Caveats / follow-ups
- **Live E2E (task 3.3) was not confirmed** — the isolated-env browser send was blocked by harness instability (interactive pi dying in detached tmux, headless bridge WS churn). The fix is correct per pi's source + unit tests, but a live `/session-summary` expansion was not observed end-to-end.
- The `PI_DASHBOARD_NO_MDNS` bridge gate is a separate server-discovery concern bundled here without its own spec; could be split into its own change later.
- Bridge changes require `npm run reload` to take effect in live sessions.


robertcsakany:
## CodeRabbit review — resolution (commit a171308c)

**Applied:**
- ✅ **NO_MDNS env edge cases** (`server-auto-start.test.ts`) — added coverage for `"yes"`, `" TRUE "` (trim+lowercase), and a falsy `"0"` that keeps mDNS running.
- ✅ **`getCommands()` hardening** (`prompt-expander.ts`) — guard the result is an array and `cmdPath()` returns a non-empty string before `existsSync`, so a malformed registry entry can't throw and shadow a resolvable sibling. Tests added for malformed/non-array returns.

**Intentionally not applied:**
- ❌ **"Move artifacts out of `openspec/changes/archive/...`"** — false positive. `openspec/changes/archive/YYYY-MM-DD-<name>/` is the repo's *documented* archive location (`AGENTS.md`), created by the `openspec archive` CLI. Moving these back to `changes/<name>/` would un-archive a completed change. The "never nest under archive/" guideline applies to *authoring new active changes*, not archived ones.
- ❌ **Externalize mDNS timeouts (2000 / 10000) to config** — out of scope and conflicts with the repo's own *Simplicity First* rule (`AGENTS.md`: "No configurability that wasn't requested"). These are pre-existing internal values; this PR only wrapped them in the `if (!noMdns)` gate. No one has needed to tune them. Happy to revisit in a dedicated change if a real tuning need appears.

PR #107, opened 2026-06-13
fix(editor): persist VS Code state + dedup concurrent code-server spawns

robertcsakany:
## Summary

Folder editors (code-server) now restore their full state across dashboard restarts and no longer spawn duplicate/stalled instances when opened concurrently.

Implements OpenSpec change **fix-editor-settings-persistence** (archived).

### 1. Persist VS Code state (settings seeding)
`writeVscodeThemeSettings` → `writeVscodeUserSettings` in `editor-manager.ts`. Seeds the per-editor `settings.json` with persistence keys, merged so existing user values win and theme keys stay authoritative:
- `window.restoreWindows: all`
- `workbench.editor.restoreViewState: true`
- `files.hotExit: onExitAndWindowClose`
- `security.workspace.trust.enabled: false`
- `update.mode: none`
- `extensions.autoCheckUpdates: false`
- `workbench.startupEditor: none`

Result: open tabs/layout restore, no Workspace Trust dialog, no welcome tab, no update banners.

### 2. Server-side concurrency dedup
`editor-manager.start(cwd)` previously checked its dedup map only *after* two `await` points, so concurrent opens of the same folder could spawn two code-servers on one locked `--user-data-dir`, leaving a stalled/orphaned instance. Now `start` memoizes an in-flight Promise per cwd (`inFlightStarts`) and clears it on settle — exactly one spawn per folder.

### 3. Client hardening (defense-in-depth)
`EditorView.tsx` adds a `startInFlightRef` guard so a single tab never fires two concurrent `/api/editor/start` calls — collapses React StrictMode double-mount, rapid remount, and the heartbeat re-start race.

## Tests
- `editor-settings-seeding.test.ts` — asserts all 7 persistence keys written, user values win.
- `editor-manager-keeper.test.ts` — concurrent `start(cwd)` spawns exactly one keeper (fails on old code).
- `EditorView.test.tsx` — single `/api/editor/start` under StrictMode double-mount (fails without guard).

All editor/keeper/client suites green (37 tests). Pre-existing unrelated `pi-image-fit` Jimp failures untouched.

## Verification
Manually verified the fix via an isolated worktree harness (temp HOME, real code-server spawn): seeded `settings.json` contained all 7 keys, and the browser showed no Trust dialog / welcome tab / Restricted Mode vs. the old behavior.


robertcsakany:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item.

**Issue:** _Add `cwd` validation at the `start()` API boundary_ (🟡 Minor)
Normalized + reject blank/whitespace `cwd` with `cwd_required` before it becomes a dedup key or spawn input. Added a regression test.

**Files modified:**
- `packages/server/src/editor-manager.ts`
- `packages/server/src/__tests__/editor-manager-keeper.test.ts`

**Commit:** `7e1e7a8f`

The latest autofix changes are on the `os/fix-editor-settings-persistence` branch.

PR #110, opened 2026-06-14
fix(client): depth-aware mobile back navigation

robertcsakany:
## Why

On mobile, the back-arrow and swipe-back couldn't reliably return to the session-card list (depth 0) from ChatView (depth 1). Repro: shrink a desktop window to mobile size while a session is open, then press back — you stay in chat, land on another chat, or leave the app.

Root cause: the MobileShell is **depth-based** (`getMobileDepth` derives 0=list / 1=detail / 2=overlay from the URL), but the back action was **history-based** (`window.history.back()`), which pops whatever URL preceded the current one — not "one depth up". After a window has been in use, the predecessor is commonly a sibling `/session/:id` or a foreign page. The `history.length > 1` guard was unsound.

## What changed (all in `packages/client/`)

- **NEW `lib/back-target.ts`** — pure `computeBackTarget(route): string | null` mapping a route to its parent one shell depth up. depth 0 → `null`, depth 1 → `/`, `/session/:id/diff` → `/session/:id`, ambiguous depth-2 overlays → `/`. `routeDepth`/`parseRouteDepthInput` parse a URL string and reuse `getMobileDepth` (single-sourced depth derivation — no change to `mobile-depth.ts`).
- **NEW `lib/nav-tracker.ts`** — in-app depth-tagged nav stack. Records every navigation (dedupes consecutive identical urls for StrictMode, `replace` overwrites top), realigns on `popstate`, exposes `predecessor()` / `popNav()`.
- **`lib/history-back.ts`** — `goBackOrHome` replaced with hybrid `goBack(navigate, currentRoute, tracker)`: `window.history.back()` fast-path when the tracked predecessor proves shallower (preserves scroll/forward), else `navigate(computeBackTarget(route))`; depth 0 → no-op.
- **`App.tsx`** — instrument the single `navigate` path to feed the tracker, seed the stack + attach the popstate listener on mount, rewire `goBack`. All `onBack={goBack}` call sites untouched.

## Tests

44 new tests across `back-target`, `nav-tracker`, `history-back`, and `back-regression` (TDD — each suite written failing first). Covers the spec scenarios: back from chat → cards regardless of prior chats; depth-2 overlay → one depth up, never a sibling; `history.back()` fast-path when predecessor is shallower in-app; session diff → its session; cold-load fallback.

Desktop unaffected (two-panel list↔detail doesn't use `goBack`; overlay back-arrows resolve to their correct parent via `computeBackTarget`).

## OpenSpec

Archives change `fix-mobile-back-depth-aware`; applies the `url-routing` `Back navigation button` requirement (now depth-aware).


robertcsakany:
Addressed CodeRabbit review in f21efdb1:

- **nav-tracker single-listener contract** — `initNavTracker` is now idempotent (detaches any prior popstate listener); added a regression test.
- **spec contradiction (artifact-tab back)** — scoped the "Tab switching" scenario to the **browser Back** button and dropped the stale `window.history.back()`/`navigate("/")` cold-load wording that conflicted with the new depth-aware app-back rule.
- **file-index paths** — switched the three new rows to repo-accurate `packages/client/src/lib/` paths.

**Not applied — move artifact out of `archive/`:** false positive. `openspec/changes/archive/<date>-<name>/` is the canonical destination the `openspec archive` CLI writes to (the repo already holds many dated dirs there). The "never nest under archive/" guideline applies to *active* change creation, not archived changes; relocating it would break the OpenSpec layout.

PR #115, opened 2026-06-14
feat(openspec): replace-proposal dialog with race handling

robertcsakany:
## Why

When a session has a proposal **manually** attached and the LLM emits a new *active* OpenSpec change (write under `openspec/changes/<name>/` or an `openspec` CLI invocation), the server previously dropped it silently — the user was never asked whether to follow the pivot. Naively rewriting a dialog target with whatever `changeName` just arrived is a footgun: the button text mutates while the user reaches for it.

## What

Server-coalesced, client-committed "replace proposal" dialog. The **suggestion target** updates freely as events arrive; the **commit target** (what the click attaches) only moves on explicit user action.

**Three-layer model:**
1. **Server coalescing** (`event-wiring.ts`) — new manual-attachment branch sets `pendingReplaceProposal` (latest-wins coalesce, same-name no-op, rejected-name ignored). Deleted-proposal bypass via `openSpecChangeExistsInCache`. `agent_end` + detach paths clear the pending/rejected lifecycle.
2. **accept/dismiss handlers** (`session-meta-handler.ts`) wired into the browser gateway. Accept reuses idempotent `applyAttachProposal` with a defensive name-match guard; dismiss records per-name rejection.
3. **Client dialog** (`SessionOpenSpecActions.tsx`) — `ReplaceProposalDialog` on the shared `Confirm` shell with a custom divergence banner. Lazy `committedTarget` keyed by `session.id`: the button commits the first observed suggestion; only `[Use latest]` advances it.

**Shared:** `pendingReplaceProposal` / `rejectedReplaceProposals` on `DashboardSession`; `accept_replace_proposal` / `dismiss_replace_proposal` browser messages.

## Tests

- **10 server** — coalesce, same-name no-op, overwrite, rejection memory, per-name re-prompt, `agent_end` clear, accept+rename, defensive reject, deleted-proposal bypass, `isActive=false`.
- **8 client** — committed-vs-latest invariant, banner + use-latest, accept/dismiss payloads, unmount-on-clear, session-switch remount.
- Fixed 4 pre-existing detach decision-matrix assertions for the new cleared fields.
- Changed files type-check clean against worktree-local shared.

## Notes

- Repairs a pre-existing malformed main-spec header (`## ADDED Requirements` → `## Requirements`) that blocked archiving.
- **Manual smoke tests not yet run** (8.3–8.5: live-dashboard + live-LLM divergence/rejection/agent_end flows) — to be verified later alongside other changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


robertcsakany:
## CodeRabbit Autofix — Applied

Reviewed 6 feedback items. Applied 4, skipped 2 with rationale.

**Fixed** (commit `c9bdfd91`):
- `packages/client/src/hooks/useOpenSpecActions.ts` — guard empty/whitespace `sessionId`/`changeName` + trim before dispatch.
- `packages/client/src/components/__tests__/ReplaceProposalDialog.test.tsx` — add edge-case test (pending set + attached null → no dialog).
- `openspec/specs/proposal-attachment/spec.md` — tighten dialog-render requirement to `pendingReplaceProposal` **AND** `attachedProposal` both non-null.
- `docs/file-index-shared.md` — escape raw `|` in the `pendingReplaceProposal` type cell.

**Skipped:**
- 🔴 *replay_complete should clear pending/rejected* — **declined.** `replay_complete` fires on reconnect, not session end. The design requires `pendingReplaceProposal` to survive reconnect so the dialog persists; pending is never set during replay (detection is replay-gated), so there is no stale state to clear. Clearing here would drop a legitimately-open dialog on every reconnect.
- 🟡 *unescaped `||` in FLOWS row (file-index-client.md:52)* — pre-existing content not authored by this change.

PR #117, opened 2026-06-14
feat(server): emit openspec pending spinner from the poll path

robertcsakany:
## Why

A new worktree/cwd's OpenSpec section "popped in" several seconds later with **no loading indicator**. The three-state loading model and folder-card spinner already exist — the gap was *which code path emits the transitional `pending: true` signal*. It was emitted only on browser cold-boot connect (`buildOpenSpecConnectSnapshot`), never from the poll path that actually discovers a newly-present openspec dir.

## What changed

Add `emitPendingIfDiscovered(cwd)` and call it once at the top of `pollDirectoryGated` — the single choke point all three broadcast wrappers (periodic tick, watcher-fired re-poll, `onDirectoryAdded`) funnel through. It broadcasts `{ initialized:false, pending:true }` before the slow `openspec list` spawn when `<cwd>/openspec/changes/` exists but the cache holds no authoritative `initialized:true` data. The final payload follows when the CLI returns.

- `pollOne` stays pure (preserves the parity/derive test suite).
- Post-bulk-archive refresh is guarded by the `initialized === true` check — no spurious spinner.
- Covers both worktree scenarios: committed `openspec/`, and a delayed `openspec init` hook discovered by a later tick/watcher poll.

### Design note
Refines the change's locked **P2** design (emit from three call sites) into the **one** shared choke point they all route through — eliminating the "easy to miss one" con the design itself flagged, while keeping `pollOne` pure. Recorded in `design.md` + spec.

## Tests
New `directory-service-pending-emit.test.ts` — **5/5 pass**:
- 2.1 pending before final (committed dir)
- 2.2 Scenario 2 discovery: pending then initialized, no straight jump
- 3.1 non-openspec dir → no pending
- 3.2 init-only (no `changes/`) → no pending
- 3.3 terminal clear on failed/empty CLI → `pending !== true`

Client confirmed: `FolderOpenSpecSection.tsx` resolves `!initialized && !pending` → render nothing (spinner clears).

## Notes
- Server-side change — needs a dashboard restart to take effect.
- OpenSpec change archived as `2026-06-14-emit-openspec-pending-from-poll`.
- Task 4.3 (live manual verification) left unchecked — to be tested manually.


robertcsakany:
Addressed CodeRabbit review in f4e2746b:

- **Major (directory-service.ts): terminal clear could be suppressed** — Fixed. `emitPendingIfDiscovered` now records emitted-pending cwds in a module Set; the tick and watcher diff-guarded wrappers OR that flag into their broadcast condition, so the terminal `{ initialized:false }` clear is always delivered even when the final JSON equals the prior cache. `onDirectoryAdded` drops the flag after its poll (event-wiring broadcasts the final unconditionally).
- **Test: order + repeated-empty coverage** — Fixed. Strengthened 2.1/2.2 to assert the exact emit sequence; added test 3.4 that drives two real ticks under fake timers and verifies every spinner is matched by a terminal clear (confirmed it fails without the flag, passes with it).
- **Minor: artifacts under `openspec/changes/archive/`** — Skipping. That path is the standard `openspec archive` CLI output for a completed change. The repo guideline ("never nest under active/ or archive/") governs *authoring* new change artifacts, not the archived terminal state the CLI itself produces.

PR #133, opened 2026-06-15
fix(worktree-spawn): clear orphaned placeholder + place session at top

robertcsakany:
## Why

Spawning a **worktree** session showed two defects (plain spawns unaffected):
1. The *"Starting new session…"* placeholder card stayed stuck at the top of the parent group.
2. The real session card landed at the **bottom** of the group instead of taking the placeholder's slot.

## Root cause

**Defect A — placeholder clear has no worktree fallback.**
`session_added` clears placeholders by Tier 1 (`spawnRequestId` match) then Tier 2 (`spawningCwds.has(session.cwd)`). For a worktree spawn the placeholder is keyed by the **parent repo** (`placeholderCwd`), not the worktree `cwd`, so Tier 2 can never match — the placeholder orphans whenever Tier 1 misses.

**Defect B — worktree session keyed under the wrong order key.**
On `session_register` `gitWorktree` is not yet known, so `resolveOrderKey` returns the raw worktree cwd and the id is inserted there. When `git_info_update` later sets `gitWorktree.mainPath`, the client groups under the parent and reads `sessionOrder["/repo"]` — where the id is absent — so it falls into the unordered bucket (bottom).

## What changed

- **Client** (`useMessageHandler.ts`): Tier 2.5 fallback in the `session_added` arm — scan `pendingSpawns` for a `kind:"spawn"` entry whose path-normalized `cwd` equals the session's, clear its `placeholderCwd`, navigate, first-match-wins.
- **Server** (`session-order-manager.ts` + `event-wiring.ts`): new `rekey(oldKey, newKey, id, { toFront })` helper; the `git_info_update` / `jj_state_update` arms re-resolve the order key and, when it changes, move the id to the **front** of the parent key, prune the stale key, and broadcast one `sessions_reordered`. Guarded to be a no-op when the key is unchanged.

No protocol or persistence-format changes. Behavior-only fix.

## Tests
- `rekey` unit tests (prune, dedupe, no-op, toFront/append).
- Event-wiring integration: fresh worktree re-key to parent front + stale-key prune + single broadcast; repeated update is idempotent (no extra broadcast).
- Client Tier 2.5: worktree clear, plain-spawn no-double-handle, resume-ignore, trailing-slash normalization.
- 38 change-scoped tests green; `npm run build` clean. Pre-existing unrelated failures only (`pi-image-fit` jimp tests, one git `resolveRemoteBase` bare-remote test).

## OpenSpec
Archived `fix-worktree-spawn-placeholder-and-ordering`; synced both ADDED requirements into `placeholder-spawn-card` and `session-ordering` main specs.

## Migration / rollback
None. Revert the two arms + helper; no state migration. Stale worktree-cwd keys are pruned on next register or harmlessly ignored.


robertcsakany:
@coderabbitai review

PR #139, opened 2026-06-17
test(server): WS broadcast load harness (head-of-line blocking)

robertcsakany:
## Why

A user reports periodic WebSocket lag and suspects `openspec_update` frames for non-focused sessions clog the single browser socket. The existing gateway tests prove correctness (serialize-once, skip-when-buffer-full) but are blind to timing — the fake socket's `send` is a synchronous spy and `bufferedAmount` is a static `0`, so head-of-line blocking cannot be reproduced.

This change adds a deterministic, virtual-clock load harness that models the WS send queue, reproduces the suspected head-of-line blocking, and asserts latency / wasted-bytes budgets. **Test-only — no production source modified.**

## What

- **`__tests__/helpers/draining-ws.ts`** — `createDrainingWs({ drainRateBytesPerMs, readyState? })`. Timing-aware fake socket: `send()` fills `bufferedAmount` by frame byte length and records the frame; `advance(ms)` drains FIFO under a caller-owned virtual clock; `timeToFlush(pred) = bytesAtEnqueue / drainRate` (pure head-of-line metric). Drop-in for `makeFakeWs`.
- **`__tests__/helpers/load-fixtures.ts`** — `seedSessions`, `makeOpenSpecPayload(sizeBytes)`, `attachClients`, `subscribeWs`, `DRAIN_FAST`/`DRAIN_SLOW` presets (illustrative, not calibrated).
- **`__tests__/browser-gateway-load.test.ts`** — scenario matrix A–E driving the **real** `createBrowserGateway` + real `broadcastToAll`/`broadcastOpenSpecUpdate`/`MAX_WS_BUFFER` guard.
- **`__tests__/draining-ws.test.ts`** — unit tests for the primitive.
- **`docs/perf-ws-broadcast-load.md`** + `file-index-server.md` rows.

## Findings the harness encodes

- **Cross-cwd leak confirmed**: a focused tab receives every other cwd's `openspec_update` bytes → `wastedBytes(focusedSocket) > 0`. `// REGRESSION TARGET: === 0` marks where a future subscription-scoped fan-out fix lands.
- **Lag is periodic**: scenario E classifies poll-cadence bursts as `PERIODIC` vs quiet traffic as `FLAT` — implicating openspec fan-out, not upstream.

The fix (subscription-scoped fan-out) is **out of scope** — this change only measures.

## Verification

- New tests: **22 pass, deterministic across 3 runs** (zero clock flakiness).
- `openspec validate add-ws-broadcast-load-harness` → valid; capability spec synced and change archived.
- Full `npm test`: new files add zero failures. The ~20 remaining failures are pre-existing flaky env/native-dep tests (image-fit/sharp, doctor-route, session-kill-e2e, keeper), unrelated to this test-only change.


robertcsakany:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item.

**Files modified:**
- `packages/server/src/__tests__/helpers/load-fixtures.ts`

**Commit:** `83cb6184`

Documented the `makeOpenSpecPayload` size contract (payload is at least baseLen bytes; no negative padding when `sizeBytes <= baseLen`). No behavior change; harness tests still 22/22.

The latest autofix changes are on the `os/add-ws-broadcast-load-harness` branch.

PR #178, opened 2026-06-27
feat(serve-agent-artifact-previews): serve agent screenshots from artifact-root anchor

robertcsakany:
Implements OpenSpec change `serve-agent-artifact-previews`.

## What

Adds a third containment anchor (**layer ③**) to `GET /api/file/raw` so agent-tool screenshots written to `~/.agent-browser/tmp` (outside every session cwd and git root) preview correctly instead of 403ing.

- New `packages/server/src/lib/artifact-roots.ts`: realpath-resolved, cached allowlist (`~/.agent-browser/tmp` + `AGENT_BROWSER_SCREENSHOT_DIR`). `isImageUnderArtifactRoot()` is image-only and realpath-contained (collapses `..`/symlinks; missing file → 404 not 500).
- `/api/file/raw` falls through to layer ③ only when cwd/git-root anchors miss. **Not** added to `/api/file` or `/api/file/render`.
- Reuses `git-root-file-containment`'s `within` + `safeRealpath` primitives (avoids `isAllowed`'s logical short-circuit which would defeat realpath escape protection).

## Tests

New `file-artifact-serving.test.ts` (9 cases): image→200, non-image→403, `..`/symlink escape→403, outside-everything→403, missing-root→cwd-only, deleted-image→404.

## Scope note

Best-effort: `agent-browser --screenshot-dir <flag>` writes are invisible to the server (uncovered, A1). Complete fix is inline transport (Fix B), tracked separately.


robertcsakany:
CodeRabbit disposition — 2 of 4 applied, 2 declined with rationale:

**Applied**
- ✅ file-index-server.md: added row for `file-artifact-serving.test.ts`.
- ✅ agent-artifact-serving spec: replaced TBD Purpose with a capability summary.

**Declined (validated as not-applicable)**
- ⏭️ *Keep live OpenSpec change under `openspec/changes/<name>/`* — the change is intentionally under `archive/` because archiving is the **final ship step** (`openspec archive` ran as part of landing). Moving it back would un-archive a completed change.
- ⏭️ *Expand `~`-prefixed artifact paths* — the client always sends the **absolute** linkified path (`/Users/.../.agent-browser/tmp/...`), never a literal `~` (see `FilePreviewOverlay.tsx:166`). `~` never reaches the route, so expansion is out of spec scope and would be speculative. The complete coverage story is Fix B (inline transport), already merged on develop.

PR #185, opened 2026-06-28
feat(extension): filter models_list by enabledModels from pi settings

s7dhansh:
## Problem

The dashboard sends all authenticated models in `models_list` regardless of the user's `enabledModels` setting in `~/.pi/agent/settings.json`. This means the model picker shows every authenticated model (e.g. 64 models across 4 providers) instead of the curated subset the user configured.

`enabledModels` already filters the TUI model selector — but the dashboard bridge was calling `registry.getAvailable()` directly and never consulting the setting.

## Solution

Add `filterByEnabledModels()` in `session-sync.ts` that reads `enabledModels` patterns from `~/.pi/agent/settings.json` and filters the models array before it is sent to the dashboard. Apply the filter at all five points where `models_list` is produced:

| Location | Trigger |
|---|---|
| `sendStateSync()` | initial connect / reconnect |
| `handleSessionChange()` | fork / resume |
| `credentials_updated` handler in `bridge.ts` | provider auth change |
| `session_start` handler in `bridge.ts` | new session |
| `request_models` handler in `command-handler.ts` | browser subscribes to a session |

## Pattern support

- Exact match: `"anthropic/claude-sonnet-4-6"`
- Provider wildcard: `"anthropic/*"`

When `enabledModels` is absent or empty the filter is a no-op, preserving existing behaviour for users who have not configured the list.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32362,6 +32362,7 @@
       "dependencies": {
         "@blackbelt-technology/pi-dashboard-bus-client": "^0.5.4",
         "@blackbelt-technology/pi-dashboard-shared": "^0.5.4",
+        "minimatch": "^10.0.0",
         "ws": "^8.18.0"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


robertcsakany (commented):
Small, clean diff, but I checked the matcher against pi's actual `enabledModels` semantics and there are two real issues — flagging before merge.

## 🔴 1. The matcher is not faithful to pi's `enabledModels` — documented configs produce an *empty* model list

`filterByEnabledModels()` only accepts two pattern forms:

```js
p === "provider/id"   // exact
p.endsWith("/*")      // provider wildcard
```

But `enabledModels` is defined as **"same format as `--models`"**, and pi's canonical resolver (`core/model-resolver.js › resolveModelScope`) matches like this:

```js
minimatch(`${m.provider}/${m.id}`, glob, { nocase: true })
  || minimatch(m.id, glob, { nocase: true })
```

…plus non-glob alias/substring resolution **and** `:thinkingLevel` suffix stripping (e.g. `anthropic/*:high`).

pi's own docs (`settings.md`) give these examples:

```json
"enabledModels": ["claude-*", "gpt-4o", "gemini-2*"]
```

Every one of those **fails** this PR's matcher:

- `gpt-4o` — bare id, no `/` → never equals `openai/gpt-4o`, doesn't end in `/*` → **no match**
- `claude-*`, `gemini-2*`, `*sonnet*` — glob on the id → **no match**
- `anthropic/*:high` — trailing thinking-level suffix → **no match**

Net effect: a user who set a **documented, TUI-working** `enabledModels` gets an **empty** dashboard model picker — strictly worse than today's "show everything." This reimplements pi-internal logic and has already drifted from it.

## 🟠 2. Hardcoded config dir breaks in Docker / custom `PI_CODING_AGENT_DIR`

```js
join(homedir(), ".pi", "agent", "settings.json")
```

pi resolves this via `getAgentDir()` / `getSettingsPath()`, which honor the `PI_CODING_AGENT_DIR` env var and `CONFIG_DIR_NAME`. Since this repo **ships a Docker image**, the hardcoded path silently misses the real settings file there and the filter no-ops. (Fails open, so no crash — the feature just doesn't apply.)

## 🟡 Minor

- `readFileSync` runs synchronously on **every** `models_list` emission (5 paths, including hot reconnect/fork). No cache/invalidation.
- Reimplementing pi-internal matching guarantees future drift as pi evolves.

## Suggested direction

Reuse pi's own scoping rather than reinventing it:

1. Prefer surfacing pi's already-resolved scoped models from the registry / ExtensionAPI if exposed; otherwise
2. use `minimatch` with the **same dual match** (`provider/id` **and** bare `id`, `nocase`), strip a trailing `:level` suffix, and treat non-glob entries via exact-or-substring; and
3. resolve the settings path through pi's config helper, not `homedir()`.

Minimum viable fix if a full rewrite is out of scope: (a) also match bare `m.id` and support `*`/`?` globs via minimatch, and (b) use the env-aware settings path. Without those, the feature is a no-op-or-worse for a meaningful set of documented configs.

PR #187, opened 2026-06-28
fix(fix-darwin-dmg-maker-macos-alias): self-heal macos-alias volume.node on darwin DMG builds

robertcsakany:
Implements OpenSpec change `fix-darwin-dmg-maker-macos-alias`.

## Problem
Local `electron-forge make` on macOS fails in the DMG-maker step with `Cannot find module '../build/Release/volume.node'` — the `macos-alias` native module (transitive: `@electron-forge/maker-dmg` → appdmg → macos-alias) is never compiled when install ran `--ignore-scripts` or Xcode CLT was absent. The `.app` builds; no `.dmg` emits.

## Changes
- **Postinstall self-heal** — `packages/electron/scripts/ensure-macos-alias.mjs` (new). Darwin-only; rebuilds `build/Release/volume.node` when missing via argv-form `spawnSync npm rebuild` (no shell interpolation). Default mode never blocks `pnpm install` (always exit 0); `--rebuild` mode exits 1 when still absent. Wired as `packages/electron` `postinstall`.
- **Build-time gate** — `build-installer.sh` gates `electron-forge make` on darwin: missing `volume.node` → rebuild → `exit 1` with `xcode-select --install` hint when it fails.
- **Doctor diagnostic** — darwin-only `macos-alias native module` row in `doctor-core.ts` (`ok`/`warn`+suggestion; omitted off-darwin or when not installed).
- Unit test (`doctor-macos-alias.test.ts`), docs (`file-index-electron.md`, `faq.md`), CHANGELOG `### Build`.

## Verification
- `ensure-macos-alias.mjs` smoke-tested on a real darwin host: located macos-alias, ran node-gyp, produced `volume.node` (both modes exit 0).
- Full suite: 8299 passing; the only red is a load-sensitive timing assertion (`doctor-route` `elapsed < 3000ms`) that passes 14/14 in isolation. Client build green.
- Spec synced into `electron-build-pipeline`; change archived.


robertcsakany:
Addressed CodeRabbit's 4 findings:

- **`ensure-macos-alias.mjs` `--rebuild` exit code (Major)** — fixed: force/build-gate mode now exits 1 with the actionable hint when `macos-alias` is unresolvable, so the gate fails loudly instead of falling through to `forge make`.
- **Spec ↔ `DoctorCheck` contract (Major)** — fixed: `electron-build-pipeline` Doctor scenarios now use the shipped shape `{ name, section, status: "ok" | "warning" }` and document the omit-when-not-installed behavior.
- **Duplicate `docker-make.sh` index row (Minor)** — fixed: merged into one canonical row.
- **Resolve `macos-alias` from the Electron package root (Minor)** — **skipped (reasoned):** this repo uses npm with root-level hoisting, so `macos-alias` resolves from the shared `doctor-core` context (verified live — the postinstall `--rebuild` path resolves and rebuilds it). The row is intentionally omitted when `macos-alias` is unresolvable. Threading an `packages/electron`-specific root into the shared `doctor-core` would couple the shared module to the electron package layout; not worth the coupling for this edge case.

PR #189, opened 2026-06-29
feat(unify-node-version-gate): single-source Node-version predicates

robertcsakany:
Implements OpenSpec change `unify-node-version-gate` (archived; delta specs synced).

## Why
The nodejs/node#58515 affected range + the `engines.node` cap were encoded in three drifting copies. The Electron `dependency-detector.ts` copy still used the old `v22.x < 18` cutoff, so the doctor reported a system Node v22.18 (and v21.x, v26+) as usable even though the server refuses to start on them.

## What
- New `packages/shared/src/node-version.ts` — single source of truth: `isAffectedNode`, `isOutOfEnginesRange`, combined `isUsableNodeVersion` (rejects unparseable strings).
- `node-guard.ts` imports + re-exports the two predicates (public API unchanged).
- Electron doctor `detectSystemNode` / `scanForUsableNodeOnDisk` gate on `isUsableNodeVersion`.
- Net effect: doctor reports a system Node usable iff the server accepts it — **22.19+, 24, 25 usable; 22.0–22.18, 24.1–24.2, 21.x, 26+ rejected**. Server runtime behavior unchanged (predicates relocated). Net −1 Biome complexity finding.

## Tests
- Shared accept-set table + re-export reference-identity assertion.
- Electron fixtures bumped off the now-rejected v22.18.


robertcsakany:
@coderabbitai full review

PR #206, opened 2026-06-30
feat(optimistic-prompt-progress): idle-scoped optimistic prompt with progress states

robertcsakany:
Implements OpenSpec change `optimistic-prompt-progress`.

## Summary
Re-introduces the optimistic prompt bubble, **scoped to idle sends**, with a `sending → sent → confirmed` progress visual. Mid-turn sends remain governed by `mid-turn-prompt-queue` (no optimistic write). A new bridge `prompt_received` ack (capture-before-send streaming verdict, bridge→server→browser) resolves the same-tick race and drives the "sent" state.

## Key changes
- `PendingPrompt` gains required `status`; new pure `applyPromptReceived` reducer.
- `handleSend` / `handleSendPromptToSession` write `pendingPrompt` only when not mid-turn (absent session state treated as idle; bridge ack reconciles a raced mid-turn send by dropping it).
- `ChatView` re-activates the bubble with sending/sent states at identical geometry (zero layout shift); drops the `queuedTexts` suppression band-aid.
- Protocol: `PromptReceivedToServer/BrowserMessage`; server pass-through in `event-wiring.ts`; bridge emits the ack from both the passthrough and slash paths.

## Tests
- Unit/render: reducer ack branches, message-handler case, idle-vs-mid-turn write guards, sending/sent render + geometry parity.
- **Playwright E2E** (`tests/e2e/optimistic-prompt.spec.ts`): idle-appears-then-confirms + mid-turn-queue-chip, against the Docker faux harness. **System-browser support** added via `PW_CHANNEL` + `npm run test:e2e:chrome`. Both pass against the container.

## Notes
- Surfaced + fixed a real bug: a freshly-selected idle session has no `SessionState` entry yet, so the original guard skipped the optimistic write — now treats absent state as idle.
- Repairs a pre-existing structurally-malformed `optimistic-prompt` main spec so archive could sync the delta.
- Full local suite: failures were all confirmed environmental real-server/timing flakes (pass in isolation); client build green.


robertcsakany:
This file lives under `openspec/changes/archive/` because the change was **archived** by the standard ship flow (`openspec archive optimistic-prompt-progress`), which is the canonical location for completed changes (`openspec/changes/archive/YYYY-MM-DD-<name>/`). The "never nest under `archive/`" guideline applies to *authoring* a new change, not to archived ones. No action needed.

PR #212, opened 2026-07-02
docs(openspec): propose fix-openspec-config-read-bundled-node

molnar-botond:
Capture root cause + design for OpenSpec global-settings failing in the bundled Electron app on macOS/Windows. The managed openspec CLI is a '#!/usr/bin/env node' shebang script; the bundled server's child-spawn PATH can lack a real 'node', and nodeScriptToArgv only node-wraps on Windows, so the unix spawn dies (exit 127). configListOrAsync swallows the failure to null, so GET /api/openspec/config returns an empty profile and Settings renders 'not found'.

Artifacts: proposal, design, specs (tool-registry + openspec-profile-config deltas), tasks. Planning only, no code changes.

### DIFF
--- a/openspec/changes/fix-openspec-config-read-bundled-node/.openspec.yaml
+++ b/openspec/changes/fix-openspec-config-read-bundled-node/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-01
--- a/openspec/changes/fix-openspec-config-read-bundled-node/design.md
+++ b/openspec/changes/fix-openspec-config-read-bundled-node/design.md
@@ -44,13 +44,13 @@ Edge case — server running under `ELECTRON_RUN_AS_NODE` (the `execpath-fallbac
 
 **Why over alternatives:** deterministic — the interpreter is supplied explicitly, independent of PATH contents or install topology. Mirrors the already-proven Windows path (single code path, less divergence). The shebang approach is exactly what fails today.
 
-### Decision 2 — Seed a real `node` dir into `buildSpawnEnv` (defense-in-depth, secondary)
+### Decision 2 — Seed a real `node` dir into `buildSpawnEnv` — TRIED, then REVERTED as redundant
 
-Add the resolved node bin dir (and the managed `~/.pi-dashboard/node/bin`, when present) to the PATH `buildSpawnEnv` constructs, so any *other* shebang-based child spawn also resolves `node`.
+Originally proposed as defense-in-depth: add the managed `~/.pi-dashboard/node/bin` to the PATH `buildSpawnEnv` constructs, so any *other* shebang-based child spawn also resolves `node`.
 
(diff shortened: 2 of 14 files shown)


molnar-botond:
## Full-sweep review — ✅ sound, complete, and correctly proven

Independent verification of the exec fix + diagnosability fix against the root cause. Ran the tests locally (`HOME=$(mktemp -d) npx vitest run`) and confirmed CI is green.

### Root-cause fix (node-wrap) — correct & genuinely in the spawn path
- `resolveJsScript` derefs the unix `.bin/openspec` shebang symlink → real `bin/openspec.js`; Windows path skipped (byte-for-byte preserved).
- `nodeScriptToArgv` node-wraps `.js` on unix **and** win32; interpreter from `registry.resolve("node")`, fallback `process.execPath`.
- Verified `resolveExecutor` applies `toArgv`, so the wrap actually reaches `spawnSync`/`spawn` — not just argv shape.
- `buildSpawnEnvForArgv` sets `ELECTRON_RUN_AS_NODE=1` only when `execCmd === process.execPath` under Electron — tight guard, no over-reach.

### Diagnosability fix — real UX, not a blank panel
- `GET /api/openspec/config` → `configListAsync` (Result) → **HTTP 502** on failure, **not cached**; success path keeps the `custom`+expanded → `expanded` alias.
- Client `fetchOpenSpecConfig` throws on `!res.ok`; `OpenSpecProfileSection` has a `loading|ready|error` state machine → renders "Couldn't load the current profile" + Retry (`data-testid="profile-error"`). Confirmed the trade of fake-empty → honest error actually lands in the UI.

### Tests — strong
- argv matrix (unix `.js`, win32 `.js`, unix `.bin` symlink-deref, passthrough).
- **stripped-PATH spawn proof** (`env:{PATH:""}` → node-wrapped argv exits 0; raw shebang would be 127) — directly reproduces & disproves the bug.
- route: 502-on-exit-127, no-cache, expanded alias.

### Re: the `buildSpawnEnv` revert (`0d2d6493`) — confirmed correct
`prependManagedNodeToPath` (from `embed-managed-node-runtime`) already guarantees the node dir for callers that use `buildSpawnEnv`; the openspec spawn is node-wrapped by absolute path so it never routed through `buildSpawnEnv`; and the formerly-flaky `process-manager-managed-path` test now passes locally. Good call.

### Follow-up pushed (`ef9415f7`)
Two non-blocking nits from the sweep:
1. **Added coverage** for the one untested branch — `buildSpawnEnvForArgv`'s `ELECTRON_RUN_AS_NODE` fallback (`runner-spawn-env.test.ts`, 5 cases). Exported the helper with injectable `execPath`/`electronVersion` for deterministic testing.
2. **Corrected `design.md`** so the archived record matches what shipped: Decision 2 (the `buildSpawnEnv` seed) is now documented as *tried then reverted as redundant*, not "ship alongside."

LGTM. 👍

PR #215, opened 2026-07-02
feat(filter-oauth-incompatible-models): filter OAuth-unreachable models from /v1/models

robertcsakany:
Implements OpenSpec change `filter-oauth-incompatible-models`.

## Problem
Model proxy `/v1/models` advertised every pi-ai model for any provider with *some* credential, filtering only at provider granularity. When the only credential is OAuth (Claude Pro/Max, Codex), the upstream accepts a subset — legacy dated snapshots (`claude-3-5-haiku-20241022`, …) were listed yet returned a confusing upstream `404 not_found_error` mid-stream.

## Change
- `InternalRegistry.getAvailable()`/`find()` filter by **credential-kind × model id** (`canRouteModel`): `api_key` routes everything; `oauth` routes a model only when not flagged in the hand-maintained `packages/server/src/model-proxy/oauth-compat.ts` override table (`oauthCompatible !== false`).
- New `GET /api/model-proxy/diagnostics` — per-model `excludedReason` (`null` / `no-credential` / `oauth-incompatible`).
- Additive: no client/protocol/config changes, no migration.

## Verification
- **Unit**: `oauth-compat`, `internal-registry`, `model-proxy-diagnostics-routes` tests. `npm test` green (8474 passed), `npm run build` ok.
- **E2E** (`tests/e2e/model-proxy-oauth-filter.spec.ts`, Playwright `request` fixture via **system Chrome** against the Docker harness): `/v1/models` **23→15** models (8 `claude-3-*` snapshots dropped), legacy id → proxy **404**, diagnostics annotates `oauth-incompatible`. Opt-in `PW_SYSTEM_CHROME=1` path added (no bundled-chromium download).


robertcsakany:
Addressed CodeRabbit feedback in f52ec0a:

- **CHANGELOG (endpoint behavior):** ✅ Reworded — `/v1/models` is *filtered* (model omitted from the list); the `404` occurs on a completion request that names an omitted id (registry `find()` miss), not on the list endpoint.
- **docs/file-index-server.md (union pipes):** ✅ Escaped the pipes (`\|`) so the row stays a 2-column table.
- **proposal.md (route module):** ✅ Repointed to `model-proxy-diagnostics-routes.ts`.
- **e2e diagnostics auth:** ⏭️ Skipped — not a still-valid issue. The `/api/*` JWT gate (`auth-plugin.ts` global `onRequest`) is only installed when an auth provider is configured; the E2E harness configures none, so the route is reachable and the test **passes as verified** (3/3 against Docker, diagnostics returns `excludedReason`). The suggested `AUTH` bearer is also inapplicable here — that proxy key gates `/v1/*` only, not `/api/*`.

PR #233, opened 2026-07-04
feat(fix-nodescript-argv-electron-execpath-fallback): env-layer guard for node-wrapped Electron-execPath spawns

robertcsakany:
Implements OpenSpec change `fix-nodescript-argv-electron-execpath-fallback`.

## Summary

Extracts a shared `electronAsNodeRequired(argv0, deps?)` predicate from `runner.buildSpawnEnvForArgv` and applies it at every executor spawn site so the argv-blind `process-manager.buildSpawnEnv` strip cannot desync from the argv-aware re-add.

In the triply-degraded `execpath-fallback` topology (server on the Electron binary, no managed runtime `node`, no PATH `node`), `spawnWt`, `spawnHeadless`, and the RPC keeper's own `[execPath, keeper.cjs]` launch now re-add `ELECTRON_RUN_AS_NODE=1` instead of only stripping it — preventing a silent Electron-GUI relaunch that exits on the single-instance lock.

## Scope / safety

- Healthy installs, standalone CLI, and `runner.buildSpawnEnvForArgv` are byte-identical (absent `argv0` ⇒ no-op).
- `definitions.ts` `nodeScriptToArgv` fallback left unedited (avoids conflict with `fix-node-electron-resolution-test-isolation`).
- No edits under `packages/electron/`. Diff confined to `packages/shared/**` + `packages/server/**`.

## Tests

- `spawn-env-electron-parity.test.ts` — both env builders agree (`ELECTRON_RUN_AS_NODE=1`) for an Electron `argv[0]`; healthy path adds nothing; absent-argv0 no-op.
- `keeper-electron-run-as-node.test.ts` — keeper own-spawn carries the flag under Electron, independently of the pi argv.


robertcsakany:
@coderabbitai full review

robertcsakany:
Applied: reworded archived task 4.5 to stop overstating suite status and removed the duplicated fragment. The flagged `node-electron-resolution.test.ts` failure was a machine-local test-isolation leak owned by the sibling change `fix-node-electron-resolution-test-isolation`, which is now on develop — the merged tree runs green. Thanks @coderabbitai.

PR #236, opened 2026-07-04
feat(add-server-keypair-pairing): optional public-key device pairing + bearer auth

robertcsakany:
Implements OpenSpec change `add-server-keypair-pairing`.

Adds an **optional** public-key device-pairing method alongside OAuth:

- **Server identity** — persistent Ed25519 keypair (`identity.key`, 0600), `sha256:` fingerprint, nonce challenge-response.
- **QR / copy-string pairing** — short-lived one-time codes; D12 compare-code approval (operator types the device's code, consume-on-approval, rate-limit + lockout, one pending device per code).
- **Bearer device auth** — opaque long-lived tokens in a revocable registry (`paired-devices.json`, 0600; only SHA-256 hash stored); one added `OR` branch feeding `request.isAuthenticated`; Settings → Security → Paired Devices revoke.
- **WS single-use tickets** (D11) — route-scope-bound, deleted on first upgrade; durable bearer never rides the WebSocket (F6).
- **Genuine-local trust** (D10, narrowed) — loopback trusted only when no proxy-forwarding header is present → closes the zrok-tunnel-as-`127.0.0.1` auth bypass; affirmative local-IPC token for same-host process callers.
- **CORS default** — `https://pi-dashboard.dev`.
- **Neutral static PWA shell** (`packages/shell`) — IndexedDB keyring, pairing UI, WebCrypto identity verification, endpoint racing; published to `pi-dashboard.dev/app/` via `deploy-site.yml`.

Also repairs two pre-existing malformed main specs (`cross-origin-client`, `server-cors` — added `## Purpose` / `## Requirements` sections) so the archive spec-sync could run.

Tests: server 2830 / client 2980 / shell 4; full `npm test` 9229 passing. QA/manual tasks deferred to post-merge verification.


robertcsakany:
Addressed all actionable CodeRabbit feedback across 2 fix commits. Two remaining threads are deliberate, non-blocking decisions:

1. **`vitest.config.ts` (shell registration)** — I chose to register `packages/shell` in the root `test.projects` so `npm test` covers the shell (single test entrypoint), and updated `packages/shell/AGENTS.md` to reflect that. Keeping it standalone would silently drop the shell from CI's default run.
2. **archived `tasks.md` WS wording** — that file is the historical record of the original plan. The Phase-B reconciliation (WS is ticket-only, durable bearer never rides the socket) is captured in the live specs (`bearer-device-auth`, `cross-origin-client`), which are the durable contract.

PR #248, opened 2026-07-06
fix(fix-double-thinking-row-on-replay-reconstruction): dedupe reasoning reconstruction on message_end

robertcsakany:
Implements OpenSpec change `fix-double-thinking-row-on-replay-reconstruction`.

## Problem
`develop` CI was red: 3 client reducer tests asserted a single `thinking` row but the reducer produced two. A turn that streamed `thinking_*` events (pushing one thinking row) and then reached a `message_end` whose default `opts.isLive` is `false` triggered the `!isLive`-guarded reconstruction block again → duplicate thinking row.

## Fix
Replaced the `!isLive` guard in the `message_end` reasoning-reconstruction block with a real dedupe: skip reconstruction when a `role:"thinking"` row already exists in the current turn window (walk back to `TURN_BOUNDARY_ROLES`). Cold replay (`state-replay.ts`) emits no `thinking_*` events → no prior row → reconstruction still fires. Client-only; no protocol/server/persistence change.

## Verification
- 3 previously-failing test files now pass; added a regression test (streamed thinking + non-live `message_end` → exactly one thinking row).
- Full client suite green (2990 passed).
- Specs synced + change archived.


robertcsakany:
@coderabbitai full review

PR #262, opened 2026-07-08
feat(banner): single-card error/retry surface, no regex, clear-only dism

molnar-botond:


molnar-botond:
Superseded: the banner single-card error/retry feature already landed on `develop` as commit 1dcb66c1. After merging develop and resolving the (unrelated) add-agent-role-model-tools openspec conflicts, this branch's tree is byte-identical to develop — merging would be an empty no-op. Closing and deleting the branch.

PR #273, opened 2026-07-12
fix(chat): content-aware row estimate + scroll-to-top affordance

robertcsakany:
Implements OpenSpec change `fix-chat-scroll-to-top-estimate-drift`.

## Problem
Scrolling up a virtualized chat transcript never converged on the first row: the largest rows (a pasted image, a 24k-char tool result) sit near the top, and the static per-role pre-measure estimate under-shot them 10–50×, so on mount `getTotalSize()` jumped and the top receded.

## Fix
- **Content-aware estimate** (`chat-virtual-rows.ts`): `base + ceil(textChars/80)*20` clamped 8000px + per-kind image reserve (300 user / 512 toolResult). `computeRowTextChars` aggregates rendered text once per `displayRows` rebuild so `estimateSize` stays O(1). Shrinks the delta that drives TanStack's built-in above-viewport correction below perception.
- **Scroll-to-top affordance** (`ChatView.tsx`): `ascendingRef` latch. `scrollToIndex(0)` is bounded (`maxAttempts=10`), so `onChange` re-issues it on measurement growth (covers the async image-load remeasure); `handleScroll` ascending branch prevents the bottom-pin re-arm race. No manual `scrollTop+=delta` (would double-move); `overflowAnchor:"none"` kept.
- **Cross-model doubt review** (DeepSeek) folded into the design before implementation: corrected the false "scrollToIndex retries forever" claim, added the re-arm-race guard, per-kind image reserve, and made the e2e the required gate.

## Tests
- Unit: monotonic estimate, per-kind image reserve, aggregate textChars (19 pass).
- Component: scroll-to-top logic guards / re-arm race (7 pass). jsdom convergence assertions deliberately omitted — the shim reports 0-height rows + a no-op ResizeObserver, so they'd be vacuous.
- **E2E (required gate)**: `tests/e2e/scroll-to-top.spec.ts` via `[[faux:scroll-top-heavy]]` — 2/2 green against the Docker harness. Test 2 fails if the top recedes (the pre-fix bug); test 1 fails without the new button.

QA/manual tasks (6.1/6.2) were covered by the automated e2e + existing virtualization suite.


robertcsakany:
CodeRabbit triage: applied nitpick 2 (added clamp-boundary + saturation + empty-payload edge-case tests). **Skipped nitpick 1** (import `SCROLL_TOP_HEAVY_TAIL` instead of duplicating `TAIL`): the sibling `chat-transcript-virtualization.spec.ts` deliberately duplicates `LONG_TRANSCRIPT_TAIL` with a keep-in-sync comment specifically so e2e specs do NOT import the pi-ai-laden `qa/fixtures/faux-scenarios.ts` module; this spec follows that established convention.

PR #278, opened 2026-07-12
feat(preserve-chat-selection-during-churn): keep transcript selection alive across churn

robertcsakany:
Implements OpenSpec change `preserve-chat-selection-during-churn`.

Keeps an active text selection in the chat transcript from collapsing when the
transcript churns (streaming, new cards, auto-scroll, virtual-window recompute).

## Changes
- **Detect** (`useActiveChatSelection`): `selectionchange`-driven `isSelecting`
  (true when a non-collapsed selection has anchor **or** focus inside the
  container — cross-boundary safe). Proactively captures the selection's
  display-row span into a ref while rows are still mounted.
- **D2 — suspend auto-scroll:** the `stickToBottom` layout effect and the
  virtualizer `onChange` bottom-pin early-return while selecting (without
  clearing `stickToBottomRef`); the pin re-fires on the collapse edge.
- **D3 — retain rows:** a custom `rangeExtractor` unions the tracked selection
  span into the mounted range so intersecting rows never unmount. Device-aware
  ceiling (desktop 100 / mobile 40); past it the selection is actively cleared
  (`removeAllRanges`) rather than handed back as a silently-truncated copy.
- **Path B (streaming tail):** stays at baseline. Filed follow-ups
  `preserve-streaming-tail-selection` + `chat-copy-fidelity-intercept`.

## Tests
- New: `useActiveChatSelection.test.tsx`, `ChatView.selection.test.tsx`, plus
  `rangeToRowIndexSpan` / `extendRangeWithSelection` cases in
  `chat-virtual-rows.test.ts`.
- Full `chat-scroll-lock` regression suite unchanged.

QA/manual task (6.3, live-browser reproduction pass) deferred to post-merge verification.

> Note: pre-existing `Jimp is not a constructor` failures in the untouched
> `pi-image-fit-extension` package are a local-env quirk (jsdom canvas / jimp
> export), unrelated to this change; CI's clean install resolves them.


robertcsakany:
@coderabbitai full review

PR #279, opened 2026-07-12
feat(gate-session-worktree-button-on-git): hide +Worktree only in confirmed non-git folders

robertcsakany:
Implements OpenSpec change `gate-session-worktree-button-on-git`.

## Problem
The session-card `+Worktree` button was offered in folders that are not git repositories, where `git worktree add` cannot succeed. The naive `gitBranch` gate is unsound — `gitBranch` is a *data-arrival* signal (absent during the register race, on probe failure, and after a server restart for cold sessions).

## Fix
A **tri-state** `isGitRepo` signal: `true` (confirmed repo), `false` (confirmed non-git, git exit 128), `undefined` (unknown — probe inconclusive / legacy). It is:
- computed synchronously on the bridge and carried on `session_register` (no arrival race),
- persisted to `.meta.json` and restored by `sessionFromMeta` on cold start,
- used to hide the button (card + folder header) **only** on `isGitRepo === false`.

A failed/timed-out probe yields `undefined`, never `false` — so a real repo never loses the button.

## Tests
- shared: `SessionMeta` round-trip preserves the tri-state.
- extension: 7 `detectIsGitRepo` cases (ok true/false, exit-128→false, not-found/timeout/spawn/other-exit→undefined).
- server: live-register persist (false/true/absent) + 3 scanner cold-start restore cases.
- client: 6 gate cases (card + folder header × false/true/undefined).
- Isolated manual smoke: real-git detection + alt-port live server persist/restart-restore, all pass.

## Notes
- Also repairs a pre-existing structural defect in `openspec/specs/git-context/spec.md` (missing `## Purpose`, stray `## ADDED Requirements` delta header) that blocked archive.
- QA/manual tasks verified via isolated smoke; deferred items marked for post-merge verification.


robertcsakany:
@coderabbitai full review

PR #283, opened 2026-07-12
feat(fix-remote-connect-cors-gates): unblock LAN remote-connect on Electron + web

robertcsakany:
Implements OpenSpec change `fix-remote-connect-cors-gates`.

## Problem
A trusted-network LAN dashboard is reachable (address-bar visit + curl work) yet selecting it as a "remote" fails on both surfaces — each blocked by an *origin*-based CORS gate a top-level navigation never faces but an in-app `fetch` does.

## Fixes
- **Server CORS** (`cors-origin.ts`, new pure + unit-tested decision): allow origins whose host matches `resolvedTrustedNetworks` (via `isBypassedHost`, CIDR/wildcard/exact). `null`-origin refusal + unknown-origin rejection preserved. Sole server-side enabler for the web path.
- **Electron attach**: reachability probed in the **main process** (`probeRemote`, Node fetch, no `Origin`, not CORS-bound) via new `dashboard:probe-server` IPC; `loading.html` does a CORS-free top-level nav for non-loopback targets instead of a `null`-origin renderer fetch a remote deliberately refuses. Loopback / older-preload fallback unchanged.
- **Web `ServerSelector`**: a CORS-blocked probe on a LAN host now renders "CORS-blocked — allowlist this origin" (new `cors-blocked` state + `isLanHost`) instead of a dead "Unreachable" row.

## Dropped during implementation (user-confirmed)
Original tasks 3.1–3.3 (WS-ticket minting) removed: both WS-upgrade gates short-circuit on a **trusted source IP before the ticket check** (`server.ts` no-auth branch + `validateWsUpgrade`), so the ticketless staging + committed sockets already pass for the trusted-network scenario. Minting was unreachable dead weight (and would have broken auto-reconnect). `server-switch.ts` / `staging-socket.ts` untouched.

## Tests
New/updated unit tests: server CORS (18), electron `remote-probe` (9), client `server-selector` + `ServerSelector` (26). `tsc --noEmit` clean for this diff; zero new Biome violations.

## Deferred
Section 5 (manual multi-machine LAN validation) deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

PR #288, opened 2026-07-12
feat(fix-ctx-running-render): args-derived chip + running preview for ctx_* cards

robertcsakany:
Implements OpenSpec change `fix-ctx-running-render`.

## Problem
`CtxToolRenderer` described a `ctx_*` call only from its result text. While a call was still running there was no result, so the header chip fell back to the bare tool name (duplicating the subtitle → `ctx_batch_execute · ctx_batch_execute`) and the body showed a bare `Running…`, discarding the `args` already in hand.

## Fix
- `argsChip(toolName, args)` — args-derived chip using the result-chip emoji vocabulary (`▦ N cmds`, `⚙ lang`, `🔍 N queries`, `🌐 host`, `🗂 source`); used as the `headerChip` fallback so the chip never collides with the tool-name subtitle during running/raw states.
- `RunningPreview` — previews pending work from `args`: batch command list, execute code, search queries, fetch url; unknown ctx tools keep a plain `Running…`.
- Result-state parsing/layout untouched.

## Tests
- Unit: `CtxToolRenderer.test.tsx` running-state cases (chip ≠ subtitle regression guard; batch labels; execute code; search queries).
- Browser E2E: `tests/e2e/ctx-running-render.spec.ts` drives `[[faux:ctx-batch-running]]` and drops the `tool_execution_end` frame to freeze the running card; asserts the args chip + command-list preview. Verified with `PW_CHANNEL=chrome PW_E2E_USE_RUNNING=1` against the Docker harness.


robertcsakany:
@coderabbitai full review

PR #289, opened 2026-07-12
feat(preserve-streaming-tail-selection): node-stable streaming tail preserves anchored selection

robertcsakany:
Implements OpenSpec change `preserve-streaming-tail-selection` (follow-up to `preserve-chat-selection-during-churn` D4).

## What

A selection anchored inside the actively-streaming tail card no longer collapses on a chunk append or at turn completion:

- **Freeze on select:** on the `isSelecting` false→true edge, if the selection sits in the live tail, snapshot `streamingText`. The tail then renders from the stable snapshot → `MarkdownContent`'s `React.memo` skips re-render → committed Text nodes are never replaced per chunk (buffer).
- **Survive turn completion:** the frozen snapshot keeps the tail mounted after `message_end` clears `streamingText`; the committed assistant twin is hidden from the view (`role:"assistant"` + `content.startsWith(snapshot)`) — **never dropped from `state.messages`**, so `tool_execution_start`/`message_end` mutations are preserved.
- **Flush on collapse:** clear the snapshot → live text; the twin is revealed.

## Tests

`ChatView.streaming-tail-selection.test.tsx` — selection survives a chunk append; survives turn completion; twin hidden→revealed. Uses a production-faithful harness (stable `ThemeProvider` above the state owner) so `React.memo` behaves as in the real tree.

**Perf (task 2.3):** tail `MarkdownContent` renders — baseline (no selection) 5/5 chunks (1-per-chunk, unchanged); frozen (held) 0/5 (fully coalesced). Idle path is behavior-identical.


robertcsakany:
@coderabbitai full review

PR #293, opened 2026-07-13
feat(copy-file-path): per-row copy-path popup on the editor file tree

robertcsakany:
Implements OpenSpec change `copy-file-path`.

Adds a hover-revealed **copy-path popup** to every editor file-tree row (files and directories):

- Copy glyph fades in on row hover / while its popup is open, flush-right.
- Glyph `stopPropagation` — never opens the file or toggles a directory.
- Anchored popup: **Copy full path** (`cwd + rel`), **Copy relative path** (`rel`), **Copy file name** (basename), with a truncated absolute-path header. Flips above the glyph near the rail bottom.
- Guarded `navigator.clipboard?.writeText` (silent no-op when unavailable), transient ✓ (~1.5 s), dismiss on outside-click / rail scroll / Escape (Escape refocuses the glyph).

New `EditorFileTree.test.tsx` suite (12 green) covers all three copies, directory copy, the three dismiss paths, no-open-on-glyph, and the clipboard-unavailable no-op. A11y: glyph is a focusable button; popup items are `role="menuitem"` buttons.

QA/manual task (live browser check) verified by hand.


robertcsakany:
@coderabbitai full review

PR #379, opened 2026-07-21
feat(shrink-client-index-chunk): split @mdi/js out of the eager index entry chunk

robertcsakany:
Implements OpenSpec change `shrink-client-index-chunk`.

## What
- Convert the two dynamic `import("@mdi/js")` sites (`ActionList.tsx`, `StatusPill.tsx`) to static namespace imports — removes the `dynamic import will not move module into another chunk` warning at its root (module was both dynamically and statically imported).
- Add `"mdi": ["@mdi/js"]` to `manualChunks` so the ~2.6 MB icon set leaves the eager `index` chunk into a separately-cacheable `mdi` chunk. **Relocates** bytes (better caching); does not reduce initial-download bytes (App eagerly imports `@mdi/js`).
- Guard: `mdi-chunk-size.test.ts` (dedicated `mdi` chunk exists, `@mdi/js` absent from entry, gz `index` ≤ 900 KB) + a CI build-log grep forbidding the `@mdi/js` dynamic-import warning.

## Results
- `index` chunk: **1388 KB gz → 580 KB gz**; `mdi-*.js` chunk emitted (804 KB gz).
- `@mdi/js` dynamic-import warning gone. Oversized-chunk (>700 kB) aggregate warning intentionally remains (documented; `monaco` is the floor; `chunkSizeWarningLimit` stays 700).


robertcsakany:
@coderabbitai full review

PR #390, opened 2026-07-22
feat(add-hermes-memory-settings-plugin): dashboard settings surface for pi-hermes-memory

robertcsakany:
Implements OpenSpec change `add-hermes-memory-settings-plugin`.

## What
New package `packages/hermes-memory-plugin` — a dashboard **settings-section** plugin for the external `pi-hermes-memory` pi extension. Reads/writes the extension's on-disk `~/.pi/agent/hermes-memory-config.json` via two validated Fastify routes and a full-coverage grouped settings form.

- `GET /api/plugins/hermes-memory/config` — per `MemoryConfig` field: effective value / default / `isDefault`, plus resolved `filePath` + `exists` + `raw`.
- `PUT /api/plugins/hermes-memory/config` — validates the body (unknown-key allowlist, type/enum/numeric-bound, `correction*Patterns` regex-compile) **before** any write; invalid → 400, no write; success → atomic full-write (tmp + rename, pretty JSON).
- Client: 9-group accordion form over every field, per-field DEFAULT badge + Reset, inline validation (Save disabled while invalid), `memoryPolicyCustomText` revealed only when style is `custom`, prefers-reduced-motion guard, raw-JSON view, "applies to new sessions" notice.
- `requires.piExtensions: ["pi-hermes-memory"]` — section + routes active only when the extension is installed. No dependency on the external package (re-declares `MemoryConfig` + defaults, source-version pinned to `pi-hermes-memory@0.8.1`).

## Tests
35 new tests: shared schema/validation, path resolution, atomic IO, routes (injected Fastify), component behavior (default badge / reset / save-PUT), manifest activation gate. Registry, publish allowlist, and bundle-server plugin list updated.

QA/manual tasks (isolated-browser QA, extension-present visibility) deferred to post-merge verification.


robertcsakany:
Addressed CodeRabbit's 11 actionable items:

**Applied** (commit b079b44a7): `validateSessionSearch` now rejects extra keys (+test); `load()` uses an `AbortController` (no setState-after-unmount); raw-JSON modal closes on Escape; added a malformed-JSON read test; alphabetized client/server `AGENTS.md` rows.

**Skipped (with reasons):**
- "Move the change out of `openspec/changes/archive/…`" + "uncheck 9.3/V.2" — false positives: the change is *correctly* archived by the ship workflow, and 9.3/V.2 are explicitly annotated `DEFERRED — manual, tested later` (the intended post-merge deferral).
- `writeResolvedConfig` require-complete-config — contradicts design D5/D6 (client owns full-resolution via `buildResolvedConfig`; the server validates present keys and never forces completeness).
- json fence tag on archived `design.md` — cosmetic edit to an archived artifact.
- Broad `htmlFor` label association / full focus-trap — nitpicks on a surface that already passed the ux-review a11y gate; deferred to avoid scope-expansion.

Note: `job-object-windows` is red but is a **pre-existing, non-required** infra job that fails on every recent run across unrelated branches; `electron-e2e (windows-latest)` passes with this plugin bundled.

PR #391, opened 2026-07-22
feat(image-fit-tool-result-images): fit oversize any-origin images via a context-event seam

robertcsakany:
Implements OpenSpec change `image-fit-tool-result-images`.

## What

Adds a second interception seam to `pi-image-fit`: a `pi.on("context")` handler that fits oversize `ImageContent` of **any origin** (tool_result / user-pasted / historical) in the per-turn message deep copy, before each LLM call. This rescues already-persisted oversize sessions on reload — the `019f8604` failure mode (an 8956×5080 browser screenshot under the byte limit but over the pixel limit) that no creation-time `read`-path hook could cover.

## How

- **`resize.ts`** — buffer helpers (`probeDimsFromBuffer`, `resizeBuffer`, `outputFormatForMime`, `estimateBytesFromBase64`) + a cheap `probeDimsFromHeader` (PNG IHDR / JPEG SOF / WEBP VP8X·VP8·VP8L / GIF header parse, no pixel decode).
- **`cache.ts`** — `ContentCache`, a bounded in-memory LRU keyed by mime-scoped SHA-256, separate from the temp-file cache; never touches disk.
- **`extension.ts`** — role-agnostic `context` handler + exported `fitContextMessages`; cheap-probe gate → hash → cache → `resizeBuffer` on miss; per-block fail-open.

Reuses the `PI_IMAGE_FIT_*` thresholds; `PI_IMAGE_FIT_DISABLE` disables **both** seams. The on-disk transcript is never rewritten. jimp only, no native deps.

## Tests

24 automated L1 scenarios (E1–E18, P1–P2, X1–X4) across the three folded test files — **120/120 pass**, Biome clean, no new tsc errors. The steady-state (within-limit) and cache-hit (repeat oversize) perf invariants are asserted as tests (P1/P2).

M1 (subagent-seam coverage probe) and M2 (end-to-end rescue reproduction) are `manual-only` — deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

PR #397, opened 2026-07-24
fix(fix-bridge-resume-disconnect): guard ctx.cwd throw on session replacement (#393)

### LINKED ISSUE #393: Bridge unconditionally disconnects the WebSocket on `session_shutdown`, so resuming/switching a session in the TUI permanently drops the dashboard connection
---

## Summary

Resuming or switching a session inside the pi TUI **permanently disconnects the bridge from the dashboard** — on both the pi side and the browser side. It does not auto-reconnect, a browser refresh does not help, and `/reload` does not help; only launching a brand-new `pi` process reconnects.

Root cause (from reading `packages/extension/src/bridge.ts`): the `session_shutdown` handler calls `connection.disconnect()` **unconditionally**, ignoring the event `reason`. On a session **resume/switch**, pi fires `session_shutdown` (`reason: "resume"`) *before* the new `session_start`, so the bridge tears down its WebSocket — and nothing ever reconnects, because `connection.connect()` is only ever called once, at bridge init.

## Environment

- **pi-agent-dashboard**: 0.6.1 (bridge `@blackbelt-technology/pi-agent-dashboard`, server CLI `pi-dashboard` 0.6.1)
- **pi** (`@earendil-works/pi-coding-agent`): 0.81.1
- **Node**: v24.16.0
- **OS**: Linux
- **Dashboard config** (`~/.pi/dashboard/config.json`): `{ "autoStart": false, "tunnel": { "enabled": false }, "port": 8000, "piPort": 9999, "spawnStrategy": "headless" }`
- Server started manually with `pi-dashboard start`; bridge loaded via pi `settings.json` `packages`.

## Steps to reproduce

1. `pi-dashboard start` (server up on loopback).
2. Launch `pi` (fresh interactive TUI). → the session appears in the dashboard and is live; the bridge is connected. ✅
3. In the TUI, resume/switch to another session (`/resume` / the session selector).
4. → The session is dropped from the dashboard. ❌

## Observed (confirmed)

- After resume, **both** the pi side (the in-TUI connection indicator) **and** the browser show **disconnected**.
- A **browser page refresh does not** bring it back.
- **`/reload` inside pi does not** bring it back.
- Only a **fresh `pi` process** reconnects (it appears as a new session).

So this is a real WebSocket disconnect on the pi/bridge side (not a stale browser view), and it does not auto-recover.

## Expected

Per `docs/architecture.md`, resuming a session should re-register with the same `sessionId` and transition it `ended → active` with no manual action. The WebSocket should stay up across an in-process session switch.

## Root cause

`packages/extension/src/bridge.ts` — the `session_shutdown` handler:

```js
pi.on("session_shutdown", safe(async () => {   // note: no event arg → reason is ignored
  if (!isActive()) return;
  ...
  connection.send({ type: "session_unregister", sessionId });
  ...
  await new Promise((r) => setTimeout(r, 100));
  connection.disconnect();   // ← unconditional
}));
```

- `connection.disconnect()` runs on **every** `session_shutdown`, regardless of `event.reason`.
- pi fires `session_shutdown` with `reason ∈ {"quit","reload","new","resume","fork"}`. On an in-TUI **resume/switch**, the sequence is `session_shutdown(reason: "resume")` → `session_start(reason: "resume")`. The shutdown handler disconnects the socket (and cancels reconnect timers), and the following `session_start` → `handleSessionChange` only tries to `session_register` over the now-closed connection.
- `connection.connect()` is only called **once**, during bridge init (`bridge.ts` ~L2347). The `session_start` / resume path never re-establishes the connection, so it stays down.

This matches every observed symptom, including why a fresh process works (it runs the init path that calls `connect()`) and why `/reload` and a browser refresh do not.

## Suggested fix

In the `session_shutdown` handler, read the event `reason` and **do not disconnect on a session replacement** — only tear down the socket on a genuine teardown:

```js
pi.on("session_shutdown", safe(async (event) => {
  if (!isActive()) return;
  ...
  connection.send({ type: "session_unregister", sessionId });
  ...
  // Session replacement (new/fork/resume) keeps the same process + connection;
  // the following session_start re-registers via handleSessionChange. Only a
  // real teardown ("quit") should drop the socket. ("reload" re-inits + reconnects.)
  if (event?.reason === "quit") {
    await new Promise((r) => setTimeout(r, 100));
    connection.disconnect();
  }
}));
```

Alternatively, keep the disconnect but have the resume path (`session_start` with `reason ∈ {"new","fork","resume"}`) re-establish the connection (`connection.connect()` if not connected).

## Notes

- We are on pi 0.81.1; the develop CHANGELOG says 0.80.10→0.81.1 has no bridge-breaking changes, so this looks version-independent — `session_shutdown` has carried a `reason` since well before 0.81.

robertcsakany:
Implements OpenSpec change `fix-bridge-resume-disconnect`. Closes #393.

## Problem
Resuming/switching/forking a session inside the pi TUI permanently disconnected the bridge from the dashboard — no auto-reconnect; only a fresh `pi` process recovered.

## Root cause (spike-confirmed)
`session_start` calls `handleSessionChange(ctx)` **before** `connection.connect()`. `handleSessionChange` reads `ctx.cwd` — a guarded getter on pi's ExtensionRunner that **throws once the session is replaced**. The throw is swallowed by the handler's `safe()` wrapper, so `connect()` is skipped and the socket the preceding `session_shutdown` closed stays terminally down.

## Fix
Add `safeCwd(ctx)` (defensive read, falls back to `process.cwd()` on throw; mirrors the existing `readEventCwd` pattern) and use it for every `session_start`-path `ctx.cwd` read: `session-sync` `handleSessionChange`, `startGitPollTimer`, and the main-body register/git reads. With the getter guarded, `handleSessionChange` completes, `connect()` runs, and the buffered `session_register` flushes on the fresh socket.

## Tests
New L1 suite `bridge-resume-disconnect.test.ts` (real `handleSessionChange` + real `ConnectionManager` + fake WS):
- **X1** injects the `ctx.cwd` throw → red→green repro of #393.
- **F1–F4** resume/new/fork/reload end live + re-registered.
- **E1** quit tears down, cleanup ran. **X2** shutdown cleanup completes regardless of reason.

7/7 green; extension package green in isolation (1318 tests).

## Deferred
Manual **M1** (real in-TUI resume keeps the dashboard connected — the true #393 repro, backstop for real-path fidelity) → post-merge verification.

## Note for CI
Local full-monorepo `npm test` showed 56 failures **all** in packages this change doesn't touch (`pi-image-fit-extension` needs native deps absent locally; `browse-endpoint`/`search-files` pass in isolation, flaky under local parallelism) — proven pre-existing with the change stashed. Relying on CI's clean env as the authoritative gate.


robertcsakany:
Thanks @coderabbitai. Resolution:

- **`settle()` fixed delay → bounded polling** ✅ applied (`waitForConnected`, commit c022e143f) — honors the spec's "poll `connection.isConnected`" wording.
- **Move change out of `archive/`** — skipped: the `ship-change` workflow *intentionally* archives the change on ship; `openspec/changes/archive/YYYY-MM-DD-<name>/` is the correct terminal location. The active dir was correctly moved by `openspec archive`.
- **M1/V3 checkbox** — skipped: per the documented `ship-change` defer convention, `manual-only` tasks flip to `[x]` at ship with "deferred to post-merge verification" noted in both the commit and PR body. M1 (real in-TUI resume) is the flagged post-merge manual check.
- **Use real bridge handlers instead of the model-mirror** — skipped: the pure model-mirror + fake-WS is this repo's prescribed L1 pattern (cf. `bridge-shutdown-reset.test.ts`, and the change's `tasks.md`). `initBridge`'s `session_start`/`session_shutdown` handlers aren't unit-testable in isolation; the suite already exercises the **real** `handleSessionChange` (session-sync.ts) + **real** `ConnectionManager` against a fake socket.

robertcsakany:
@coderabbitai full review

PR #404, opened 2026-08-04
feat(openspec): model & effort selector in launch dialogs

robertcsakany:
Implements OpenSpec change `openspec-dialog-model-effort-selector`.

## Summary
Adds a shared **"Runs with"** run-config row (model + thinking-effort selectors) to the Explore, Propose, and New Change OpenSpec launch dialogs, so the user sets model/effort at launch without leaving the dialog. A **confirm-before-send gate** waits for the session to report the chosen `(model, effort)` before sending the prompt — closing the bridge concurrency race proven by the Task 1 spike (a naive `set_model` → `send_prompt` runs on the OLD model).

- `OpenSpecRunConfigContext` — session model/effort/models/favorites + setters, provided once at the app root; hook throws outside the provider (fails loud at all 3 mount sites).
- `useOpenSpecRunConfigRow` — the row, dirty-state disclosure, degraded (no model list) state, and the send gate (emit → wait for reported pair → send; timeout-and-notify fallback; cancel-on-unmount).
- Shared header anatomy across the three dialogs: icon tile, static title, Explore name chip + hint, persistent paste field note disclosing Cmd/Ctrl+Enter.
- Additive a11y on the reused `ModelSelector`/`ThinkingLevelSelector` triggers: `aria-haspopup`/`aria-expanded`/`aria-controls` + `role="listbox"` (also improves the composer).
- `connection.ts`: documents the intentionally-concurrent handler dispatch.

## Tests
New `OpenSpecRunConfig.test.tsx` (context throw, seeding, dirty, degraded, aria, all 4 gate paths) plus updated Dialogs/NewChange/SessionOpenSpecActions/MobileActionMenu specs. Full `npm test` green except 2 pre-existing environmental failures (docker-compose shim + a git tmp-bare-remote test) in `packages/server`/`scripts` — files this change never touches.

## Follow-up
Files `serialize-bridge-message-pump` (task 1.5) to close the bridge race for **all** callers (not implemented here).

QA/manual tasks (7.2–7.5 contrast/target/keyboard/reflow, 8.4 isolated run-check) deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

robertcsakany:
Thanks — addressed the actionable items for this change in the latest push (fieldset keyboard-disable, ARIA listbox/option correctness, stronger cancel test, spec Purpose, proposal signal note, and back-pressure/failure-isolation in the follow-up spec).

Intentionally not changing:
- **archive/ path**: this repo's `ship-change` workflow archives completed changes to `openspec/changes/archive/YYYY-MM-DD-<name>/` by design — that is the correct terminal location here, not a prohibited path.
- **`preserve-inline-terminal-transcript/*`**: those are a *different* change's planning artifacts that happen to sit in this branch's base history; out of scope for this PR.
- **App.tsx selected-session sourcing**: the run-config context sources from the selected session per design Decision 2 (documented); the composer is the primary/tested path.

robertcsakany:
Re: "wait for the complete selected pair" — the gate already does this correctly. `pending` stores a field **only when the user changed it**; the confirmation watcher requires `cfg.model === pending.model` (when model changed) AND `cfg.thinkingLevel === pending.level` (when level changed). When the user changed *both*, it waits for both. When the user changed only the model, it intentionally does not gate on the level — the prompt runs on the session's current level (which the row keeps in sync), which is the desired behavior. No partial-state fire.

The keyboard-disable (fieldset), ARIA listbox/option, proposal signal note, spec Purpose, and follow-up back-pressure/failure-isolation are all in the latest push.

PR #405, opened 2026-08-04
feat(update-pi-core-0-83-adopt-apis): bump pi 0.81.1→0.83.0 + feature-detected adoptions

robertcsakany:
Implements OpenSpec change `update-pi-core-0-83-adopt-apis`.

## Summary
Bumps the pi runtime pin `0.81.1 → 0.83.0` and adopts the genuinely-applicable new APIs behind runtime feature-detection (`minimum` stays `0.78.0`).

- **Pins:** server dep `^0.83.0`, `piCompatibility.recommended 0.83.0`, `docker/Dockerfile @0.83.0`, extension devDep `typebox ^1.3.7`, `pnpm-workspace.yaml minimumReleaseAgeExclude`, lockfile.
- **Coherence:** new `checkPiPinCoherence` gate in `verify-release-deps.mjs` (normalized 3-pin coherence, fails + names drift).
- **`ctx.scopedModels`:** `list_models` scope-aware via **non-empty** gate (empty-when-unscoped semantics), ref derived from the Model object; byte-identical fallback on absent/empty.
- **`"pending"` stop reason:** classifies as normal (in-progress); empty-actionable guard intact for genuinely idle turns; errors still win.
- **TypeBox 1.3.7:** removed-API audit (zero usage) + nullable-array validation on the real runtime validator.
- **Not applicable (recorded):** streaming `bash_execution_update` + bash session env (no RPC-bash path — dashboard uses `pi.exec`); `outputPad` (TUI padding setting, no web surface).

## Verification
- 37 new-scenario tests green (T1–T19: scopedModels, coherence, typebox nullable-array, pending, version-skew).
- Extension suite 1347/1347; server 3451 pass; scripts 125 pass — on the merged tree (pi 0.83.0 + typebox 1.3.10).
- Two pre-existing environmental failures (git-remote fixture, docker-compose interpolation) are unrelated to this change and require infra absent in the dev sandbox.

Planning went through 2 cross-model doubt-review cycles (F1–F6 corrections). Manual-only tasks (bash spike, outputPad no-op) recorded in specs.


robertcsakany:
@coderabbitai full review

robertcsakany:
Thanks — triage of the review findings:

**Applied:** the `piCompatibility.minimum` consistency finding. The requirement now states `minimum` is an **independent broad-support floor** that does not track the pinned/tested version (tests exercise the pinned `recommended`; `minimum` stays at the broad floor, at or above the bundled-extension peer-deps). Applied to both the live synced spec and the archived copy.

**Not applied — "move OpenSpec artifacts out of `archive/`" (2 threads):** this is a false positive. The guideline ("store artifacts under `openspec/changes/<name>/`, never under `active/` or `archive/`") governs where a change is **authored during planning**. Archiving is the documented **terminal ship step**: `openspec archive <change>` moves the change to `openspec/changes/archive/YYYY-MM-DD-<change>/` and syncs the delta specs into `openspec/specs/`. `develop` already contains sibling archive dirs from previously shipped changes (e.g. `archive/2026-08-04-openspec-dialog-model-effort-selector/` from #404). Reverting the archive would break the ship contract and leave the change permanently un-archived.

**Not applied — findings on `openspec-dialog-model-effort-selector` / `preserve-inline-terminal-transcript`:** those are other in-flight changes, not this one. They appeared only because this branch predated #404's archive; after re-merging `develop` and dropping the stale active copy, they are no longer in this PR's diff.

PR #407, opened 2026-08-04
feat(preserve-inline-terminal-transcript): keep inline terminal transcripts across PTY exit

robertcsakany:
Implements OpenSpec change `preserve-inline-terminal-transcript`.

## Problem

An inline terminal card whose shell exited **before** the user clicked ✕ froze with an empty transcript — the flagship repro is `ls`, then `exit`, then ✕. Separately, a card the user never touched left a blank "Terminal closed" tombstone behind, live and on every replay.

## What changed

**Transcript survives PTY exit.** Both exit paths (`pty.onExit` and the Windows kill fallback) now write a bounded, ephemeral-only tombstone *before* `entries.delete(id)`. Removal semantics are untouched. `releaseTranscript` is a **sticky, time-bounded** suppression flag (`RELEASED_TTL_MS` 60 s), not a delete — a count cap would defeat stickiness itself, letting a late `onExit` re-tombstone an already-closed card. Close is idempotent: `released` is checked **first**, before any liveness or tombstone lookup, so a two-browser double close can't emit a second, divergent transcript.

**Cap by serialized bytes, not code units.** `maxEventDataSize` is enforced in JSON bytes; a `string.length` cap diverges up to 6× on ESC-dense or CJK scrollback, so a "capped" transcript could still trip the ceiling and collapse `data` into a placeholder — destroying the `terminalId` the reducer keys on, *only* for non-ASCII/colorized users. `capTranscript` binary-searches the largest tail slice whose serialized size (marker included) fits. The budget is **derived** from the store ceiling (75 %), and one `eventDataCeiling` now feeds both the store and the derivation so they cannot drift.

**"Never interacted with" is server-side.** Two text-inspection rules were rejected in design: a line count inverts under a natural control-char strip (LF is C0 — every card would vanish) and deletes the `exit` repro; a whitespace rule deletes cards for users who pressed only arrows/Tab. Instead `sawInput` is set at the sole user-input `pty.write` sites, and the client's only predicate is `transcript === ""`. One predicate, so client and server cannot disagree; no ANSI-parsing code enters the client.

**Trim safety.** `inline_terminal_open`/`_close` join `ESSENTIAL_CHAT_EVENT_TYPES` — trimming one of a structurally-paired set would replay the card at the stream tail instead of its original position.

## Verification

- **133 L1 unit scenarios** — byte-boundary caps (incl. CJK/ESC byte-vs-unit proofs), boot config validation, retention bounds + ephemeral filtering, release-order independence, close idempotency, the oversize acceptance gate through a **real** `createMemoryEventStore` with production args, and reducer predicates (E17/E18 specifically pin that no text inspection was reintroduced).
- **11 L3 Playwright scenarios** green against the docker harness on its derived `dashboardPort` — flagship exit-then-close, untouched-card removal (incl. multi-line prompt), frozen-on-any-input, replay convergence, two-browser identity, live/dead-PTY reattach regressions, close→frozen latency.

## Notes for review

- **P1 deviation:** the plan asked for a 20-iteration p95; the spec asserts a single-shot latency bound instead (documented inline) because 20 iterations of 256 KB scrollback is prohibitively slow in CI.
- **Pre-existing, unrelated:** `tests/e2e/terminal.spec.ts` fails on unmodified `develop` at the default viewport — the composer falls under the `@[44rem]` container query, hiding the desktop terminal button. This PR's spec works around it (wider viewport + ⋯ overflow fallback); fixing that spec is left out of scope deliberately.
- `InlineTerminalCard` gains `data-testid="terminal-card"` + `data-terminal-state` — an untouched card is now *removed*, so absence is the assertion.


robertcsakany:
@coderabbitai full review

PR #409, opened 2026-08-04
feat(auth): allow config-file override of OAuth redirect base URL

Philogag:
buildRedirectUri now prefers auth.redirectBaseUrl from config.json over the active tunnel URL, falling back to http://localhost:<port>. Threaded through auth-plugin (incl. hot reload) and preserved by writeConfigPartial.


pabloxio:
@robertcsakany +1 on this PR as it would be a nice addition. Running the dashboard on a LAN hostname (e.g. `http://my-server:8000`) without a tunnel is a common setup.

robertcsakany:
## Review — `feat(auth): allow config-file override of OAuth redirect base URL`

Thanks for this — the deployment shape it fixes (dashboard behind a reverse proxy on a stable custom domain, no dashboard-managed tunnel) is real and currently has no supported answer, and the wiring is correct in the places that are easy to get wrong. In particular you threaded the override through **all three** call sites including `/auth/callback/:provider`; a change that only patched `/auth/start` would have produced a login that looks green right up until token exchange fails. `||` rather than `??` is also the right call and I'd like it to stay — with `??`, a config containing `"redirectBaseUrl": ""` would emit the relative `/auth/callback/github`.

I've captured the design + spec for this in an OpenSpec change (`config-override-oauth-redirect-base`) so the behaviour is recorded rather than living only in the diff. Findings below, in priority order.

### 1. The behaviour the PR exists for is untested (blocking)

`packages/server/src/__tests__/auth.test.ts` says it plainly:

> No tunnel runtime exists in tests, so `getTunnelUrl()` returns null and the localhost fallback is exercised.

So every new assertion is really *override vs localhost*. The precedence that justifies the feature — **override beats an ACTIVE tunnel** — has zero coverage. `getTunnelUrl` is a plain named import in `auth.ts`, so `vi.mock("../tunnel/tunnel.js")` closes it:

```ts
const tunnelUrl = { value: null as string | null };
vi.mock("../tunnel/tunnel.js", () => ({ getTunnelUrl: () => tunnelUrl.value }));

it("override beats an active tunnel", () => {
  tunnelUrl.value = "https://abc.share.zrok.io";
  const uri = buildRedirectUri("github", 8000, "https://pi.example.com");
  expect(uri).toBe("https://pi.example.com/auth/callback/github");
  expect(uri).not.toContain("zrok");
});
```

I've written the full suite (6-row precedence table, normalization, route-level `Location` assertions via `fastify.inject()`, a token-exchange echo assertion, and hot-reload via `_reloadAuth`) plus a container-level Playwright spec. Verified **11 failures on `develop`, 19/19 green with this patch applied** — so it is not a vacuous gate. Happy to push it here or land it alongside, whichever you prefer.

### 2. Misconfiguration is silent (should fix before merge)

Any string is accepted: `pi.example.com` (no scheme), `https://x?tenant=a`, `ftp://…`. The operator's observable symptom is a provider-side `redirect_uri_mismatch` or a login loop, with **nothing in the server log**. Suggested: validate at registration and at every reload (absolute `http(s)`, no query, no fragment) and log one warning naming `auth.redirectBaseUrl` and the value — but keep using the value, so a typo does not silently degrade into "my setting does nothing".

No security concern otherwise: the value is operator-owned config, never request-derived, and the provider's own `redirect_uri` allowlist is a second gate. Worth stating so future readers don't have to re-derive it.

### 3. Hot reload only works if auth booted with a provider (document)

`registerAuthPlugin` returns early on an empty provider registry **before** installing `_reloadAuth` and before registering any `/auth/*` route. So on a dashboard that booted with no providers, adding `redirectBaseUrl` via `PUT /api/config` is inert until a restart. Pre-existing, not caused by this PR — but it's exactly the state a new operator is in, so it belongs in the docs.

### 4. Two sources of truth for "my public origin" (scope note)

After this change OAuth uses the override while pairing QR codes, `/api/tunnel/endpoints`, and "Accessible at" still use the tunnel URL. That's a defensible slice boundary — I'm not asking you to widen the PR — but it should be stated in `docs/architecture.md` rather than discovered. The forward-compatible shape is a later top-level `publicBaseUrl` feeding every public URL, with `auth.redirectBaseUrl` staying as a narrower higher-precedence override; nothing here blocks that.

### 5. Smaller items

- **Docs**: per this repo's protocol, config-reference changes go in `docs/architecture.md`; the PR updates only the three `AGENTS.md` rows (which are good and accurate).
- **CHANGELOG**: no `[Unreleased]` entry, so the feature would ship invisible in the release notes.
- **No UI**: `writeConfigPartial` now accepts the key but nothing in `SettingsPanel.tsx` sends it — hand-edit-only. Fine as a slice; worth an explicit note.
- **Docs should state** that the operator must also register the same URI with the OAuth provider.
- Trailing-slash stripping now applies to the tunnel and localhost bases too (a no-op today, since neither carries one) — already noted in your `auth.ts.AGENTS.md` row. 👍

robertcsakany:
Pushed four follow-up commits onto this branch (`maintainerCanModify`) rather than opening a competing PR — your commit stays first and keeps authorship.

| commit | what |
|---|---|
| `docs(openspec)` | the change captured as spec: proposal, design (D1–D6), spec deltas for `oauth-authentication` + `shared-config`, test plan |
| `feat(auth)` | `warnOnInvalidRedirectBase()` — finding #2. Warns at registration and every reload, **still uses the value** so a typo fails visibly rather than becoming a silent no-op |
| `test(auth)` | finding #1 — the override-beats-tunnel precedence, route-level `Location` assertions, the token-exchange echo, hot reload, plus a container-level Playwright spec |
| `docs(auth)` | `docs/architecture.md` config reference + the two traps, and the missing CHANGELOG entry |

Verification: the new vitest file was run **red on `develop` (11 of 19 failing, and it did not type-check without your 3-arg `buildRedirectUri`)** and **green on this branch (37/37)**. `tsc --noEmit` clean. Biome violation counts on `auth.ts` / `auth-plugin.ts` are byte-identical to `develop`, so the quality ratchet is untouched.

Two notes:

- `tests/e2e/oauth-redirect-base.spec.ts` has **not** been executed yet — it needs the Docker harness. It restarts the server (the auth plugin registers no `/auth/*` route when the boot-time provider registry is empty) and it seeds `auth.bypassUrls: ["/"]` plus restores in `afterAll`, because requests from the Playwright host arrive at the container as non-loopback and would otherwise auth-gate every spec that runs after it. Please treat its first CI run as the real verification.
- Findings #3 (docs) and #4 (scope) are addressed as documentation only — the pairing-QR / `/api/tunnel/endpoints` split is now written down in `architecture.md` and named in `design.md` D5 as a deferred `publicBaseUrl` slice, not fixed here.

Shout if you'd rather I split any of this out into a separate PR.

robertcsakany:
## Design mockups for the follow-up scope

**This is context, not a change request on this PR.** Nothing below asks you to widen `#409` — your change stays exactly as scoped, and the review findings I raised earlier are already addressed on the branch. I'm posting these so the design work that grew *around* your PR is visible rather than landing later as a surprise.

Sections 1–4 cover a gateway-URL action; 5–9 cover the surfaces the follow-up needs (drift repair, settings, provider deletion, diagnostics, first run). Every screen is rendered from the dashboard's own theme tokens, in both themes.

<details>
<summary><b>Why this exists</b> — the two-sources-of-truth note from my review</summary>

Review finding #4 said OAuth would use `auth.redirectBaseUrl` while pairing QR / `/api/tunnel/endpoints` / "Accessible at" keep using the tunnel URL, and named a future top-level `publicBaseUrl` as the forward-compatible shape.

**That premise turned out to be wrong**, and it is worth recording: `pairing.publicBaseUrls: string[]` already exists (`packages/shared/src/config.ts`), is already operator-editable in the Gateway UI, and already drives every non-OAuth public URL. A new scalar `publicBaseUrl` would have been a *third* answer, not a unification.

The resolution keeps your key exactly as you designed it. `publicBaseUrls` unifies the pairing/endpoint surfaces only; `auth.redirectBaseUrl` stays the OAuth-only override — justified by **arity**, not preference: `publicBaseUrls` is a list of every address the dashboard answers on, while an OAuth `redirect_uri` must be a single pre-registered scalar. Inferring one from the other was designed, reviewed, and then **removed** as unsound.

</details>

---

### 1 · Gateway list — every row states its status

Status is glyph + word, never colour alone (WCAG 1.4.1). Each row explains *why* it is not OK, in the operator's terms rather than key names.

s1-dark

<details><summary>Light theme</summary>

s1-light

</details>

---

### 2 · Add gateway — the scheme decides what is possible

An `http://` gateway cannot ride the pairing QR (the TLS gate) and cannot be an OAuth callback (providers reject non-TLS), so trusted-network becomes mandatory. Ineligible options are disabled **with the reason inline**, rather than accepted and rejected after save.

s2-dark

<details><summary>Light theme</summary>

s2-light

</details>

---

### 3 · Remove — reverses exactly what add wrote

Removal is driven by a provenance record, so it reverts only what the action created and leaves operator-authored values alone. The confirmation lists every field.

s3-dark

<details><summary>Light theme</summary>

s3-light

</details>

---

### 4 · Fix — restores the delta, and says what it restores

The underlying keys stay independently editable, so drift is expected. Fix reconciles to the record; it never re-runs "add".

s4-dark

<details><summary>Light theme</summary>

s4-light

</details>

---

### 5 · Ineligible — Fix must ask again, not guess

When a recorded choice stops being legal, the system must not silently pick a replacement.

s5-dark

<details><summary>Light theme</summary>

s5-light

</details>

---

### 6 · Settings ▸ Security — shows the value in force

Not a write-only box: it reports which value actually won the precedence chain, and links to the gateway action rather than duplicating it.

s6-dark

<details><summary>Light theme</summary>

s6-light

</details>

---

### 7 · Removing the last OAuth provider — a lockout, not a disable

**The one I would most like a second opinion on.** `_reloadAuth` only mutates `authState`; the `onRequest` gate is installed after the boot-time early-return and there is no `removeHook`. So deleting the last provider at runtime leaves the gate **enforced with no way to sign in** — the opposite of the boot-time "no providers ⇒ auth disabled" case. A remote operator whose JWT expires is locked out until restart. Hence plain-language consequence + an explicit acknowledgement, not a second "Are you sure?".

s7-dark

<details><summary>Light theme</summary>

s7-light

</details>

---

### 8 · Diagnostics — which tier actually won

The chain has multiple tiers, so "is my value valid" is the wrong question. Losing tiers are struck through. Deliberately reachable over loopback and from the terminal, because a remote operator with broken OAuth cannot authenticate to reach a gated web page — the tool has to work inside the failure it diagnoses.

s8-dark

<details><summary>Light theme</summary>

s8-light

</details>

---

### 9 · First run — the same component, not a copy

The setup step embeds the same dialog the operator meets later (Jakob's Law); an e2e assertion pins that the markup is identical.

s9-dark

<details><summary>Light theme</summary>

s9-light

</details>

---

### Verification

Contrast was computed rather than eyeballed: **54 checks, 0 failures**, minimum **4.65:1** dark / **4.54:1** light. Captured at 375 / 768 / 1440 with no overflow, 44 px touch targets on coarse pointers, and destructive actions exposed as `disabled` until acknowledged.

Two pre-existing issues surfaced while grounding against real components — **neither is from this PR**, both worth their own fix:

1. `GatewayEndpoints.tsx:154,163` uses `var(--border)` and `var(--danger,#ef4444)` — **neither token is defined** in `index.css`; both silently ride their fallback. The defined names are `--border-primary` and `--accent-red`.
2. White on `--accent-blue` is **3.68:1** — below WCAG AA anywhere it fills a primary button.

Sources live in `openspec/changes/config-override-oauth-redirect-base/mockups/` (`index.html` + `ui-plan.md`, which cites the rule behind each decision).

PR #411, opened 2026-08-05
feat(drag-folders-across-workspaces): drag folders into, out of, and between workspaces

robertcsakany:
Implements OpenSpec change `drag-folders-across-workspaces`.

## What

Lets a user change a directory's workspace membership by dragging:

- a **pinned directory** or a **workspace folder** INTO a workspace — header = append, inner slot = positional
- a workspace folder OUT to the pinned tier (ejected **and** pinned)
- a folder BETWEEN two workspaces

One server round-trip per gesture, no optimistic client state.

## Why it needed three walls down at once

The sidebar shipped three drag gestures, all *ordering* only. Three separate guards kept membership changes out — `sameTypeClosestCenter`, the `activeType !== overType` early return in `handleDragEnd`, and the cross-workspace guard in `resolveWorkspaceFolderReorder`. Removing only one leaves the feature inert, so all three change together.

## Load-bearing details

- **Validate before mutate.** The store resolves `toWorkspaceId` *before* any detach — the naive "detach from all, then insert" leaves the folder in **zero** workspaces on a stale id.
- **`index` clamped to `[0, len]`**; a negative index makes `splice` count from the end. A non-integer `index` is rejected outright — `splice(NaN, 0, x)` coerces to a *front* insert where the contract promises an append.
- **Broadcast order on eject is observable.** `pinned_dirs_updated` goes out *before* `workspaces_updated`; the two arrive as separate frames in separate macrotasks, so the reverse order leaves the folder in neither list for one render frame and it visibly flashes out of the sidebar.
- **Eject reuses the handler-level pin path**, not the bare `pinDirectory` array push, so an ejected folder gets the same on-disk session discovery a hand-pinned one does.
- **Own-header drop is a no-op** — without that rule it would detach-and-re-append, silently jumping the folder to the bottom.
- **Unmatrixed drag types keep today's same-type wall**, never a weaker closestCenter-over-all.
- **`MeasuringStrategy.Always`** is required: spring-load mounts folder droppables *mid-drag* and the default `Optimized` strategy would not remeasure them.

## Tests

- **46 new L1** — store contract (clamping, single-membership, validate-before-mutate), handler guards + broadcast ordering, pure resolvers incl. a p95 perf bound over ~420 droppables, component-level message dispatch, header droppable wiring.
- **5 new L3 Playwright** (`tests/e2e/folder-membership-drag.spec.ts`) for what jsdom cannot express: spring-load dwell, mid-drag remeasurement, empty-tier eject geometry, dwell-timer jitter resistance, and two-client convergence. Verified green against a freshly built docker harness.

Full suite: **11,572 passing, 0 failing**; `tsc --noEmit` clean repo-wide.

## Note on the first commit

`fix(kb): unbreak dox-triage child_process ban + test types` fixes a **pre-existing** breakage on `develop` (from `4904ef094`), not something this change introduced — `packages/kb/src/dox-triage.ts` tripped the repo-wide `no-direct-child-process` invariant and its test had 3 tsc errors. Kept as a separate commit. `packages/kb` is deliberately self-contained, so it takes the same per-line `ban:child_process-ok` opt-out its sibling `sources.ts` already uses rather than growing a dependency.

QA/manual tasks (4-theme visual pass) deferred to post-merge verification.


robertcsakany:
Thanks — went through each finding against the code. One applied, one rejected with evidence, two deferred as out-of-scope.

### ✅ Applied — `test-plan.md` E20 (case variance)
Correct. `canonicalizePath` normalizes then `realpath`s, and `realpath` does not fold case on a case-sensitive filesystem. The implemented test already asserted only the trailing-separator half; the plan text overstated it. Scoped the row in 7a0079241.

### ❌ Rejected — `directory-handler.ts` "unpin after a successful workspace move"

The premise ("appears in both tiers") does not hold, and the change would regress a shipped, spec-tested invariant. Four independent checks:

1. **No double-render.** `SessionList.tsx:428-434` already strips workspace-claimed folders out of the pinned tier — the code comment is literally *"strip out workspace-owned folders so they don't double-render"*:
   ```ts
   const claimed = new Set<string>((workspaces ?? []).flatMap((w) => w.folders));
   return pinnedGroups.filter((g) => !claimed.has(g.cwd));
   ```
2. **It would diverge from the shipped sibling.** `handleAddFolderToWorkspace` (the existing path for the same state transition) does not unpin either. Unpinning only on the *drag* path would make two routes to identical state behave differently.
3. **It would break an existing test.** `preferences-store.test.ts:399` — `"adding folder does NOT touch pinnedDirectories"`.
4. **It contradicts an explicit, recorded design decision.** `design.md` Risks: *"A pinned directory dragged into a workspace stays in `pinnedDirectories` … **Accepted** — pinning and membership are orthogonal by design and this exactly matches shipped `add_folder_to_workspace`. Changing it would be a separate behavior change."*

Pinning ("keep this visible even with no live sessions") and membership ("which workspace owns it") are deliberately independent axes. Making membership silently destroy the pin is a product change that belongs in its own proposal, not this one.

### ⏭️ Deferred — the two `packages/kb/src/dox-triage.ts` findings

Both are real observations but they target code from `4904ef094`, not this change. This PR touches that file only to unbreak `develop`: a one-line `ban:child_process-ok` marker (the repo-wide `no-direct-child-process` invariant) plus 3 type-only test fixes. Widening it into `parseRows` DOX-section scoping and the `mkdirSync` fix would put unrelated behavior changes in a sidebar-drag PR. The `mkdirSync(dirname(stalenessFile), { recursive: true })` one looks like a genuine `ENOENT` crash and is worth its own PR.

PR #412, opened 2026-08-05
feat(bridge): serialize the inbound message pump

robertcsakany:
Implements OpenSpec change `serialize-bridge-message-pump`.

## Problem

`ConnectionManager.ws.onmessage` invoked the async `onMessage` handler **without `await`**, so every inbound frame started an independent task. A `set_model` handler yields at its first `await` and a `send_prompt` arriving in the next socket tick runs to completion during that yield — submitting the prompt on the **old** model. Silent, and produces a run that looks correct. (Proved by the `openspec-dialog-model-effort-selector` spike, which closed the race only for its own dialogs via a client-side gate.)

## Change

Inbound messages now go through a single drain loop that awaits each handler to completion.

- **Immediate lane** — `prompt_response`, `server_restarting`, `kill_process` bypass the queue. Each is incapable of invalidating a message queued behind it (correlated by request id / lifecycle signal / pgid-keyed).
- **`abort` stays serialized.** Bypassing it would let it run *ahead* of the `send_prompt` it cancels and silently lose the cancellation. Unknown types default to the serialized lane.
- **Bounded inbound queue** — `maxInboundQueue` (default 1000), separate from the outgoing `maxBufferSize` ring; drop-newest so the accepted prefix keeps its ordering guarantee.
- **Separate counters** — overflow refusals vs routine disconnect discards (one counter would let reconnect churn mask the overflow signal). Overflow reaches `/api/health` via `ProcessMetrics.refusedInboundFrames`.
- **Epoch-bound drain loop** — teardown clears the queue and releases the running-guard *immediately*, so an uncancellable in-flight handler cannot stall the replacement connection; the superseded loop retires without dispatching.

## Tests

19 L1 scenarios in `packages/extension/src/__tests__/connection-inbound-pump.test.ts`, folded 1:1 from `test-plan.md` (E1–E11, P1, X1–X7): ordering, cancellation-not-reordered, immediate-lane membership (positive + negative), BVA around the cap, drop-newest prefix preservation, failure isolation (reject + sync throw), reconnect/backlog/in-flight lifecycle, counter separation, warn rate-limiting, and an O(n²) shape guard.

## Notes

- The client-side confirm-before-send gate is **kept** as belt-and-suspenders (task 3.1).
- Known pre-existing failures on `develop`, unrelated to this PR and reproduced on a clean tree: `bus-client/send-types` (`tsc` ENOENT) and 3 `image-fit` files (jimp default import); plus `move_folder_to_workspace` tsc errors from #411.
- Pre-existing gap filed as follow-up, not fixed here: `abort` does not cancel an in-flight `handleBashCommand` child (`kill_process` is the only working path).


robertcsakany:
@coderabbitai full review

PR #413, opened 2026-08-05
feat(plugin-settings-pages): one settings page per plugin, tab inert

robertcsakany:
Implements OpenSpec change `plugin-settings-pages`. Archives + syncs specs.

## What changes

Every `settings-section` claim now renders on its owning plugin's page at `/settings/plugins/<id>`, under host-owned chrome the plugin cannot suppress. `SettingsSectionByPluginSlot` becomes the single render path and **gains intent consumption**, so intent-driven and JSON-Schema-descriptor contributions survive the flip. `SettingsSectionSlot` is inert, its 15 mounts are gone, and `forTab` is deleted. `claim.tab` is still accepted but read by nothing; the validator no longer rejects unknown values.

The nav rail lists enabled plugins that contribute settings (keyed on `enabled`, not `loaded` — a failed plugin is exactly when you need its page). The Save Bar names every dirty page and labels plugin pages `Plugins › <name>`. The host assigns each draft source its page id (`plugins/<id>`) from the hook, not the registry — an ancestor closure cannot read a descendant's context.

`PluginsSection.tsx`: 634 → 210 lines (shared internals extracted to `plugin-row-parts.tsx` + `usePluginToggle`).

## Two things worth a reviewer's eyes

**1. A latent bug this surfaced.** `App.tsx`'s settings route was still `/settings/:page?`, so *every* deep link to `/settings/plugins/<id>` fell through to the dashboard root — the panel never mounted. Both route sites now carry `/settings/:page?/:sub?`, with a lint test pinning it.

**2. Enable-state is desired state, not runtime state.** `POST /api/plugins/:id/toggle` writes config and returns `restartRequired: true`, but `/api/plugins` and `/api/health` both serve `getPluginStatusStore()` — the snapshot captured at server boot — so **neither reflects a toggle until restart**. The design assumed `usePluginEnabledSet` re-filtered live; verified against the harness, it does not. `usePluginList` therefore keeps a desired-state overlay, seeded from `GET /api/config`.plugins (survives reload) and updated from the `plugin_config_update` payload. `status.loaded` is left alone as runtime truth, and the plugin page now carries the restart-required banner so an empty body is never unexplained.

The overlay is cosmetic-only by construction: it cannot mount a disabled plugin's component, because `SettingsSectionByPluginSlot` renders claims from the boot-time registry and gates intents on the server-authoritative `registry.isPluginEnabled`.

## Tests

41 automated scenarios from `test-plan.md`:

- **L1 (vitest)** — route resolution incl. two consecutive optional wouter segments, nav membership decision table, claim/intent ordering, intent-only plugins, disabled-plugin intent filtering, draft page-id rewrite, and the repo lints. Full suite: **11643 passed**.
- **L3 (Playwright, docker harness)** — **18/18**, idempotent across 3 runs. The harness now seeds two plugin fixture states (`e2e-broken` → load error, `e2e-needs-req` → unmet requirement) so the error/requirement paths are real tests, not skipped stubs.

A cross-model adversarial review caught two defects that are fixed here: intent-only plugins were gated out of both the nav and the page (silently dropping the exact contribution form the change exists to preserve), and a just-enabled plugin rendered an unexplained empty body. Both have tests verified red-without-fix.

## Note for the archive

Per an explicit call during implementation, `design.md`'s D6 paragraph was left as written even though its *mechanism* claim about `usePluginEnabledSet` is factually wrong for this build. The correct account lives in `docs/architecture.md` and the `usePluginToggle.tsx` tree row.

QA/manual tasks (§11, `test-plan: manual-only`) deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

robertcsakany:
Both remaining threads are addressed or intentionally declined:

**`RolesSettingsSection.tsx:516` — duplicate headers in the other plugin bodies.** Fixed in d56c3243. You were right that I only de-headered `roles`; `subagents`, `flows`, and `automation` each still rendered their own `<section>`/`<h3>` and double-headered under the host chrome. All three now use the same slim-body pattern, verified in the browser against the docker harness.

**`tasks.md:133` — deferred manual checks marked complete.** Declined, by project convention rather than oversight. The `ship-change` skill marks `manual-only` tasks done for **post-merge** verification — that is its documented step 1 defer rule, the section header reads "Manual verification (deferred post-merge)", each row carries an explicit `(test-plan: manual-only)` tag, and the PR body states the deferral. Flipping them back would block the archive gate on checks that cannot run pre-merge.

PR #422, opened 2026-08-05
feat(project-scope-disable-global-resources): write the pi-standard form for each resource origin

robertcsakany:
Implements OpenSpec change `project-scope-disable-global-resources`.

## The defect

Disabling a globally-defined skill for one project silently did nothing. The write picked a force-exclude relative to the project's own `.pi`, which matches nothing for a resource pi resolves from `~/.pi/agent`. The dashboard reported success and pi kept loading the resource.

## What changed

**Origin classification (`resource-origin.ts`, new).** Longest-prefix path match against the candidate base dirs — never `metadata.scope`/`source`/`baseDir`, which the write itself mutates (a metadata-keyed classifier could not recognise on re-enable the resource it re-declared). Longest-prefix also survives `cwd === $HOME`, where `<cwd>/.pi` is a strict ancestor of `~/.pi/agent`.

**Four pi-standard write forms (`resource-activation-toggle.ts`, rewritten).**

| origin | project-scope form |
|---|---|
| loose under `<cwd>/.pi` | `-<rel to .pi>` |
| loose under an `.agents` base | `-<rel to that base>` |
| package-contributed | `{ source, autoload: false, <type>: ["-<rel to root>"] }` |
| loose under a global base | the resource's own **file** as a `~` entry + an anchored glob exclusion |

- `autoload: false` is mandatory and destructive to omit — without it pi resolves at project scope, misses the user install path, and drops the package's whole contribution.
- The delta form is **project-scope only**; at global scope a second same-scope entry is discarded by pi's dedupe, so global keeps mutating in place.
- Package entries are matched by **normalised identity** (npm without version, git across SSH/HTTPS), not raw source string.
- The **file** is re-declared, never a directory: prompts, themes and flat `.md` skills have the shared root as their directory.
- The exclusion is an anchored glob — an absolute path is machine-local, and a `~` pattern is never expanded and therefore inert everywhere.

**Round-trip + safety.** Re-enable strips the exclusion and writes nothing in its place (a `+` never round-trips). Existing exclusions are stripped as an equivalence class **by exact spelling**; a user's broad glob is never evaluated and never removed, and a `+` force-include is never touched. Ownership of plain entries lives in `~/.pi/dashboard/resource-entry-ownership.json`, so the settings file stays pi-standard, each toggle is one settings write, and ownership stays machine-local.

**Trust gate (`resource-toggle-trust.ts`, new).** An untrusted folder now returns a trust prompt instead of a false success. `always` proceeds *without* recording (a user who chose it wanted fewer prompts, not durable trust in every folder they toggled); a refusal recorded for an *ancestor* falls through to the prompt, as pi's own dialog allows. `POST /api/resources/trust` accepts only an option **id**, re-derives the updates server-side, **and** requires an outstanding challenge raised by a real toggle — so a caller can choose neither the updates nor the path.

**Loud failures.** An unparseable `.pi/settings.json` (pi silently skips the write) now returns 409; a resolver throw returns 500 and writes nothing. The false "JSONC-preserving" claim in the module header and the docs is corrected — pi does a whole-file `JSON.parse`/`JSON.stringify` round trip.

**Client.** The folder surface now toggles at **local** scope (it was writing at the resource's own global scope — the cross-scope disable never happened from the UI); a disabled global row stays where the user acted and gains a *folder controls activation* badge; failures surface with server-refusal vs never-reached-the-server distinguished; the repository-wide blast radius is stated (the file is tracked and shared, each toggle is a whole-file diff).

## Tests

- 43 `resource-activation-toggle` + 14 `resource-activation-routes` + 10 `ResourceCard` unit tests. Behavioural assertions read **observed activation** from pi's own resolver rather than pattern text, so a semantics drift in pi fails loudly.
- 5 Playwright specs (`resource-activation-trust.spec.ts`, test-plan F1–F5) green against the docker harness.
- Findings from an adversarial security pass are folded in (trust-endpoint grounding, inherited-ancestor refusal, ownership recorded only after a successful flush).

## Rider

`tests/e2e/helpers/index.ts::pinDirectory` targeted the retired `PinDirectoryDialog` and never matched the new `AddFoldersDialog`, breaking the pin step for the ~61 specs that use `ensureGitSession`. Fixed here.

**Known, pre-existing, NOT fixed here:** those specs still fail at the *next* step — clicking "New Session" leaves the UI on "Starting new session…" and the server logs no spawn attempt at all. Untouched by this PR and worth its own investigation.

## Deferred

QA/manual tasks (9.7, 10.1, 11.1–11.7) are deferred to post-merge verification.


robertcsakany:
### CodeRabbit disposition

CodeRabbit is rate-limited and will not re-review, so recording the disposition here. Fixed in 532bf05 + 4d0df63:

**Accepted — real bugs, each with a regression test**

| Finding | Fix |
|---|---|
| Delta-removal deletes a user-authored global `{source, autoload:false}` entry | Only splice at project scope. This module never *creates* a delta globally (an absent package 404s), so such an entry is always the user's. Test: `never removes a user-authored autoload:false entry at global scope`. |
| Global-loose exclusion used the wrong evaluation base for `~/.agents` | `evaluationBase` now returns the resource's own base at global scope. Test: `strips a stale exclusion for a ~/.agents resource at global scope`. |
| Inherited ancestor refusal + `defaultProjectTrust: always` proceeded | Prompt instead — pi resolves the inherited refusal and would ignore the file, reinstating the false success this gate removes. Test: `does not let defaultProjectTrust:always proceed over an inherited refusal`. |
| Ownership-store write failure swallowed | `save`/`recordOwnedEntry`/`clearOwnedEntry` return a boolean; the toggle logs the consequence. The toggle itself still succeeds — the settings write landed and the resource *is* in the requested state. |
| `worktree-init-feedback.spec.ts` referenced the removed `pinDirectoryDialog` key | Defers to the shared `pinDirectory`. Good catch — this was a real break from the helper change. |
| `trustOptionsFor` paths not canonical | Canonicalised, matching pi's trust-store key normalisation. |
| `aria-modal`, MD056 pipes, spec's global package branch, archived task refs, caveman-style docs pass | All applied. |

**Rejected, with evidence**

- **scp-style git sources in `packageIdentity`.** pi's own `parseSource` treats a bare `git@host:path` as a *local* path (`isLocalPath` returns true — no `git:`/`ssh:`/`https:` prefix). Verified against the pinned dist:
  ```
  "git@github.com:acme/tools.git"      -> local:<baseDir>/git@github.com:acme/tools.git
  "git:git@github.com:acme/tools.git"  -> git:github.com/acme/tools
  "https://github.com/acme/tools.git"  -> git:github.com/acme/tools
  ```
  The mirror already matches pi exactly. "Fixing" it would make the dashboard disagree with the resolver it delegates to.

- **Suites mutate the real `$HOME/.pi/agent`.** They do not. `packages/server/vitest.config.ts` wires `setupFiles: shared/src/test-support/setup-home-perfile.ts`, which assigns `process.env.HOME` to a fresh `mkdtemp` **before** any test module loads (so `AGENT_DIR`, captured at import, is already isolated), and `globalSetup: setup-home.ts` aborts the whole run if `HOME` ever equals the real user home.

- **Move artifacts out of `archive/`.** `openspec/changes/archive/` is the correct post-ship location; `AGENTS.md` prohibits `active/`/`archive/` for *in-flight* changes only. This one is archived because it shipped.

- **Focus trap in `ResourceTrustDialog`.** `aria-modal` added. A focus trap would be a new pattern here — every existing dialog in this repo (e.g. `FirstLaunchDisplayModal`) has the same Escape-only behaviour, so introducing it in one dialog is worth its own change rather than a drive-by.

PR #426, opened 2026-08-06
fix: support Node 26 and build the client under --omit=dev (#357)

### LINKED ISSUE #357: `pi install` fails: Node 26 engine restriction, then client `prepare` requires omitted devDependencies
## Summary

Installing from the git repository through Pi fails:

```bash
pi install git:github.com/BlackBeltTechnology/pi-agent-dashboard
```

There are two independent blockers:

1. The package declares Node 26 unsupported (`"node": ">=22.19.0 <26"`), so npm aborts on Node 26.5.0 with `EBADENGINE`.
2. After bypassing that check to investigate further, the same install path fails because Pi invokes `npm install --omit=dev`, while the client workspace's `prepare` script runs a Vite build that requires packages declared only in `devDependencies`.

The second blocker means the package cannot currently be installed through Pi's production-style install path, independent of whether Node 26 support is added.

## Reproduction

```bash
pi install git:github.com/BlackBeltTechnology/pi-agent-dashboard
```

## First failure: Node 26 engine restriction

```text
npm error code EBADENGINE
npm error engine Unsupported engine
npm error notsup Not compatible with your version of node/npm: @blackbelt-technology/pi-agent-dashboard@0.5.4
npm error notsup Required: {"node":">=22.19.0 <26"}
npm error notsup Actual:   {"node":"v26.5.0","npm":"11.17.0"}
Error: npm install --omit=dev failed with code 1
```

I understand this is currently declared unsupported. Since Pi and my other extensions work on Node 26, could you please validate Node 26 and either remove/raise the upper bound or document the compatibility issue requiring it?

## Second failure: client build cannot run under `--omit=dev`

To isolate the next failure, I ran:

```bash
cd ~/.pi/agent/git/github.com/BlackBeltTechnology/pi-agent-dashboard
npm install --omit=dev --engine-strict=false
```

It fails with:

```text
npm error path .../pi-agent-dashboard/packages/client
npm error command failed
npm error command sh -c node scripts/vite-build.mjs build && node scripts/precompress.mjs
npm error Error: Cannot find module 'vite/package.json'
npm error Require stack:
npm error - .../pi-agent-dashboard/packages/client/scripts/vite-build.mjs
npm error code: 'MODULE_NOT_FOUND'
```

A full install succeeds:

```bash
npm install --engine-strict=false
```

## Root cause

`packages/client/package.json` runs its build during installation:

```json
"prepare": "node scripts/vite-build.mjs build && node scripts/precompress.mjs"
```

However, the build needs packages omitted by Pi's `npm install --omit=dev`:

- `tsx` is imported by `packages/client/scripts/vite-build.mjs`, but is a root `devDependency`.
- `vite` is resolved by that same script, but is a client `devDependency`.
- `@vitejs/plugin-react` and `@tailwindcss/vite` are imported by `packages/client/vite.config.ts`; both are client `devDependencies`.

So moving only `vite` to `dependencies` would not completely fix the installation path.

## Expected behavior

`pi install git:github.com/BlackBeltTechnology/pi-agent-dashboard` should complete successfully without needing the user to install development dependencies manually.

## Suggested fixes

Either:

1. Ship prebuilt client assets and avoid running the client build from `prepare` during git/consumer installation; or
2. If building during installation is intentional, move all direct build-time requirements (`tsx`, `vite`, `@vitejs/plugin-react`, and `@tailwindcss/vite`) out of `devDependencies`.

## Environment

- `@blackbelt-technology/pi-agent-dashboard`: 0.5.4
- Pi install source: `git:github.com/BlackBeltTechnology/pi-agent-dashboard`
- Node.js: v26.5.0
- npm: 11.17.0
- OS: macOS

## Related issues checked

I found no existing issue for `EBADENGINE`, `omit=dev`, or this missing `vite/package.json` failure.

robertcsakany:
Implements OpenSpec change `fix-pi-install-node26-and-omit-dev-build`. Fixes #357.

## Why

`pi install git:github.com/BlackBeltTechnology/pi-agent-dashboard` fails at two independent points, both still present in released `0.6.1`:

1. **Node 26 `EBADENGINE`** — root `package.json#engines.node` caps `>=22.19.0 <26`, and pi installs with engine-strict, so the install aborts before anything runs.
2. **Client `prepare` build under `--omit=dev`** — pi runs `npm install --omit=dev`, which drops `devDependencies`; the web client's `prepare` Vite build then dies with `Cannot find module 'vite/package.json'`.

The published npm path was never affected (it ships a prebuilt client and runs no `prepare`).

## What changed

- **Engines cap raised one major, `<26` → `<27`.** Three sites: root `package.json`, `packages/shared/src/node-version.ts` (`major >= 27`), and the hardcoded `Required:` range string in `packages/server/src/auth/node-guard.ts` — the third being a documented hole in the lockstep contract, now closed by a repo-lint. Node 27+ stays refused.
- **Node 26 added to the CI smoke matrix** (`_smoke.yml` `standalone-install-smoke-linux`, bookworm-slim + alpine). The Node 26 legs deliberately run **without** the `--config.engine-strict=false` override the other legs carry, so the `EBADENGINE` half of #357 is regression-tested rather than assumed.
- **Client build deps moved to runtime `dependencies`**: `vite`, `@vitejs/plugin-react`, `@tailwindcss/vite`, `tailwindcss`; plus `tsx` declared explicitly (it previously resolved only by hoisting the `packages/server` copy — an undeclared cross-workspace coupling).
- **Guards** so neither fix silently regresses: 5 new `verify-release-deps.mjs` rules, and repo-lints for cap single-source, the cap message literal, client dep placement, and the smoke Node matrix. New `qa/tests/15-omit-dev-build.sh` covers npm/pnpm × engine-strict.

### Corrected a false comment

`_smoke.yml:75-76` justified the engine-strict override as "the transitive appdmg engine range fail-fasts on Node 24/25". That is wrong: `appdmg@0.6.6` declares `engines: {node: '>=8.5'}` with `os: [darwin]` — no upper bound, and platform-filtered out of Linux installs entirely. A whole-lockfile scan finds exactly one upper-bounded engines range (`>=6 <7 || >=8`), which admits 26. The real historical cause per `.npmrc` was native compilation of `macos-alias`. Corrected unconditionally so the next engineer doesn't re-derive the same wrong conclusion.

## Verification

- **L1** — full `npm test`: **1206 files / 11970 tests passed, 0 failed**.
- **L2 (X1)** — `qa/tests/15-omit-dev-build.sh` on **Node 26**: all three arms PASS, including **arm 2**, the full-fidelity reproduction of the real #357 codepath (`npm install --omit=dev` + `.npmrc engine-strict=true` on a fresh `git clone`).
- **Negative control** — the same script against pre-fix `develop` on Node 26 reproduces `npm error code EBADENGINE / Required: {"node":">=22.19.0 <26"}`.
- **New lints mutation-verified non-vacuous** — reverting the cap to `major >= 26` reds `node-cap-single-source`; moving `vite` back to `devDependencies` reds `client-build-deps-runtime`.
- **Lockfile** — pure importer bucket move plus the one new `tsx` entry (already resolved at `4.23.1`); no package version changed, so the built client is unchanged. `pnpm install --frozen-lockfile` clean.
- **Browser E2E** — the docker harness was built from this branch and run. The 3 persistent failures (`editor-pane` ×2, `change-summary-table`) reproduce **identically on a freshly-built `origin/develop` baseline image**, so they are pre-existing and unrelated; this diff touches zero client source.

## Deferred to post-merge

- **M1** (task 5.1) — real-world `pi install git:...` on a fresh Node 26 machine. Its two failure signatures are already covered by X1 arm 2 above; only the `pi` wrapper itself is unverified.
- Task 6.4 — comment on and close #357 once the Node 26 CI leg is green.


robertcsakany:
@coderabbitai full review

robertcsakany:
Thanks — triaged all 13. Summary of what I changed and what I pushed back on.

**Accepted (4):**

1. **`openspec/specs/server-startup-node-version-guard/spec.md` stale `>=26`** — good catch, and the most valuable one. That line lives in the spec's *Purpose* preamble, which sits **outside** the MODIFIED requirement blocks, so `openspec archive --sync` never rewrote it. Fixed to `>=27`.
2. **`qa/tests/15-omit-dev-build.sh` defaulted to remote `develop`** — correct and important. As written it validated a different tree than the one under test, and would always pass on an unlanded branch. It now defaults to the enclosing checkout at its current `HEAD`; clone switched to `--no-checkout` + `git checkout <ref>` so a bare sha works. The `OMIT_DEV_REPO_URL`/`REF` overrides are retained.
3. **arm 2 guard `>= 26` → `== 26`** — agreed. Below 26 the assertion is vacuous; above 26 the engines cap refuses *by design*, which would read as a #357 regression.
4. **`packages/shared/src/AGENTS.md` rows too long** — right in effect: my rows pushed the file past the repo's 30000-byte `AGENTS_BYTE_CAP` (`kb dox lint` `over-threshold`). Trimmed all four; that still left it 15 bytes over, so I promoted the heaviest *pre-existing* row (`browser-protocol.ts`, 2243 chars) to a `browser-protocol.ts.AGENTS.md` sidecar per the Documentation Update Protocol. 30015 → 27917 bytes, flag cleared.

**Rejected, with reasons:**

- **"Move the change out of `openspec/changes/archive/`" (5 of the 13).** `openspec/changes/archive/YYYY-MM-DD-<name>/` is precisely what `openspec archive` emits and what the ship gate requires. The `AGENTS.md` rule being cited governs where **new** change artifacts are authored (never under `active/`/`archive/`), not where completed ones land. Applying this would undo the archive step.
- **"Un-check tasks 5.1 / 6.4."** Both are `(test-plan: manual-only)` rows, deferred to post-merge under the project's documented ship-change defer convention. 5.1 (M1) is additionally covered in substance by X1 arm 2 below.
- **Edits to the archived `tasks.md` / `test-plan.md`** (§2 ordering note, E5/E8 wording). These are accurate observations, but the files are now archived planning artifacts — historical records of what was planned. The live specs and code carry the corrected statements.
- **"Shorten `scripts/AGENTS.md` rows."** That file is nowhere near a cap and the rows match their neighbours' density.

**Verification on this branch:** `ci` green · all 9 `_smoke.yml` legs green including **both Node 26 legs, which run without the `--config.engine-strict=false` override** (that's X2/X3, the `EBADENGINE` regression gate) · `qa/tests/15-omit-dev-build.sh` on Node 26 passes all three arms, including arm 2, the full-fidelity `npm install --omit=dev` + `engine-strict=true` reproduction of #357 · the same script on pre-fix `develop` reproduces `EBADENGINE`.

PR #427, opened 2026-08-06
feat(apple-tools): provision iMCP (Apple PIM) via a dashboard plugin + CLI

robertcsakany:
Implements OpenSpec change `add-apple-tools-imcp-plugin`.

Provisions **iMCP** (Apple Calendar, Contacts, Reminders, Messages, Location, Maps, Weather) for pi via `pi-mcp-adapter`, as a new package `@blackbelt-technology/pi-dashboard-apple-tools`. macOS-only. **Apple Mail is explicitly out of scope** — iMCP exposes no Mail service; the skill redirects to `apple-mail-fast-export`.

## What lands

**Reusable core — a fourth requirement category**
- `PluginRequirements.paths` for absolute paths (`.app`-bundled binaries the PATH `binaries` probe cannot see), with single-`${configKey}` interpolation from the plugin's validated config.
- `probePath()` — existence check only; never executes the path, never builds a shell string, treats the manifest value as untrusted.
- Loader resolves per-plugin config (`configSchema` defaults + stored config) so `${configKey}` actually resolves in production.
- Unsatisfied `paths` render as a non-actionable warning pill (no `[Install]` — a path has no package source).
- Backward compatible: an absent `paths` behaves byte-identically, `missingRequirements` ordering unchanged.

**The package**
- Nine-state provisioning machine with `--check` as the write-suppressed twin backing the CLI, the doctor probe, and the settings panel — one implementation, three surfaces.
- Merge-only atomic config writers; refuses a present-but-unparseable file rather than "fixing" it; `brew` invoked with an argv array, never a shell.
- Dashboard settings panel, agent skill, doctor module, `BUNDLED_PLUGINS` entry, `RECOMMENDED_EXTENSIONS` entry.

## Verification

- **11,836 unit tests** pass (75 in `packages/apple-tools`). tsc + Biome + client build clean.
- **7/7 Playwright e2e** against the docker harness.
- `openspec validate` clean; 30 requirements synced across 4 capabilities.

Three suite failures are unrelated to this diff: two environmental (docker-required port derivation; git `denyDeleteCurrent`) and `rpc-keeper 3.5`, which arrives from develop, touches no file here, and is green on develop's own CI.

## Notable findings from the gates

The harness and a real-macOS pass caught four defects unit tests could not:
1. Server entry exported only `registerPlugin`; the loader calls `mod.default(ctx)` → plugin silently never loaded.
2. `${configKey}` interpolation was implemented but never wired into the loader.
3. `set-disabled` took a browser-supplied `cwd` straight into `mkdir -p` + write → now validated against `host.knownFolderCwds`.
4. `--check` reported "iMCP provisioned" and named a nonexistent binary on a host without iMCP; the doctor reported `requiresRemediation: false` there. Check-mode *predictions* were leaking into *reports*.

Security hardening from the audit: atomic writer (`wx` + random temp name, `0600`, `fsync`, cleanup), malformed-but-parseable configs refused, 10s status TTL, GET no longer mutates, and `run-installer` never shells out to `brew` in-process (it would have blocked the event loop for up to 10 minutes).

## Deferred to post-merge

`11.5`–`11.8` are manual-only: real macOS install → menu-bar TCC grants → live Calendar round-trip, and the 4-theme visual pass. TCC grants cannot be automated (Apple's security model has no API). `11.8` (non-macOS no-op) is largely covered by e2e `#F8` and doctor `#X19`.


robertcsakany:
Thanks — most of these were valid and are fixed in 05398399. Two I'm resolving differently, with reasoning on the record:

**1. "Move this OpenSpec change out of `archive`" — not applying.**
The guideline ("store artifacts under `openspec/changes/<name>/`, never under `active/` or `archive/`") governs where *in-flight* changes live. This directory is the terminal state of the ship pipeline: `openspec archive` moved a completed change to `openspec/changes/archive/2026-08-06-<name>/` and synced its 30 deltas into `openspec/specs/`. Moving it back would un-archive a finished change and re-open it as active work.

**2. "Keep the installer action available when iMCP is absent" — spec updated instead of code.**
You're right that implementation and spec disagreed; I fixed the spec, not the behaviour, because the behaviour is a deliberate availability fix.

The provisioning traversal is synchronous by design (pure and unit-testable — 79 tests depend on it). Its install branch shells out to `brew install --cask` under a **ten-minute** timeout. Running that inside the dashboard server blocks the Fastify event loop for the duration: every session's WebSocket and every other plugin's HTTP stall. A single click could freeze the whole dashboard for ten minutes.

So provisioning is split: the server performs only the fast config-write half (which is all that's needed once iMCP is on disk), and the CLI entry point owns the long, network-bound install. When the app is absent the panel says so and names `pi-apple-tools-install` rather than offering an action that would refuse — so there is always an in-UI path forward. The server also refuses the action if invoked by other means (defence in depth).

That contract is now written down in `openspec/specs/apple-tools-provisioning/spec.md` under "The dashboard never runs the network install in-process", and the run-installer scenario is scoped to the app-present case.

robertcsakany:
Second round applied in 484e9033 (spec: both disable scopes; SKILL.md: `READY` removed from the CLI state list; e2e `#F9`: all seven services asserted).

Two more I'm declining, with evidence:

**`package.json` — "Publish a JavaScript installer target for `bin`."**
A `.ts` bin target is the established convention for pi packages in this repo, not an oversight. Four siblings ship exactly this shape:

| package | bin |
| --- | --- |
| `dashboard-plugin-skill` | `src/bin/scaffold.ts` |
| `nano-banana` | `src/bin/nano-banana.ts` |
| `video-production` | `src/bin/veo.ts` |
| `video-transcription` | `src/bin/transcribe.ts` |

These are consumed through pi, which loads packages under a TypeScript-capable loader. Switching this one package to a compiled JS target would make it the odd one out and add a build step no sibling has. If the repo wants compiled bins, that's a cross-cutting change for all five, not this PR.

**`requirement-probes.ts` — "Cache path results by path key."**
Declining as speculative. `probePath` is reached through `runRequirementProbes`, and the loader's refresh cycle is already behind the 30s report cache keyed by plugin id — the path that actually runs on a schedule. A second cache layer keyed by path would add invalidation surface (the whole point of a `paths` probe is that the file can appear or vanish) to save a `existsSync` call that no measurement has shown to matter. The repo's guidance is simplicity-first and measure-before-optimize; happy to revisit with a profile showing it on a hot path.

robertcsakany:
@coderabbitai full review

robertcsakany:
Final status on the remaining open threads:

- **`env.ts` — "Handle partial writes before `fsyncSync`"** — already fixed in 05398399; the thread predates it. `writeFileAtomic` now drains the buffer before the fsync:
  ```ts
  const buf = Buffer.from(content, "utf8");
  let off = 0;
  while (off < buf.length) off += writeSync(fd, buf, off, buf.length - off);
  fsyncSync(fd);
  ```
- **`archive/tasks.md`, `package.json` bin, `client/index.tsx` installer action, `requirement-probes` path cache** — declined with reasoning in the two comments above (terminal archive state; `.ts` bin is the convention across four sibling packages; the installer split exists to stop a ten-minute event-loop block and the spec now documents it; the path cache is speculative without a profile).

Merging with `job-object-windows` red. It is a known-flaky infra job: it self-labels its bail `INFRA: app never brought a server up on :8000`, and it flip-flops on unrelated branches (`os/fix-electron-auto-update-pipeline` and `os/add-macos-electron-launch-smoke` each have both a pass and a fail). The one hypothesis that would implicate this PR — a bundled plugin breaking Electron's server boot on Windows — is disproved by `electron-e2e (windows-latest)` passing, which boots the same packaged app. `ci` (lint + test + build) is green.

PR #428, opened 2026-08-06
feat(deps): declare every undeclared dependency, gate publish correctness

robertcsakany:
Implements OpenSpec change `cleanup-undeclared-dependencies`. Archives + syncs specs.

## What this does

Drives Biome's `noUndeclaredDependencies` to **zero at repo-root scope (1411 → 0)** and adds a publish-correctness checker that catches a defect class Biome structurally cannot see.

## Reviewer's attention, ranked

**1. The verification oracle changed (design defect found during implementation).**
Biome 2.5.1's `--only=<rule>` force-enables the rule and **bypasses `overrides` severity entirely**. The change's specced probe (`biome lint --only=... .`) could therefore never reach zero while tests and build-configs are resolved by override — which D1 deliberately chose. Only `files.includes` exclusions survive `--only`.

Resolution: the rule is now `error` in the **base** config and the oracle is a plain `biome lint .` filtered to the category. Stronger end state — a permanently enforced rule rather than an ad-hoc probe. The asymmetry is pinned by a test so it cannot be "simplified" back into a broken oracle.

**2. Published-surface changes — the parts worth the hardest look.**
- **35 pre-existing `"*"` ranges** de-wildcarded (root + 7 workspaces). Replacements are **lower bounds (`>=`), not carets**: a caret would exclude older hosts the `"*"` admitted and break already-published consumers. Concreteness was the requirement; tightening was not.
- **17 workspaces stopped shipping `__tests__/` to npm.** `files: ["src/"]` was sweeping `src/__tests__/` into tarballs; those shipped tests import devDeps (never installed for a consumer) and dangle relative paths absent from the tarball. Also shrinks tarballs — `bus-client` shipped 16 of 25 files as tests.

**3. Four genuine consumer-visible breaks**, found by running the checker against the real repo before calling the work done (the plan required this):
- `dashboard-plugin-runtime/src/server/config-validator.ts` imported `ajv` from devDependencies — **would have thrown for a consumer**. Moved to `dependencies`.
- `vite` → peer (vite-plugin export); `fastify` → peer in `kb-plugin` (matching 4 siblings); `bus-client/src/codegen/` excluded (dev tooling, not an export).

## The checker

`scripts/verify-published-imports.mjs` derives each workspace's shipped file set from `npm pack --dry-run` (so `files`/`.npmignore` match the registry exactly) and asserts every specifier resolves against that workspace's **own** manifest. Deliberately stricter than Biome: a `devDependency` does **not** satisfy a shipped import.

> Biome asks *"is this import declared anywhere?"* — this asks *"will this import resolve for someone who installed the tarball?"*

Faults are errors, never silent skips (pack failure, unparseable source). Bounded guarantee, stated in the file: it proves **declaration, not installability** — a declared `^99.0.0` passes.

One suppression exists: `@pi/anthropic-messages` returns **E404**, so it can be neither declared (unresolvable dep in a published manifest) nor overridden (it sits in shipped source). Carries its reason inline.

## Test placement

E15/E26/P1 are `ci`-level per the test plan's own level column, not L1. Left in the default suite they shell out to `npm pack`/`npm publish --dry-run` across 32 workspaces and their CPU spike **starved unrelated 5s-timeout tests** — the suite failed in `packages/extension`/`packages/server`, pointing away from the real cause. Gated behind `RUN_CI_SCENARIOS=1` and wired as its own CI step. Suite: 6 failures → 0, runtime 408s → 220s.

## Verification

| Gate | Result |
|---|---|
| `noUndeclaredDependencies` @ repo root | **0** (was 1411) |
| Publish checker | **0 errors / 32 workspaces**, ~26s (budget 60s) |
| `npm test` | **12,174 passed, 0 failed** |
| `tsc --noEmit` | clean |
| `biome lint .` | **0 errors** |
| `openspec validate --strict` | valid |

No QA/manual tasks deferred — the manifest has 32 automated rows and no manual-only rows. No docker harness: the test plan records 0 L3/L2 scenarios (this change renders nothing).

## Out of scope, recorded not fixed

8 pre-existing unsatisfiable ranges (e.g. `vitest: "^2.1.8"` while 4.1.10 resolves) are pinned as a **one-way ratchet** — they stay visible and may shrink, but a new violation fails the build.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai Thanks — four of the five were valid and are fixed in 0188058. One disputed.

**Applied:**

1. **`@earendil-works/pi-ai` compatibility (the best catch).** You were right, and it was a real bug. `packages/extension` had `^0.75.5` while every other de-wildcarded host peer got `>=`. On a 0.x version a caret pins the *minor*, so `^0.75.5` admits only `>=0.75.5 <0.76.0` — rejecting hosts at 0.76.0 and 0.80.10, precisely the consumers the previous `"*"` accepted. Now `>=0.75.5`. I also took the second half of your suggestion: added `selectHostPeerRange()` so the host-peer rule has a name instead of being applied by hand, plus a **set-based** test asserting no optional peer anywhere in the repo is a caret, so a future one cannot regress the same way.

2. **`mapfile` masking discovery failure.** Correct — and it is the same vacuous-pass class this PR exists to eliminate, so it would have been an embarrassing one to ship. Captured into a variable under `set -e` with an explicit empty-set rejection.

3. **`concurrency: 0` → zero runners.** Correct; it returned an empty finding list that reads as a pass over nothing. Now throws a `TypeError`, with a table-driven test covering `0, -1, 1.5, NaN, "8"`.

4. **Unbounded `npm pack`.** Correct — it executes each workspace's `prepack` lifecycle script, so a hang would stall the run indefinitely and blow the 60s CI budget with no diagnostic. Bounded at 120s, surfacing through the existing `pack-failed` path.

5. **`TBD` Purpose** in the archived spec — replaced with the real purpose, including the explicitly bounded guarantee (proves *declaration*, not *installability*).

---

**Disputed: "Move this OpenSpec change out of `archive/`."**

This one is a misreading of the rule. `AGENTS.md` says:

> Place change artifacts at `openspec/changes/<name>/` (never under `active/`/`archive/`); prefer `openspec change new <name>`.

That governs where a change is **authored** — the guidance is "don't hand-create your new change inside `active/` or `archive/`". It is not a prohibition on the archive directory itself, which is the required **terminal** state of a completed change. `ship-change` step 3 mandates it:

> Archive to `openspec/changes/archive/YYYY-MM-DD-<change>/`

and step 8.5 makes it a hard merge gate — the PR may not merge *unless* the change is archived there. The move was performed by the official tooling (`openspec archive`), which also synced the delta specs into `openspec/specs/` and named the directory `2026-08-06-cleanup-undeclared-dependencies` per that convention. Note your suggested target — `openspec/changes/2026-08-06-cleanup-undeclared-dependencies/` — keeps the archive *date prefix* while dropping the archive *directory*, which is neither the authoring convention nor the archive convention.

Reverting it would break the ship gate and de-sync the specs, so leaving as is.

PR #429, opened 2026-08-06
feat(notify): split notify out of the prompt_request protocol

robertcsakany:
Implements OpenSpec change `split-notify-from-prompt-request`.

## Why

A freshly spawned session showed **"Needs you"** within ~30 s and never stopped. `ctx.ui.notify` — a fire-and-forget toast — shipped over `prompt_request`, so every consumer treated it as an unanswered ask: `trackPromptRequest` armed `currentTool = "ask_user"` and the M1 fold re-armed it at every quiescent moment. Blast radius: needs-you card + stripes, inflated folder rollup, `questionFirst` reorder, false unread dot, a client-side `interactiveRequests` leak, and a session the embed-lifecycle reaper could **never** reclaim.

## What changes

- **shared** — `NotifyMessage` / `BrowserNotifyMessage` (`{type:"notify", sessionId, notifyId, message, level?}`): no `promptId`, no `component`, no `placement`. `NotifyLevel` includes `success` (already rendered today). `normalizeNotifyLevel` shared by both send sites.
- **extension** — `createNotifyProxy` replaces the `prompt_request` proxy; no `as any` at the send site; never enters PromptBus.
- **server** — a `notify` branch that only logs + fans out (no tracking, no `currentTool` write, no unread stamp, no reorder, no `session_updated`), plus a **permanent** skew guard that normalizes the legacy shape from already-published bridges *before* `trackPromptRequest`.
- **durability** — bounded per-session notify log (cap 50, oldest-first), persisted via `sessionToMeta`/`sessionFromMeta`, carried across a bridge reattach, replayed by `replayNotifyLog` at all four subscribe sites. It never feeds `hasPendingAsk`; an ended session keeps its rows (reapability by exclusion, not deletion). Cap evictions counted on `/api/health#notifyLog`.
- **client** — `addNotify` appends **only** an `interactiveUi` row, deduped by `notifyId` (not text), in **both** reducers.

## Verification

- 12 145 unit/integration tests green; `tsc --noEmit` clean; client build green.
- The load-bearing assertions (`#X3`/`#X4` no re-arm after a turn, new **and** legacy shapes) were confirmed **red** before the fix.
- Pinned negatives kept: a genuine `prompt_request` is still tracked / folded / unread-stamped, and still re-arms after `tool_execution_end`.
- L3 docker harness: 5 notify scenarios green (`tests/e2e/notify-channel.spec.ts`), driven through the real `ctx.ui.notify` path via a new `e2e_notify` fixture tool.
- Live instance: server restarted on this build, no session flipped to `ask_user` over an 80 s at-rest window, and a previously stuck session cleared on restart (design Decision 5).

## Drive-by fixes (unrelated, called out)

- `docker/test-entrypoint.sh` pins the dashboard bridge **first** in `settings.json#packages`. pi resolves duplicate tool names **first-registration-wins**, and pi-flows' `ask_user` (`{question,type}`) was shadowing the dashboard's (`{method,title}`) — every faux ask scenario failed schema validation, including the untouched `faux-ask.spec.ts`. That spec is green again.
- `recommended-routes.test.ts` count assertion 19 → 20 (`video-transcription` was added on develop without bumping it; fails on develop itself).

## Deferred

Manual-only tasks 8.1 (per-level render comparison) and 8.2 (old-server + new-bridge skew, accepted per design Decision 9) → post-merge verification.


robertcsakany:
@coderabbitai review

PR #430, opened 2026-08-07
feat(settings): field description contract + page reorganisation

robertcsakany:
Implements OpenSpec change `reorganize-settings-pages-and-descriptions`.

## Why

Only **8 of 52** shared-component settings fields carried a description, and none of the four shared field components associated their `<label>` with their control — so those controls had **neither an accessible name nor an accessible description**. Fields were also ordered by accretion, and `displayPrefs.debugTools` had **two live controls with divergent persistence**.

## What changed

- **Field contract.** `NumberField`/`ToggleField`/`SelectField`/`TextField` gain `useId()` label association, a **required** `hint: React.ReactNode` wired via `aria-describedby`, and an optional `unit` chip rendered inside the label. `hint` is required so omitting it is a **type error** — `tsc` enumerated exactly 52 sites. No allowlist file, no source-scanning test (both explicitly rejected in design D1).
- **44 descriptions authored.** 6 sites keep `hint={null}` deliberately: OAuth `Client ID`/`Client Secret`/`Issuer URL` and provider `Base URL`/`API Key` are terms of art.
- **Units moved into chips** across **10 i18n keys** in **en / zh-CN / hu**. The translated strings embedded the unit, so English-only stripping would have rendered it twice in those locales.
- **Duplicate `debugTools` control deleted** (`DebugToolsToggle` + the Developer "Chat Display" section). ⚠️ **User-visible:** that preference is now Save-Bar-buffered instead of instant-apply.
- **Sessions regrouped**; `defaultModel` moves from position 12/13 to **first**, in a `--severity-info-*` callout stating it applies only to brand-new sessions.
- `dashboardName` → General ▸ Interface (with its `CONFIG_FIELD_PAGE` entry, so the Save Bar chip follows).
- **`GatedGroup`** indents gated controls (presentational only — no gating logic changed).

## Explicitly NOT done

- **`tunnel.watchdog.*` stays on Server** (design D2). `computeConfigPartial` emits `partial.tunnel` as one top-level key and `dirtyPages` attributes per top-level key, so the chip would read Server wherever the JSX lived. A test guards the reversal.
- **`spawnRegisterTimeoutMs` stays bespoke** (D3). Its bounds check blocks the write *and* disables Save; the shared `NumberField` is `parseInt(…) || 0` with no bounds, so converting it would delete an enforced constraint. Pinned by a boundary-value suite.

## Tests

40 scenarios in `test-plan.md`; all **37 automated** rows implemented and green (**83** settings assertions), including **2 L3 Playwright rows** verified against the docker harness in a real browser accessibility tree.

## Two unrelated develop-side fixes included

Both blocked this change's verification gates:

1. **`docker/test-entrypoint.sh` unbound `PI_DIR`** — evaluated outside the `if` that gates it, so under `set -euo pipefail` an unseeded `test-up.sh` died immediately after its own smoke printed "dashboard ready". This was breaking the **entire E2E harness**, not just this change.
2. **`recommended-routes.test.ts`** hardcoded `toHaveLength(19)`, stale since video-transcription was added. Now derived from `RECOMMENDED_EXTENSIONS.length` so it cannot go stale again.

Manual-only test-plan rows (visual sweep, copy review, change-summary disclosure) are deferred to post-merge verification.


robertcsakany:
Thanks — 5 of 6 applied in `deeb8c8d0`. Disputing the sixth with evidence.

**Manual checkboxes in the archived `tasks.md` (12.1–12.3): intentionally checked, not a false claim.**

This repo's `ship-change` skill defines a manifest-aware defer rule (step 1):

> a leftover `- [ ]` is deferrable **only if** it maps to a `manual-only` manifest row — either an inline `(test-plan: manual-only)` tag or a `(test-plan #<id>)` reference resolved against the manifest … Flip each deferrable `- [ ]` → `- [x]` (validated post-merge).

All three rows carry the literal `(test-plan: manual-only)` tag and map to `test-plan.md` rows F11, F21, F22, each of which records `disposition: manual-only` with level `—` because its observable is a human judgment (visual sweep at 3 breakpoints, copy quality, change-summary disclosure). Leaving them unchecked would make `openspec status` report the change incomplete and block the archive gate, which is what the convention exists to avoid.

The deferral is stated in the PR body and in the archive commit message, so the post-merge obligation is recorded rather than hidden.

The remaining 4 comments were outside the diff and are addressed in the same commit.

robertcsakany:
@coderabbitai full review

PR #432, opened 2026-08-08
fix(promises): handle 93 floating/misused promise sites in client, plugins, shell

robertcsakany:
Rung 1b of the type-aware lint ladder (`add-typeaware-lint-gate`). Clears every promise site claimed by `cleanup-client-plugin-promises`.

## What this clears

| Rule | Before | After (this change's scope) |
|---|---|---|
| `noFloatingPromises` | 143 repo-wide | **0** in client / flows / roles / shell / subagents / automation / `scripts/` |
| `noMisusedPromises` | 11 repo-wide | **0** in client + server |

Remaining repo-wide: 54 floating + 6 misused, all owned by `cleanup-async-semantics-server-extension`.

## Instrumentation lands first (design D2/D7)

- **`packages/client/src/lib/report-error.ts`** — `reportError` / `logRejection` / `installUnhandledRejectionReporter`. The client had no central logging path, so "route handlers to the existing logging path" was undefined for this package. This defines it and gives every handler one observable seam a test can assert on.
- **Global unhandled-rejection reporter** in the client entrypoint and in electron main (through the existing `log()`), installed before any startup work. This is the change's regression guard: an escaped rejection stops being silent.

## The misused sites were not what the proposal assumed

4 of 5 were promise-in-conditional truthiness guards used for inflight memoization — correct by accident. Fixed by **per-site** narrowing, no behaviour change: `!== undefined` for `WorktreeActionsMenu` (declared `| undefined`; `!== null` would be a TS no-overlap error), `!== null` for the other three.

The 5th was a real defect: `tunnel-core.ts:167` used `new Promise(async (resolve) => …)`, which swallowed any throw in the executor — the outer promise never settled and every caller hung forever. Test E1 reproduces the hang before the fix.

## A predicted coupling landed

Making `createTunnel` able to reject turned the sibling's floating `promise.finally()` at `tunnel-core.ts:160` into a **live** unhandled rejection, so it is fixed here rather than left dangling. Ledger settles at **89 + 54** (was 88 + 55); the sibling's proposal and the gate's ownership table are updated to match. Both proposals had flagged this exact same-function coupling.

## Fix distribution (D5 deliverable)

86 of 86 product-code floating sites took `.catch(handler)`; 0 `await`; 0 `return`; **0 bare `void`** (banned by D1, verified mechanically over the diff). Every site lives in a React lifecycle or event path where "return the promise" is structurally unavailable. Full record + the severity evidence for `add-typeaware-lint-gate`: `openspec/changes/archive/2026-08-08-cleanup-client-plugin-promises/classification.md`.

Repeated discards were extracted rather than duplicated: `makeSafeSend` (flows-plugin, 7 sites), the roles-plugin `dispatch` seam (5 sites), `logRejection` (client).

## Verification

- Full unit suite: **12307 passed, 0 failed**. `tsc --noEmit` clean. Zero new Biome warnings on changed files.
- New scenarios green against the docker harness: F1 (packages), F2 (provider-auth), F3 (EmlPreview), F4 (network-discovery), F6 (handler installed before app work), X1 (aborted request reported, not swallowed); E1–E6, F5, X3, X4 at L1.
- **X2 downgraded to manual-only**: `ci-electron.yml` is a dispatch-only installer build matrix and never launches the app, so it cannot assert main-process log output. A static wiring guard covers it meanwhile. Recorded in `test-plan.md` per the task's own fallback.

### Known limitation, stated plainly

The repo's browser-E2E baseline is red in ~10 specs **independently of this change** — each was reproduced identically against a harness built from the base commit (bus-client-goal-plugin, change-summary-table, editor-pane ×3, file-preview-survives-churn, openspec-artifact-dialog ×2, project-init-button, roles-custom). Browser E2E is not in CI, so develop drifted. The full 80-spec suite could not be run to completion locally: the harness container goes unhealthy under sustained load. A follow-up issue tracks this.

### Note for anyone re-deriving these counts

Biome's type-aware rules **silently undercount in a worktree with no `node_modules`** — inference degrades and whole packages report zero (measured 95 floating vs the true 143). Run `pnpm install` first, and count from Biome's own diagnostic total, including `.cjs`.


robertcsakany:
## Review loop — round 2

Thanks; one of these caught a real hole in my own testing.

### Fixed

- **F6 was near-vacuous** (`promise-rejection-surfaces.spec.ts`). It asserted the *spec's own* init-script listener saw the injected rejection, never the application's reporter — it would have passed with `installUnhandledRejectionReporter()` deleted. It now fires inside the loaded app and asserts the console record carries `[pi-dashboard] unhandled error`. Verified non-vacuous with a negative control (same rejection on a page without the bundle produces no such record). The "installed before app work" half moved to a source-order guard, since no browser test can prove it: a rejection fired before the bundle runs can never reach a listener the bundle installs.
- **F1 race** (latest round): `checkNow` is already enabled, so asserting `toBeEnabled` after the click could pass before the work started. Both the check and the search now arm a `waitForResponse` *before* acting.
- **Polling paths**: `usePiCompatibility`, `usePiCoreVersions`, `usePiResources` interval callbacks now handle rejections. Biome never flagged these (an arrow that *returns* the promise isn't "floating"), but the rejection was equally unobserved — good catch.
- **`dispatchFlowAction`** is the single funnel for every flow action, so the discard is stated there rather than at 3 of 5 call sites. No raw `send(` remains in that file.
- **E1** now requires the injected error to *reject* (`resolve(null)` is indistinguishable from the legitimate no-binary path) and clears its deadline timer.
- **Electron wiring guard** now proves the *reason* reaches `log()`, mutation-checked: replacing the body with a fixed string fails it.
- **eml F3** no longer hides behind `if (isVisible)`; it requires the attachment path and waits for a terminal UI state.
- **QR payload**: a malformed payload now surfaces as a pairing error instead of being mislabelled by the camera catch.
- **Sibling ledger** (`cleanup-async-semantics-server-extension`): server production count corrected to 12. I deliberately did **not** renumber the "10 of the 13" coverage split — only that change's owner can say which bucket the absorbed `tunnel-core.ts:160` fell into, and guessing would be worse than flagging it. Marked for re-derivation.
- Typo, MD028, and task 6.3's overclaim (it no longer reads as "the whole E2E suite is green" — it states what ran, what did not, and why).

### Disputed / already present

- **`## Discipline Skills` section**: it is present in the proposal (one occurrence). Reads like a false positive.
- **`AutomationRunMonitor` structured reporting**: `automation-plugin` has no central reporter, and `getRunResult` already maps ordinary fetch failures to `null`, so this handler fires only on an unexpected throw. Introducing a reporting seam for one plugin is outside this change's scope; `console.error` matches every other plugin here.

### CI

The one remaining `ci` failure is `openspec-poller-parity` on **`add-folder-actions-menu`** — a change from develop that has no `tasks.md`. The same test was already failing on develop's own CI run at this branch's base commit (on `reorganize-folder-card-header`, before my merge). Not introduced here. `job-object-windows` self-reports `INFRA: app never brought a server up on :8000` and last passed on develop in July. Both tracked in #433.

PR #434, opened 2026-08-08
refactor: break all 17 module import cycles (noImportCycles → 0)

robertcsakany:
Takes `biome lint --only=lint/suspicious/noImportCycles .` from **17 → 0**, discharging the ratchet precondition that `add-typeaware-lint-gate` blocks on.

The 17 diagnostics are **edges, not cycles** — the rule emits one per participating import. They decompose into 4 disjoint SCCs, cut once each. Every predicted drop landed exactly:

| Cut | Package | Drop | Technique |
|---|---|---|---|
| D1 | server | −2 | extract `isLoopback` → `auth/loopback.ts` |
| D2 | flows-plugin | −2 | extract `formatCost` → `client/format-cost.ts` |
| D4a | client | −2 | extract `isExternalHref` → `preview/is-external-href.ts` |
| D3 | client | −5 | split `viewerRegistry` by open path + `EditorPane` dispatch |
| D4b | client | −6 | invert `MarkdownContent → FileLink` via injected `ToolContext.fileLink` |

17 → 15 → 13 → 11 → 6 → **0**, re-probed after every cut.

## Notes for review

**`lazy()` is not a cycle fix.** Biome traverses dynamic `import()` edges, so `React.lazy` leaves the diagnostic standing. Two independent designs picked it before the mechanism was checked; it is now recorded as a spec requirement.

**D3 — the split needed a dispatch, not just a split.** `CappedViewer` can't import both halves (that re-forms the cycle) and can't import only half (a) while still receiving pseudo-tab kinds. So `EditorPane` gained a kind discrimination, and `CappedViewer`'s prop narrows to `OpenPathViewer`. The narrowing uses an explicit `isPseudoTabViewer` guard rather than a cast — `fileKind()` returns the full `ViewerKind` from `packages/shared`, which is out of scope to narrow, and TS won't narrow through a bare `.includes()`.

The partition is proven at compile time by four `_AssertNever<T extends never>` checks. Worth knowing: the obvious spelling `const _x: T[] = []` is **vacuous** (an empty array literal is assignable to every array type, including `never[]`) — verified against this repo's `tsc`. There is a test that mutates the guard four ways and asserts it fails closed.

Those checks prove the union is *partitioned*, **not** that each kind is in the correct half — a swap still partitions. That is covered by a runtime decision-table test.

**D4b — the risk is a silent regression.** `fileLink` must stay optional (`ToolContext` is published via `chat-embed`), so a builder that forgets it drops linkification with no type error and no cycle-probe signal. Mitigated by routing both production sites through `makeToolContext`, merging a default in `ChatView` for embedder-supplied contexts, and moving the gate at **all three** trigger sites (`p`, `li`, and the inline-`code` override). The regression test asserts through the real builders, not fixtures — a fixture only proves `MarkdownContent` honours the field, never that production sets it. It is mutation-verified: dropping `fileLink` turns 9 assertions red.

`ChatView` merges once in a `useMemo` and passes the merged context to **all six** consumers — including `ToolBurstGroup` and `CollapsedToolGroup`, which render `MarkdownContent` internally and are easy to miss.

## Verification

- `noImportCycles`: **0** (exit 0)
- Full suite: **12318 passed**. Only red is `doctor-route`'s 3s timing assertion, which passes 3/3 idle and references nothing in this diff.
- `tsc --noEmit`: no new errors (the one `cli.ts` error reproduces on `origin/develop`)
- `npm run build`: exit 0
- E2E vs a harness **rebuilt from local code**: 16 passed. The 5 red specs are **pre-existing** — an identical harness built from `origin/develop` fails the same 5.

## Known limitations

- **The E2E oracle is weaker than planned.** `file-preview-survives-churn` (the intended D4b oracle) is already red on `develop`, so the unit-level mutation-verified test is the real evidence here.
- **C1 resolved: a claimed behaviour change was a no-op.** `/api/file` resolves `diff:<rel>` to a non-existent path, so the probe always missed → `size` 0 → the size gate never fired for a pseudo-tab. Routing them around `CappedViewer` only stops a spurious always-404 request. Scenario E8 retired.
- Task 7.5 (manual visual parity) is `manual-only` and deferred post-merge.


robertcsakany:
Thanks — applied 9 of the 11 findings in `dbaaec2`. Two I'm disputing with evidence.

**Applied**
- `make-tool-context.ts` → `.tsx` in design.md + tasks.md.
- Test-path records corrected to where the tests actually landed: `flow-agent-cost.test.tsx` (D2), `preview/__tests__/is-external-href.test.ts` (E9), `__tests__/MarkdownContent.test.tsx` (F6).
- Scenario totals restated after E8 was retired: **automated 20**, edge 9, L3 3, `retired 1`.
- E5 now specifies comparing registry keys as **sets**, not by insertion order.
- Added an explicit E2E completion record (the Major finding): 16 passed / 5 failed, each failure named and shown red on an `origin/develop` baseline harness, plus the substitute evidence actually relied on.
- AGENTS.md de-staled: `viewer-registry` no longer claims the pseudo-tab kinds, `FlowAgentCard`/`FlowAgentDetail` no longer claim `formatCost` lives in the card, superseded duplicate `ChangesRailSection` row removed.

---

**Disputed 1 — `ChatView.tsx:321`, "pass the normalized context to every markdown consumer" (`MessageBubble` at 945-952).**

Correct observation, but **pre-existing and out of scope**. On `origin/develop` the same asymmetry already exists — the first `MessageBubble` omits `context`, the second passes it:

```
$ git show origin/develop:packages/client/src/components/chat/ChatView.tsx | grep -n -A7 "<MessageBubble"
936:  <MessageBubble
941-    onFork={onForkFromMessage}          # no context
1100: <MessageBubble
1106-   context={toolContext}
```

My diff does not touch either call site (only the prop rename plus the `useMemo` merge). Adding `context` there would newly **enable** linkification on a surface that never had it — a behaviour change, in a change whose stated goal is zero behaviour change. Better as its own change; happy to file it.

**Disputed 2 — "move the change artifact out of `archive/`".**

Working as intended. `openspec archive` deliberately moves `openspec/changes/<name>/` to `openspec/changes/archive/<date>-<name>/` as the final ship step, and it synced the `code-quality-loop` spec delta in the same run. The "not under archive/" rule applies to *in-flight* changes; this one is complete.

PR #435, opened 2026-08-08
Update pi core to 0.84.1 and adopt its APIs

robertcsakany:
## Why

pi `0.84.1` is published; the repo pinned `^0.83.0`. The 0.84.x line carries eight upstream breaking changes plus five new features (all breaks landed in 0.84.0; 0.84.1 is a patch with none).

This bumps the governed pins, resolves the breaks that actually reach dashboard code, and — equally important — leaves **checkable evidence** for every break judged not-applicable, so a future reader can tell "audited" from "not checked".

## Verification overturned three planning premises

Each was disproved by reading the installed tarball / repo rather than trusting the plan. All were amended in-flight with the reviewer in the loop:

| Premise | Reality | Landed as |
|---|---|---|
| `minimum` is anchored to bundled-extension peer-deps | `packages/electron/resources/bundled-extensions/` **never existed** (`git log --all` empty); `.gitignore`d build artifact; `forge.config.ts:109` records its removal | D2a — clause dropped, tasks 2.4/2.8 + test-plan E2 retired. The delta also **repairs the stale main spec** on archive |
| 0.84 replaced the harness session model → needs feature detection | The break is in **pi-agent-core**, not the `pi-coding-agent` SDK we call. `createAgentSession`, `SessionManager`, `.inMemory` all still exported; `sdk.d.ts:101` still documents our exact call | D3a — no unreachable branch built |
| `bridge.ts:426` is a fragile call into pi internals (listed as the change's largest risk) | It is a **comment**. The code uses public `pi.sendUserMessage`/`pi.isStreaming`. Behaviour re-verified on 0.84.1; only line numbers moved | D3b — risk retired, citation fixed |

Section 8's adoption targets were likewise misaddressed: pi owns `AGENTS.override.md` resolution end-to-end, and `pi-resource-scanner.ts` never classified context files. The real sites were the kb tooling (D7a) and three whitelisting layers for `samplingParams` (D7b).

## Breaks fixed (each red before the fix)

- **Null header-deletion markers** — `getApiKeyAndHeaders()` now returns `string | null`. The namer's `Object.keys(headers).length > 0` gate stayed true for a null-only map, so it issued a request with **no credentials at all**. Markers are forwarded unchanged (they exist to stop a placeholder OpenAI key reaching Cloudflare AI Gateway).
- **`ModelRegistry.refresh()` options/result** — the severe half was silent: refresh became **async**, so `refresh(); getAvailable()` read the catalogue *before* it resolved. Now awaited and inspected via a shared `reportRefresh()`; a credentials reload refreshes only the providers its diff touched.
- **OAuth abort signal** — both call sites passed credentials alone, so a provider that never answered hung every request through that credential. Now bounded by a real `AbortSignal`; a late answer can't persist a discarded credential.

## The harness earned its keep

`/api/health` reported `current: 0.83.0` with a spurious `upgradeRecommended` **while running 0.84.1** — several workspaces declare a broad `>=0.80.10` range while the server pins exactly, so pnpm's hoisted linker resolved two copies and the probe read the hoisted one. Pre-bump both were 0.83.0, which masked it. A green typecheck and 12k green unit tests both said this was fine. Fixed by pinning *resolution* via a workspace override — no declared floor moves (D8).

## Recorded as audited-not-applicable (with executable evidence)

`message_update` delta-only (stdout surface, not in-process — E7/E8), `tool_call` `terminate` (bridge never blocks — E17), `ModelsStreamTransforms` rename, `setRuntimeApiKey()`, provider `context.stored`/`publish()`, fullscreen TUI + terminal Mermaid/LaTeX. Baseten and Qwen Token Plan Individual need no wiring — the provider catalogue is built generically.

## Verification

- Full suite **12326 passed** (baseline 12281; +45 assertions). One unrelated, untouched perf budget (`display-fit-perf`) flakes under parallel load — passes 3/3 in isolation, failed 3 different ways across 4 runs.
- Docker harness rebuilt on the moved pin: `pi-084-runtime.spec.ts` 4 passed; existing `chat-render-fx` + `chat-transcript-virtualization` 13 passed; `enhance-tool-call-grouping` 3 passed. F2/F3 ride those existing suites (D9) rather than the fixture-flaky duplicates I first wrote.
- `verify-release-deps.mjs` exit 0 · `openspec validate --strict` clean · 59/59 tasks.

## Note for review

`openspec/specs/pi-core-version-check/spec.md:412/416` still use `0.83.0` as *illustrative* examples in a requirement this delta didn't modify — dated but not incorrect. Left alone to keep the diff surgical.


robertcsakany:
Thanks — round 1 triaged. Fixed the still-valid findings and pushed; disputing the rest with evidence.

## Fixed

- **`internal-auth-storage.ts` — deadline not enforced independently.** Correct, and it was a real hole in this change's own fix: `abort()` only *notifies*, so a provider that ignores its signal would hang the call forever **and** hold that provider's `refreshLocks` entry. Now races the provider against an abort-rejection (late rejection swallowed so it can't surface unhandled). New test drives a never-settling provider and asserts the lock is released — it would hang against the previous code.
- **`pi-version-skew.test.ts` — prerelease.** Correct: `parseVersion` discards the suffix, so `0.84.1-rc.0` passed while below stable `0.84.1`. Prereleases now rejected outright.
- **`commit-draft-agent-session.test.ts` — resolution boundary.** Correct. Now uses `vi.importActual` from inside `packages/extension`, the same boundary `runForkSubagentDraft`'s dynamic import uses, instead of a hard-coded server-local path.
- **`qa/02-server-start.sh` — `curl -s` masking failure.** Correct. Now `curl -fsS`, plus a parser that exits non-zero on unparseable JSON.
- **Two E2E tests overclaimed.** Correct on both. `X12` renamed to assert probe/pin *agreement* (the single-copy invariant lives in the `pnpm-workspace.yaml` override, not in a health assertion), and `F4` now states it covers only the absence half, with KaTeX/Mermaid left to their existing suites. `test-plan.md` rows F1/F4 realigned with what the tests actually assert.
- **Duplicate `D8` + stale Goals.** Both correct — `tool_call terminate` is now `D5a`, and Goals/Non-Goals no longer contradict D3a/D7a/D7b.

## Disputed

**1. "Move OpenSpec artifacts out of `archive/`" (6 threads).** This misreads the rule. `AGENTS.md` says *"Place change artifacts at `openspec/changes/<name>/` (never under `active/`/`archive/`)"* — that governs **authoring a new change**, so nobody hand-creates an `archive/` directory. Archiving a **completed** change is the sanctioned final step: the `ship-change` procedure makes archive-and-sync a hard gate before merge, and `openspec archive` itself writes to `openspec/changes/archive/<date>-<name>/`. Moving these back would undo the required completion step and leave the change permanently "active". Note the diff also contains the resulting `openspec/specs/**` sync, which is only produced by archiving.

**2. `dox.ts` — "condition override shadowing on runtime support".** Declined, and the approved spec says the opposite explicitly: *"the dashboard SHALL NOT carry a version- or shape-gated fallback branch for it."* Reasons: (a) `pi-api-feature-detection` forbids version-string gating, and a file-*name* convention has no shape to detect; (b) `agentsChain` feeds the dashboard's kb doc-lookup, not pi's context injection — pi owns injection and already resolves the override first (`dist/core/resource-loader.js:32`); (c) a repo that adds `AGENTS.override.md` is opting into 0.84+ semantics, so honouring it consistently is the intended behaviour. On a pre-0.84 pi no such file is loaded by pi at all, which the spec records as the floor-pi scenario.

**3. `agent-session-context-injection` — "archived and canonical disagree".** No longer accurate post-sync. The canonical requirement now carries the *"Pi owns injection; no dashboard feature-detection branch"* scenario, which replaced the older floor-pi fallback scenario; the two texts agree. The comparison appears to be against the pre-sync canonical.

**4. `command-handler.ts` — "expose degraded refresh outcomes to callers".** Deferring, deliberately. The requirement this change lands is that a provider error must not be *swallowed* — it is now surfaced with the provider named, and cancellation is distinguishable from success. Signalling degraded-ness to the **client** means adding a field to the `models_list` protocol message and teaching the UI to render staleness, which is a protocol + UX change well outside a runtime pin bump. Worth doing; not here.

PR #437, opened 2026-08-09
feat(add-folder-actions-menu): collapse the folder header cluster into one menu

robertcsakany:
Implements OpenSpec change `add-folder-actions-menu`.

## What changes

The folder header's trailing cluster — urgency sort, pin, add-to-workspace, remove-from-workspace, open-in-new — collapses into a **single** `mdiFolderCogOutline` trigger that opens a grouped popover, or a full-width sheet whenever the app's existing compound mobile predicate (`<768w OR <600h`) is true. That predicate is reused verbatim, so a short-but-wide window gets the sheet too.

- **New `FolderActionsMenu`** with a fixed host-owned group taxonomy (`workspace` → `directory`); a group renders only when it holds at least one item. Open state is owned by `SessionList` and keyed by scope `folder:<cwd>`, mirroring `addToWsMenuFor`, so a folder row and a same-cwd session card can never co-open.
- **Placement gating preserved, not widened.** Add-to-workspace renders only where the affordance rendered before, remove-from-workspace only on workspace-owned rows, pin only outside a workspace container. The `add-to-workspace-btn-<cwd>` test id and the `AddToWorkspaceMenu` popover survive verbatim via a `node` escape hatch on the item type.
- **`mdiOpenInNew` deleted** (`folder-open-home-<cwd>` is gone). The header row is the only open affordance; the leaf name underlines on hover to say so.
- **`SessionCard.renderAddToWorkspace` removed.** Workspace membership is directory-scoped, so rendering it per session produced N identical buttons with one effect.
- **Directory Settings cog moves off `FolderActionBar`**, which now renders `null` when it holds nothing — the emptiness predicate is composed from `shouldShowProjectInit` / `shouldShowWorktreeInitButton` exported by the two init buttons, so it cannot drift from what they render.
- **Pinned state becomes a non-interactive `mdiPin` indicator** in the name region (not the cluster, which must stay at exactly one control).
- `onOpenPiResources` → `onOpenDirectorySettings` (route unchanged; only the name lagged behind the rendered "Directory Settings" label).

## Testing

All **31 automated test-plan scenarios** implemented — 19 L1 (vitest) + 12 L3 (Playwright).

L3 verified against the docker harness: `directory-home.spec.ts` 9/9, `kb-folder-slot.spec.ts` 3/3, `folder-membership-drag.spec.ts` 8/8. Six pre-existing specs were migrated off the deleted test ids.

Unit suite: 12545 passed. The 11 failures are **byte-identical to the `origin/develop` baseline** (rooted in the unlinked `@blackbelt-technology/pi-dashboard-apple-tools` workspace package) — zero new failures. That same missing package makes `npm run build` fail locally; CI builds from a clean install.

Biome: zero new diagnostics (`FolderActionBar` 2w→0w, `SessionList` 21w→19w).

## Reviewer notes

Two things worth a look:

1. **Archive/sync handling.** Superseded scenario headings in the main specs were pre-renamed so `openspec archive` would pass its name-set guard. Separately, four **still-live** scenarios that the deltas had accidentally omitted were carried into their MODIFIED blocks rather than dropped: the legacy `pi-resources` redirect, the branch-row indent invariant, no-duplicate-pin-on-left, and the pinned-collapsed header.
2. **Two pre-existing findings recorded, not fixed** (both would widen the delta set): the `AddToWorkspaceMenu` popover's remove entry is unreachable on folder rows (top tiers filter workspace-owned folders out, so `owningWsId` is always `null` there), and `mdiCogPlayOutline` still echoes `mdiCog` — design D8 assigns that glyph swap to a later change.

QA/manual tasks (test-plan `manual-only` rows F11, F12) deferred to post-merge verification.


robertcsakany:
@coderabbitai Triage of this review — 1 accepted, 3 disputed, 6 out of scope.

**Accepted and fixed** in 54b33951a:

- *"Make whole-row navigation keyboard accessible"* — correct, and it was a regression this PR introduced. Deleting `folder-open-home-<cwd>` removed the only **focusable** open control, leaving `folder-home-row-<cwd>` as a bare `div` + `onClick`. Link semantics now live on `folder-header-name-<cwd>` (`role="link"`, `tabIndex={0}`, Enter handler, `focus-ring`) rather than on the row itself — the row also hosts the `FolderActionsMenu` trigger, and a `button` may not nest inside a `role="link"`. Recorded in the `directory-home-page` spec + archived delta, with three L1 tests.

**Disputed — "Move this OpenSpec artifact out of `archive/`" (3 threads).** This misreads the convention. `AGENTS.md` governs where a change is **authored**: *"Place change artifacts at `openspec/changes/<name>/` (never under `active/`/`archive/`)"* — i.e. do not hand-create a change inside `archive/`. Moving a **completed** change into `openspec/changes/archive/<date>-<name>/` is the terminal step of shipping, performed by `openspec archive`, which also syncs the delta specs into `openspec/specs/`. That sync is visible in this same diff (`+4 added, ~10 modified, -2 removed` across 8 capabilities). Reverting the move would un-ship the change and orphan the synced specs.

**Out of scope — 6 threads** on files this PR does not touch: `docs/research/user-browser-in-editor-view.md`, `openspec/changes/fix-replay-cache-partial-payload-cursor/`, `openspec/changes/purge-replay-cache-on-reset-paths/`, `openspec/specs/oauth-authentication/spec.md`, `packages/extension/.pi/skills/browser/SKILL.md`, `packages/extension/.pi/skills/browser/scripts/check-panerelay.sh`.

This PR is a single commit against `origin/develop` touching 51 files (`git diff --name-only origin/develop..HEAD`); none of those six appear in it. They belong to `develop` and need their own PRs — the replay-cache and OAuth findings in particular look worth filing.

**Note on CI:** the two failing suites (`session-action-handler-headless-reload.test.ts`, `session-api.test.ts` — `TypeError: piGateway.isSessionConnected is not a function` and `expected 200 to be 409`) fail identically on `develop`'s own CI (run 31263227738). They are inherited, not caused by this branch.

robertcsakany:
**Correction to my previous comment** — I owe CodeRabbit an accurate record on one point.

I claimed this PR was *"a single commit against `origin/develop` touching 51 files"* and that the six flagged files were never in its diff. **That was wrong at the time the review ran.**

What actually happened: this worktree was branched from a local `develop` that was **ahead of `origin/develop`**. When the PR opened, it carried five inherited commits (`7fb76996`, `21d194d2`, `61a843e0`, `a976b4d4`, `ea6fd27d`) and **86 files**. So when CodeRabbit reviewed, those files genuinely *were* in the diff and flagging them was correct behaviour, not a whole-repo false positive. My "out of scope" framing was right about **authorship**, wrong about **what the diff contained**.

Those five commits have since landed on `develop` independently. The live compare is now:

```
GET /repos/.../compare/develop...os/add-folder-actions-menu
ahead_by=2  behind_by=0  files=51
  76af291f feat(add-folder-actions-menu): collapse the folder header cluster into one menu
  54b33951 fix(a11y): give whole-row directory-home navigation a keyboard route
```

None of the six files appear in that diff any more. So the six findings are now genuinely outside this PR — but they are **real findings against `develop`** and should be filed separately rather than dismissed. The replay-cache ordering/fencing pair and the OAuth redirect-base validation in particular look worth their own changes.

The accepted a11y finding and the disputed `archive/` findings in my previous comment stand unchanged.

PR #442, opened 2026-08-09
docs(openspec): archive anchor-chat-selection-against-row-growth + sync specs

robertcsakany:
## What

Archives `anchor-chat-selection-against-row-growth` (shipped in #439) and syncs its delta into `openspec/specs/chat-selection-preservation/spec.md` — **+2 requirements**, 7 scenarios.

Docs-only: 4 file renames into `openspec/changes/archive/2026-08-09-…` plus the spec sync. No source changes.

## Honesty carried into the archived record

Tasks **6.1–6.7** are checked but annotated **`DEFERRED — spec authored, NOT executed`**, and **7.6** likewise (manual QA). The Playwright specs shipped in #439 have never run: the docker browser-E2E harness is broken repo-wide — an **untouched** control spec (`tool-collapse-narration`) fails identically on a pristine tree with a rebuilt image, and the suite is not in CI (`.github/workflows/` runs only `test:e2e:electron`). The unblock recipe and its sanity gate (*confirm the control spec passes first, or a `selection-anchor` failure means nothing*) remain in the archived `tasks.md`.

Marking them silently done would have asserted a completeness that does not exist; leaving them open would have blocked the archive. The annotation is the honest middle.

## One spec correction

The synced "single clock" requirement scoped the synchronous read to consumers *"outside React's render cycle"*. The #439 review disproved that scoping: the sticky-bottom `useLayoutEffect` — squarely **inside** the render cycle — was gated only on the debounced state and still ran `el.scrollTop = el.scrollHeight` on the first frame of a drag (reproduced: `expected 5000 not to be 5000`).

The requirement now covers both kinds of consumer, and adds that a debounced copy MAY exist for effects needing the `→ false` edge but SHALL NOT be the only gate on a suspend. Applied to **both** copies — the main spec and the archived delta — since `openspec archive`'s sync structurally cannot reach the latter.

## Follow-ups (not blocking)

1. Repair the faux round-trip through the dashboard browser path, then run and tick 6.1–6.7.
2. Put `tests/e2e/` in CI — its absence is precisely why this rot went unnoticed.


robertcsakany:
Both Majors were valid and are fixed in `6ac98a28b` (the threads auto-resolved on push, so recording the resolutions here).

**1. "Define compensation by residual shift, not only by resize type"** — the requirement contradicted itself: its opening covers "resizes, is inserted, or is reordered", then the normative sentence narrowed compensation to in-viewport *resizes*. The implementation never worked by mutation type; D1 keys off the anchor's **residual** shift measured *after* the virtualizer applies its own correction, precisely so it never has to mirror TanStack's `start < scrollOffset` predicate. Insertions and reorders were always covered in code — only the spec text was narrow. Reworded to define compensation by residual, and to say why the double-move cannot happen (an already-corrected shift presents ~0 residual). I did not take the committable suggestion verbatim: it left a dangling `so that the` fragment.

**2. "Align the anchor guarantee with the browser scroll clamp"** — correct, the guarantee was unachievable at `scrollTop === 0`. Added scenario **"Correction is clamped at a scroll boundary"**: apply as much as the clamp permits, guarantee holds only that far.

It also pins the part that actually bites, which is not the clamp but the *next* commit: the compensator re-baselines from the **actually applied** delta (`nextTop − applied`, read back from `scrollTop`), not the requested correction. Without that the un-applied remainder reads as fresh drift and is re-issued every commit — a permanent write loop against a wall. That shipped in #439 and is covered by the unit test *"re-baselines from the actually-applied delta when the write is clamped"*, so this is the spec catching up with the code, not new behaviour.

D7 in the archived `design.md` now documents the same boundary explicitly, rather than leaving "the browser's clamp is the only bound" as an unexamined aside — which is what let the spec overpromise in the first place.

All edits applied to **both** copies (main spec + archived delta), since `openspec archive`'s sync cannot reach the latter. `openspec validate chat-selection-preservation --type spec --strict` passes; 18 → 19 scenarios.

PR #445, opened 2026-08-09
fix(client): only persist a replay cursor that descends from a replay

robertcsakany:
## Problem

`browser-gateway` broadcasts every live `event` to **every** browser socket, regardless of subscription, and `useMessageHandler` recorded them unconditionally. So a tab accumulated replay buffers for sessions it never opened, and `replay-persist` derived the cursor from the buffer itself — persisting `{maxSeq: <high>, payload: [1 stray row]}`. A **cursor without provenance**.

On reload the delta subscribe (`lastSeq = maxSeq`) skipped every earlier event, the session rendered a single stray row, and the tab re-persisted the same entry — poisoning itself on every reload. Field case: session `019fd8d5` showed only `/reload completed` while the server replayed 277 events in <100 ms and a fresh browser rendered everything. Clearing site data was the only fix.

## Change

- **Provenance flag** (`descended`) per session in `createReplayPersister`. Set by `seed()` and by a **replay-origin** `record()`; never by a live-origin one. `flush()` persists only when set.
- **Skip, never delete** on the non-descended path — the IndexedDB store is shared across tabs while buffers are per-tab, so an observer tab must not destroy a sibling's valid cursor. Silent, because a non-descended flush is *normal* for every background session.
- **Live-path gap detection**: a live `seq > max + 1` means a frame was dropped (gateway back-pressure), so the buffer stops being persistable until re-seeded. Replay-path gaps stay exempt — compaction legitimately leaves holes.
- **Explicit origin** threaded through the two `record()` call sites. Deliberately *not* gated on `shouldReset`: a compacted or capped cold replay starts past `seq 1` and must still establish provenance, or it would be permanently unpersistable (the exact traffic regression this capability exists to avoid).
- **`REPLAY_CACHE_SCHEMA_VERSION` 1 → 2** so already-poisoned browsers self-heal on upgrade with no user action, via the existing mismatch-purge path.

No protocol change, no server change, no change to `rehydrate-session.ts` semantics.

## Tests

14 L1 vitest scenarios (`replay-persist`, `replay-cache`, `rehydrate-session.poisoned-cache`) — authored first and verified RED, notably `#E1` (broadcast-only never persists), `#E6` (lost live frame voids the cursor), `#X1` (an observer does not destroy a sibling entry), `#E9`/`#P1` (schema bump purges once).

2 new L3 Playwright scenarios in `tests/e2e/replay-delta-on-reload.spec.ts`, green against the docker harness:

- `#F1` — a broadcast-only session (driven over REST so the tab never subscribes) recovers its **full** history after reload.
- `#F2` — a planted `schemaVersion: 1` partial-payload entry self-heals on upgrade.

`#F3` (the pre-existing delta-replay spec) still green. The harness image was verified to actually carry `schemaVersion = 2`, so the L3 passes are not vacuous.

## Follow-up

- The `Manual confirmation` task (originally-affected browser `019fd8d5` renders its full transcript after upgrade without clearing site data) is checked off as **to be confirmed post-merge** — it needs the real affected profile.
- Narrowing the broadcast fan-out to actual subscribers stays out of scope (server-side, own blast radius).

Change: `fix-replay-cache-partial-payload-cursor` (archived + specs synced in this PR).


robertcsakany:
Both findings were valid and are fixed in abc7814e7.

The root cause is exactly as described: `record()` dedups by seq, so a compacted replay whose events all sit at or below the buffered max is discarded wholesale — yet it still marked the buffer descended. A tab that saw live seq 250 and then replayed 5..250 would have persisted the original stray row as an authoritative cursor, reproducing the very poison this change exists to prevent.

Fix: contamination is now tracked explicitly and is **not** clearable by a replay batch.

- A live append onto a non-descended buffer, or a live-gap hole, marks the session contaminated.
- Replay-origin `record()` establishes provenance **only** for an uncontaminated buffer.
- Only `seed()` — which replaces the buffer wholesale, so no unauthorized row survives — clears contamination.

Two L1 scenarios added, both verified RED first: a replay batch does not re-authorize a live-contaminated buffer, and a replay batch does not restore provenance across a live gap. The `session-replay-persistence` spec gained a matching requirement (both the live spec and the archived delta), answering the open question about whether replay-origin records may restore a live-gap-invalidated buffer: they may not.

PR #447, opened 2026-08-09
fix(extension): stop a stale extension ctx from killing the pi process

robertcsakany:
A pi session could answer a prompt and then **die**, leaving the composer disabled. Root-caused from a crash stack captured in the docker harness.

OpenSpec change: `openspec/changes/fix-bridge-stale-ctx-crash/`

## The bug

`autoStartServer()` is long-running and asynchronous. Its `notify` / `onLaunchStart` / `onLaunchEnd` callbacks and its terminal `.then()` / `.catch()` continuation all capture the extension `ctx` and touch `ctx.ui` — and any of them can land **after** the session has been replaced or reloaded.

Since pi 0.84, `AgentSession.dispose()` calls `extensionRunner.invalidate(...)`, after which every `ctx.ui` getter throws (`ExtensionRunner.assertActive`). Inside a floating-promise continuation that throw has **no owner**: it surfaces as an unhandled rejection and terminates the process.

```
The quick brown faux jumps over the lazy dog.          <-- model answered fine
.../pi-coding-agent/dist/core/extensions/runner.js:360
Error: This extension ctx is stale after session replacement or reload. ...
    at ExtensionRunner.assertActive (.../runner.js:360:19)
    at get ui (.../runner.js:465:24)
    at stopSpinner (/app/packages/extension/src/bridge.ts:2553:11)
    at /app/packages/extension/src/bridge.ts:2610:21
```

## The fix

`runUiSafely()` — a small helper applied at the three reachable `ctx.ui` sites.

Staleness is **not observable**: `assertActive()` throws rather than returning a verdict, and pi exposes no `isActive` / `isStale` predicate. There is no test-then-call option, so attempt-and-swallow is the only available guard. It is sound here because every guarded site is UI presentation for a session that no longer exists — dropping the work is the correct outcome, not merely the safe one.

Two deliberate constraints:

- The guard wraps **only** `ctx.ui`. Auto-start's own errors (discovery, launch, port reconciliation) still reach its existing `.catch()`.
- `stopSpinner` now clears its interval **before** the guarded call, so a stale ctx still releases the 1s label refresh.

## Verification

Same repro, after the fix — clean answer, clean exit, no stack:

```
$ pi -p "[[faux:plain-text]] go" --model faux/faux-1
[dashboard] sendFlowsList: 1 flows, sessionId=019fe855
The quick brown faux jumps over the lazy dog.
```

Tests import the **real** helper (not a mirror) and have fails-on-revert teeth: defeating the `try/catch` turns 5 of 8 red — including the `unhandledRejection` assertion — and restoring it leaves `git diff` clean of production changes. `packages/extension` otherwise green (2 pre-existing `pi-version-tracker` failures, unrelated).

## Scope note

This fixes a real crash, but it is **not** the cause of the faux E2E failures I originally attributed it to. Follow-up investigation (recorded in `test-plan.md`) showed spawn, bridge registration, prompt delivery and the faux model all work over REST. The `Dashboard server failed to start: readiness timeout` notice is a cosmetic boot-race artifact, and the remaining E2E failure is client-side (optimistic prompt stuck at `sending`). Tasks 4.4/4.5 are therefore left unticked rather than claimed.

---

## Scope update (ship-it)

`develop` merged in. The change is archived (`openspec/changes/archive/2026-08-09-fix-bridge-stale-ctx-crash/`) and specs synced (`bridge-auto-start-lifecycle`).

**Test-plan rows F1/F2 (`tests/e2e/faux-text.spec.ts`, `faux-ask.spec.ts`) are reclassified OUT OF SCOPE.** They stay red for a *second, independent* client-side fault, proven distinct: driven over REST against the same container, `POST /api/session/spawn` + `POST /api/session/:id/prompt` answer correctly and the session returns to `idle` — spawn, bridge registration, prompt delivery and the faux model are all healthy. Through the browser composer the optimistic prompt stays at `sending`. Tracked by the new `fix-optimistic-prompt-stuck-sending` proposal (included in this PR).

The acceptance gate for this change is therefore **E1–E6 with fails-on-revert teeth** (defeating `runUiSafely`'s try/catch turns 5 of 8 tests red, including E1's unhandled-rejection assertion) plus the in-container repro above.

Full suite: the five `pi-coding-agent` ENOENT failures were worktree install drift (cleared by `pnpm install`); `doctor-route.test.ts` fails a 3s timing budget under machine load — pre-existing and untouched by this diff.


robertcsakany:
Addressed in 30407a8e2:

- **Placeholder Purpose** — filled in `openspec/specs/bridge-auto-start-lifecycle/spec.md`.
- **F1 requirement vs recorded outcome (one source of truth)** — agreed, this was a real inconsistency. The round-trip requirement is now scoped to the **session** (answer returned, session returns to `idle`, process stays alive — exactly what this change proves in-container and over REST). The browser-composer claims moved out: settlement of the optimistic card is now governed by the `optimistic-prompt` capability, which is what actually fails. Mirrored into the archived delta so the canonical spec and the delta cannot drift.
- **Settlement path when acknowledgement does not arrive** + **acknowledgement vs response rendering** — good catches, both folded into the follow-up spec: settlement is now required on *every* terminal outcome (ack / send failure / timeout), `prompt_received{fresh:true}` is named as the settlement point distinct from response rendering, and a reset/replay scenario forbids resurrecting `sending`.
- **Faux E2E should verify settlement** — the scenario now additionally asserts no pending card remains in `sending` and the composer is enabled at the end of the round-trip.

PR #453, opened 2026-08-10
feat(gate-notify-rows-by-level): gate notify rows by level with a re-toned renderer

robertcsakany:
Implements OpenSpec change `gate-notify-rows-by-level`.

Adds `DisplayPrefs.notifyMinLevel` — the first display preference for
extension-authored `ctx.ui.notify` rows. Defaults to `"all"` in every preset,
so **nothing that renders today stops rendering** until a user opts in.

## What

- **shared** — `NotifyMinLevel = "all" | "success" | "warnings" | "errors"`,
  `NOTIFY_MIN_LEVELS`, and one exported predicate `isNotifyRowVisible(row, minLevel)`
  consumed by both chat-view gate sites.
- **server** — `backfillDisplayPrefs` defaults legacy files to `"all"`;
  `setDisplayPrefs` carries the field on both write paths.
- **client** — gate at `isRowVisible` + the `interactiveUi` render branch;
  `NotifyRenderer` re-toned onto `InlineMessage`; a 4-stop selector in Settings
  and in the ⚙ View popover.

## Decisions worth reviewing

- **`success` deliberately outranks `info`** on the ladder `info < success < warning < error`.
  A success notify reports an outcome; info is chatter. Product decision, not a
  property of the level.
- **No `"off"` value.** `errors` is the floor, so a failing extension can always report.
- **Fails open on BOTH inputs.** A row is only treated as a notify when it carries
  BOTH markers (`content === "notify"` AND `args.method === "notify"`) — an ask
  misclassified as a notify would deadlock a session with no visible cause, while
  the reverse is cosmetic. An unrecognized floor degrades to `"all"`; without that,
  a garbage persisted value makes every rank comparison `NaN` and hides even `error`.
- **Two separate rank maps** (singular level vocab vs plural floor vocab) so a
  `"warning"` / `"warnings"` typo cannot silently alias.
- **Both gate sites, asymmetric on purpose.** A row filtered from `displayRows` is
  never counted nor mounted, so `isRowVisible` alone is functionally sufficient; the
  render-branch gate is defensive (mirroring the `rawEvent` precedent). Consequence:
  the row-count invariant passes with *either* site alone, so it does not prove both —
  the render branch has its own dedicated assertion.

## Tests

33 automated scenarios folded from `test-plan.md`: the 4×4 ladder truth table, the
error floor, both fail-open paths, the AND-discriminator, ask safety at the strictest
floor, the two-gate-site invariant, renderer tone/icon/level-word, and the controls.
Full suite **12953 passing**.

## Known-unchecked tasks (deliberate — not oversights)

- **2.33 / 2.34 (L3 browser E2E)** — authored (`tests/e2e/notify-min-level.spec.ts`)
  but NOT executed. Blocked by a **pre-existing harness defect**, not by this change:
  the bridge's `autoStartServer` path resolves its port from `config.json` (which the
  harness leaves without `port`/`piPort`) and never reads `DASHBOARD_PORT`, so a second
  dashboard launches on the default `8000`/`9999` while Playwright drives the real one
  on the harness port. Both answer `/api/health`; browser and session end up on
  different servers, and every faux-driven spec fails with "No response from session" —
  including the unrelated pre-existing `faux-text.spec.ts` canary. Confirmed identical
  with this change's source entirely absent. A follow-up OpenSpec proposal covers the fix.
- **8.2 (local redeploy)** — a deploy step for the local instance, not a PR gate.

Manual QA (8.3/8.3a/8.4) marked done for post-merge verification per the ship convention.

## Note on the second commit

`chore(kb): split packages/shared/src/AGENTS.md into sidecars` is mechanical: adding the
`notifyMinLevel` row pushed that file past the 30 KB `AGENTS_BYTE_CAP` (it is injected
every turn), so the prescribed `scripts/split-large-agents.mjs` promoted 45 heavy rows to
pull-only sidecars. All 52 rows preserved, no content dropped. Kept as its own commit so
the churn stays out of the feature diff.


robertcsakany:
Thanks — the critical one was a genuine bug and is fixed in 788386e4.

**Applied**

- **`in` vs `Object.hasOwn` (Critical)** — correct, and worse than reported. The same flaw was in the shared predicate's floor lookup, where `notifyMinLevel: "toString"` resolved its rank to a *function*, making every `>=` false and **hiding `error` notifies** — the one guarantee this axis makes. Both sites now use `Object.hasOwn`, with regression tests over `toString`/`constructor`/`valueOf`/`hasOwnProperty`/`__proto__` for both the level and the floor.
- `finally` around the `globalThis.fetch` mock.
- `inline-message-log-primitives` base requirement now covers the whole `--severity-*` family — good catch, the ADDED requirement structurally could not reach that line through the archive sync.
- `interactive-renderers/AGENTS.md` stale `_promptBusComponent.props.*` mention removed.
- `docs/AGENTS.md` pipe escaping; `test-plan.md` L1 total 30 → 31 and the future `Generated` date.

**Not applied, with reasons**

- **"Move this OpenSpec artifact out of `archive/`"** — the guideline forbids *authoring* changes under `archive/`; it does not forbid archiving a completed one. `openspec archive` moving `openspec/changes/<name>/` → `openspec/changes/archive/YYYY-MM-DD-<name>/` is the prescribed final step of the ship pipeline, and the specs were synced in the same operation.

- **"Correct the virtualizer test model" (chat-view spec)** — the spec already states exactly this. Lines 15-23: *"Gating only `isRowVisible` is functionally sufficient on its own — a filtered row is neither counted nor mounted — so the render-branch gate is defensive"* and *"the row-count invariant SHALL NOT be treated as coverage of both sites: it is satisfied when either site alone is gated. The render-branch site SHALL be pinned by a direct assertion."* That direct assertion exists (`#F2`).

- **`packages/shared/src/AGENTS.md` pipe escaping + row ordering** — both are **byte-identical to `develop`** and predate this PR; they only surface in the diff because the file was mechanically rewritten by `scripts/split-large-agents.mjs` (the `notifyMinLevel` row pushed it past the 30 KB `AGENTS_BYTE_CAP`). Verified: `git show origin/develop:packages/shared/src/AGENTS.md` has the same unescaped pipes and the same `live-server.ts` / `model-id.ts` / `recovery-timing.ts` ordering. Repairing unrelated pre-existing rows would violate this repo's surgical-changes rule; worth a separate cleanup pass.

- **`purge-replay-cache-on-reset-paths/*` findings** — those artifacts belong to a *different, unstarted* change that rode along on the local `develop` this branch was cut from. Out of scope here; flagged to the author separately.

**Known-unchecked tasks 2.33/2.34** are deliberate, not oversights — the L3 browser E2E is blocked by a pre-existing harness defect (the bridge's `autoStartServer` ignores `DASHBOARD_PORT`, so a second dashboard starts on `8000` and the browser and session land on different servers). Reproduced with this change's source entirely absent, including on the unrelated `faux-text.spec.ts` canary. A follow-up proposal covers it.

robertcsakany:
@coderabbitai full review

robertcsakany:
Round 2 applied in ed9a30d4.

**Applied — both were good catches**

- **`tasks.md` §1.8 "zero consumers"** — correct, and this one mattered. That exact false claim was corrected across the proposal, design and four spec deltas during the pre-implementation doubt review; it survived only in the ground-truth section, where it contradicted task 2.25. Now inventories the real consumers (`Toast.tsx:47-48`, `extension-ui/ToastSlot.tsx:55`), so `NotifyRenderer` is documented as a *new* consumer of an existing triple, not its first.
- **Render-branch vs virtualizer guarantee** — right, and my wording was genuinely wrong. Once `isRowVisible` admits a row the virtualizer has already counted it, so `return null` in the render branch cannot reclaim the estimated slot; "no measured height" was unachievable for that site in isolation. The render-branch site now requires *no rendered element*, and the height guarantee is scoped to the combined case. Applied to both the synced spec and the archived delta.
- Docs style: `## Notify rows` rewritten one-fact-per-line; `docs/AGENTS.md` row shortened ~470 → ~300 chars.

**Still not applied**

- **"Move these OpenSpec artifacts out of `archive/`"** — as in my previous reply: the guideline forbids *authoring* changes under `archive/`. Archiving a completed change there is the prescribed final step of this repo's ship pipeline (`openspec archive` performs the move and the spec sync in one operation). Applying this would mean un-shipping the change.
- **`purge-replay-cache-on-reset-paths/*` (2 findings)** — a different, unimplemented change whose planning artifacts rode along on the local `develop` this branch was cut from. Not in scope for this PR; forwarded to that change's author.

FYI, review also surfaced an unrelated regression I've fixed: the planning commit had swept up a deletion of `.pi/skills/manage-flows/SKILL.md` (357 lines, still present on `develop`). Restored verbatim in 6aaa9fd2.

robertcsakany:
Round 3 applied in 8ba2a674 — this one was a real gap, thanks.

`normalizeNotifyMinLevel` is now exported from shared and used by both the `ChatViewMenu` popover select and the `SettingsPanel` `SelectField`, so each renders the *effective* floor rather than a value that matches no `<option>`.

Worth flagging honestly: the jsdom assertion I first wrote for this was **vacuous**. jsdom resolves an unmatched controlled `<select>` to the FIRST option — which here *is* `"all"` — so `expect(select.value).toBe("all")` passes whether or not normalization exists. That test now pins only the invariant it can actually prove (the rendered selection is a real option), with the caveat recorded inline, and the teeth moved to direct `normalizeNotifyMinLevel` unit tests: every real stop passes through, 13 unusable inputs map to `"all"`, and the result is always a member of `NOTIFY_MIN_LEVELS`. Confirmed fail-closed by mutation — reverting the `Object.hasOwn` guard turns 14 of them red.

Full suite: 12995 passing, build clean.

Remaining open threads are the ones covered in my two earlier replies: the `archive/` placement (archiving is the prescribed ship step, not artifact authoring), the pre-existing `packages/shared/src/AGENTS.md` ordering/pipes and the `model-id` / `node-version` sidecar wording (all byte-identical to `develop`, surfaced only because `split-large-agents.mjs` mechanically rewrote the file), and the `purge-replay-cache-on-reset-paths/*` findings (a different change's artifacts).

robertcsakany:
@coderabbitai full review

robertcsakany:
Round 4 applied in 8f6a078b.

- **`Object.hasOwn` explanation** — you're right, and the mistake was mine in an unusually literal sense: the `sed` I used to restore the guard after a mutation test *also* rewrote the comment's `value in FLOOR_RANK` into `Object.hasOwn(...)`, so it ended up asserting that `hasOwn` matches inherited names. It doesn't. The comment now states the actual history — `in` matched `Object.prototype` names, `hasOwn` consults own properties only, so those fall through to the `"all"` fallback.
- **Test name overclaimed** — agreed. `never yields an undefined notifyMinLevel to the predicate` → `fails open when the store yields no notifyMinLevel at all`. It asserts fail-open for an *absent* floor; it never proved the store cannot emit `undefined`, and the name implied it did.
- **Oversized per-file records** — promoted both to sidecars per the 200-char rule: `InlineMessage.tsx` (637 chars) and `NotifyRenderer.tsx` (972). Dir rows keep a summary + `→ see`; the detail is preserved verbatim in `InlineMessage.tsx.AGENTS.md` / `NotifyRenderer.tsx.AGENTS.md`.
- **`docs/chat-display-preferences.md` line 7** — `## What` split into one fact per line.

Not applied, unchanged from earlier replies: the `archive/` placement (archiving is the prescribed final ship step, not artifact authoring), the pre-existing `packages/shared/src/AGENTS.md` ordering/pipes and the `config.ts` / `model-id.ts` / `node-version.ts` sidecar wording (byte-identical to `develop`; they surface only because `split-large-agents.mjs` mechanically rewrote that file, and editing them would violate this repo's surgical-changes rule), `autoNameSessions` in the `preferences-store` record (pre-existing gap, unrelated to this change), and the `purge-replay-cache-on-reset-paths/*` findings (a different change's artifacts).

Suite: 12995 passing, `kb dox lint` clean for the directories this change touches.

Issue #452, opened 2026-08-10
Shutting down a tmux-spawned session leaks the pi process (tmux is the docker default)


robertcsakany:
Found while running the acceptance run for `fix-e2e-harness-memory-exhaustion`. Looks like a production defect, not just test infrastructure — and plausibly the real cause underneath #433.

## Symptom

Shutting down a session spawned under `PI_SPAWN_STRATEGY=tmux` removes it from the dashboard and reports success, but the `pi` process (~127 MB RSS) keeps running forever, orphaned.

Measured at a single instant inside the E2E harness:

```
tmux panes:      21
resident pi:     21      <- ~127 MB each
server records:  0       <- dashboard has forgotten every one
memory.current:  2550 MiB / 4096 MiB, still climbing
```

One pane per orphaned process; nothing left in the system knows they exist.

## Root cause

`handleShutdown` (`packages/server/src/browser-handlers/session-action-handler.ts`):

```ts
piGateway.sendToSession(msg.sessionId, { type: "shutdown", sessionId });
await headlessPidRegistry.killBySessionId(msg.sessionId);   // headless only
killHeadlessBySessionId(msg.sessionId);                     // headless only
sessionManager.unregister(msg.sessionId);
broadcast({ type: "session_removed", sessionId });
```

Both kill paths are headless-only. A tmux-spawned session is never registered in `headlessPidRegistry`, so the only thing asking it to stop is the advisory gateway message. When that does not cause an exit, nothing escalates — but the record is unregistered and broadcast as removed regardless, orphaning the process.

```
$ grep -rn "kill-window\|kill-session\|killWindow" packages/server/src/
(no matches)
```

**No code path in the server ever kills a tmux-spawned session.**

`POST /api/session/:id/shutdown` shares the same weakness (see #449 for its separate liveness-write divergence).

## Why it matters beyond tests

`PI_SPAWN_STRATEGY=tmux` is the **docker default** (`docker/.env.example:37`, `docker/README.md`). Any dashboard on that default leaks a full pi process on every session shutdown, while the UI shows success. Memory grows until the container hits its cgroup cap.

## Fix options

1. Give tmux-spawned sessions a real teardown — `tmux kill-window` for the session window, or register their PIDs and run the same SIGTERM to 2s to SIGKILL ladder `killBySessionId` uses.
2. Do not broadcast `session_removed` / unregister until termination is confirmed, so a failed shutdown is visible instead of silently orphaning.

[...]

Issue #451, opened 2026-08-10
E2E harness: one worktree's container is destroyed by another worktree's harness lifecycle


robertcsakany:
Observed while running the acceptance work for `fix-e2e-harness-memory-exhaustion`.

## Claim vs observation

`fix-parallel-e2e-docker-collisions` and the `ship-it` skill both state that per-worktree compose-project isolation keeps one worktree's harness from affecting another. Observed behaviour contradicts that.

## Timeline (docker events)

```
1786323727  destroy  pi-dash-test-464109120-pi-dashboard-1   <- worktree A tears down
1786323863  destroy  pi-dash-test-719368873-pi-dashboard-1   <- worktree B (mine), NOT its own action
1786324015  create   pi-dash-test-464109120-pi-dashboard-1   <- worktree A brings up
```

Worktree B's container was destroyed between worktree A's `down` and `up`, mid-test-run, causing every remaining spec to fail against a dead port.

## Ruled out

`docker/test-down.sh` is correctly scoped: it derives the project name from `$PWD` via `derive_project` and runs `docker compose -p "$project" ... down -v`, then removes only `pi-dashboard:${project}`. It cannot reach another project. So the cross-project destruction originates elsewhere (a broader prune, or a manual `docker rm`) — worth finding, because it silently invalidates any long E2E run.

## Aggravating factor

The Docker VM has **8 GB total** and each harness sets `MEM_LIMIT` 4 GiB. Two concurrent worktree harnesses therefore saturate the VM — which can itself induce the memory-exhaustion symptom that #433 is about, from outside the container under test.

## Ask

1. Find what destroys across projects and stop it.
2. Consider refusing to start a second harness (or warning) when one is already up, given the 4 GiB x 2 vs 8 GB arithmetic.

robertcsakany:
Blocking a specific piece of work now: tasks 6.3 / 6.4 of `fix-tmux-session-shutdown-leak` (the full-suite acceptance run — all specs in one container with `globalTimeout` overridden, test-plan #P4 — and recording its spec-level results, which #433 part 1 has been waiting on).

Everything that can be verified without an exclusive VM is done and green:

```
P4 OK: container still healthy, no unexplained daemon restart   (2x15-spec chunks)
P1: early=935MiB late=1020MiB limit=1029MiB                     → flat
P3: resident pi=3 reported live sessions=3 divergence=0         → was 21 vs 0
```

[...]

Issue #450, opened 2026-08-10
playwright.config.ts globalTimeout (15 min) is too short for an unattended full E2E run


robertcsakany:
Split out of `fix-e2e-harness-memory-exhaustion` (task 7.4). Deliberately NOT changed there — that change overrides the value on the CLI for its acceptance run and leaves the committed default alone, so the memory fix and a timeout change cannot be confused for one another.

## Problem

`playwright.config.ts` sets `globalTimeout: 15 * 60_000`. The 87-spec suite cannot finish in 15 minutes: a measured 30-spec chunk alone ran ~50 min pre-fix and did not complete.

So plain `npm run test:e2e` can never produce a full-suite verdict; it dies on the global timeout regardless of product health.

## Evidence

See `openspec/changes/fix-e2e-harness-memory-exhaustion/measurements.md`. The acceptance run in that change has to pass `--global-timeout` explicitly.

## Decision needed

Raise the committed default to something a full run can actually reach, or split the suite into shards with a per-shard budget. Related: #433 part 2 (deciding the CI trigger) depends on which way this goes.

robertcsakany:
Planned in OpenSpec change `stabilize-browser-e2e` (`openspec/changes/stabilize-browser-e2e/`): drop the committed `globalTimeout` (#450), sharded nightly/dispatch/label-gated `ci-e2e-browser.yml` with a merged report (#433 part 2), triage + quarantine-with-issue for the red baseline (#433 part 1), `test-up.sh` oversubscription guard + `docker/harness-audit.sh` events capture (#451). #451 part 1 (the external destroyer) stays open with the audit helper as the next step.

Issue #449, opened 2026-08-10
REST session shutdown omits setLiveness(closedReason:"manual"), leaving REST-closed sessions as recovery candidates


robertcsakany:
Split out of `fix-e2e-harness-memory-exhaustion` (task 7.3). Real production defect; that change routes around it rather than fixing it.

## The divergence

`POST /api/session/:id/shutdown` (`packages/server/src/session/session-api.ts:107`) is a **parallel implementation** of the same action as the browser-protocol `{type:"shutdown", sessionId}` handler (`packages/server/src/browser-handlers/session-action-handler.ts:625`).

The WS `handleShutdown` writes:

```
setLiveness({ live: false, closedReason: "manual" })
```

The REST route does **not**. `handleForceKill` (`session-action-handler.ts:779`) also writes it, so REST is the odd one out.

## Why it matters

`isRecoveryCandidate` (`packages/shared/src/session-meta.ts:193`) gates on `closedReason !== "manual"`. So a session closed over REST:

- stays a cold-start recovery candidate,
- is restored by the boot scan into `GET /api/sessions`,
- reappears as a session the user explicitly shut down.

## Impact on the E2E work

`fix-e2e-harness-memory-exhaustion` reaps per-test sessions over the **WS bus specifically** because of this. Reaping over REST would leave every reaped session a recovery candidate, polluting the session list that change's delta snapshot and residual-session budget both read.

## Fix sketch

Converge the two paths — have the REST route delegate to the same handler as the WS message rather than reimplementing it, so the liveness write cannot drift again.

PR #454, opened 2026-08-10
fix(e2e): per-test session reaping, import guard, and a typecheck gate for tests/

robertcsakany:
Implements OpenSpec change `fix-e2e-harness-memory-exhaustion`. Addresses part 3 of #433.

## The problem, measured

All 90 specs share ONE container with a 4 GiB cap, and a spawned `pi` session costs 150-280 MB. Specs spawned sessions from **138 call sites** and ended essentially none, so the container filled up and the daemon died mid-run. Every remaining spec then failed in ~400 ms — which read like mass regression but was just the harness being down. That is why #433 sat undiagnosed.

Pre-fix baseline, one 30-spec chunk:

| metric | at start | after 114 tests |
|---|---|---|
| `memory.current` | 290.9 MiB (7.1%) | **2638.2 MiB (64.4%)** |
| `pids.current` | 17 | **243** |
| resident `pi` | 0 | **19** |

## What this PR does

- **Automatic per-test session reaping.** Every spec imports `test` from `./fixtures.js`; the fixture snapshots session ids before/after each test body, settles the list adaptively, and shuts down **only the delta** over the WS `shutdown` path (which records `closedReason:"manual"` — REST does not, see #449). Pre-existing harness sessions are never touched.
- **A guard that makes bypass impossible.** `scripts/check-e2e-fixture-import.mjs` fails any spec importing `test` straight from `@playwright/test`. It has **already earned its keep**: merging `develop` mid-change brought in two new specs written the old way, one of which spawns sessions.
- **A residual-session budget** (8, with its derivation recorded) that fails at the spec that leaks rather than collapsing 40 specs later.
- **Fail-loud harness-down.** Declared down only after 3 consecutive probe failures — a slow harness under memory pressure is the measured state right before death, so one timeout must not condemn it. Then one clear failure, rest skipped.
- **`tests/` is now typechecked** (`npm run lint:e2e`, wired into CI + `quality:changed`). It never was. That immediately found 5 real errors, including a popup watcher in `subagent-detail-dialog.spec.ts` that could never fire (`popup` is a `Page` event, not a `BrowserContext` event) — the assertion had never once armed.
- **Out-of-band memory probe** + an L2 smoke script.

## Scope was narrowed mid-implementation — please read

The acceptance run **disproved a premise of the original design**. Reaping releases the session *record*; it does not terminate the *process*. The harness runs `PI_SPAWN_STRATEGY=tmux`, and `handleShutdown`'s only kill paths are headless-only, so no server code path kills a tmux-spawned session.

Measured at one instant mid-run, with the reap working perfectly and reporting **0 budget breaches**:

```
tmux panes:      21
resident pi:     21     (~127 MB each)
server records:  0
memory.current:  2550 MiB, climbing
```

So the **full-run memory guarantee moved out** of this change into `fix-tmux-session-shutdown-leak` (authored in this PR, unimplemented) and issue **#452**. tmux is the docker default, so that is a production defect: every session shutdown leaks a `pi` while the UI reports success.

This PR now claims exactly what it delivers and verifies. `SHIP_IT_BLOCKED.md` records the full evidence and the decision.

## Verification

- L1: **254 passed** in the `scripts` project (30 of them the new reap/guard logic).
- L3: `tests/e2e/session-reap.spec.ts` — **12 passed / 1 skipped** against a live harness; container ended at **538.6 MiB with 0 resident `pi`**.
- `npm run lint` (tsc) and `npm run lint:e2e` clean over all 90 specs.
- Full suite: 12900 passed / **11 failed**, and those same 11 `resource-activation-*` tests **fail identically on `origin/develop`** (verified by checking out develop in the same worktree). Pre-existing, not from this PR.

## Honest notes

- An earlier post-fix measurement in `measurements.md` is marked **RETRACTED**: 112 of its 120 tests died at `browserType.launch` on a reaped `TMPDIR`, so it measured tests not running. The valid run is recorded as Group 3 and it does *not* show a flat curve — which is what exposed the tmux leak.
- Two defects the run itself surfaced were fixed here with L1 coverage: the budget counting closed-but-listed records (96 false breaches), and `shutdownSession` blocking on an ack for an already-dead session.

## Follow-ups filed

**#452** tmux shutdown leak (blocks the memory guarantee) · **#449** REST shutdown omits the liveness write · **#450** `globalTimeout` too short for a full run · **#451** one worktree's harness destroyed another's container.

Parts 1 and 2 of #433 remain open and are now blocked only on #452.


robertcsakany:
@coderabbitai review

PR #458, opened 2026-08-10
feat(show-replay-in-flight-indicator): pill while a replay is still streaming

robertcsakany:
Implements OpenSpec change `show-replay-in-flight-indicator`.

## What

A replay that has already painted its first batch looked *finished* — `loadingHistory` clears on first content, so the transcript sat visibly incomplete with no indicator while the remaining batches streamed.

**Server** — `sendEventBatches` terminates an EMPTY replay payload explicitly with one `event_replay { events: [], isLast: true }`. A warm empty delta / a cold zero-event session previously sent no frame at all, leaving the new flag without a clearing edge. Empty case only: a non-empty payload's final batch already carries `isLast: true`, so an exact multiple of `REPLAY_BATCH_SIZE` must NOT get an appended terminal frame (the E7/E9 falsification rows).

**Client** — second per-session flag `replayInFlight`, armed alongside `loadingHistory` at every subscribe site, cleared only by the TERMINAL batch, the `session_updated { dataUnavailable: true }` edge, or the safety net. Every non-terminal batch (content batches included, not just the empty heartbeat) re-arms the ceiling through the timers ref only — never the setter — so a multi-batch replay costs no extra transcript render.

**UI** — `ChatView` paints an indeterminate pill (`data-testid="replay-in-flight-pill"`, `role="status"`, `aria-busy="true"`, i18n label) after `REPLAY_PILL_DELAY_MS`, as an absolutely-positioned overlay so it cannot perturb scroll anchoring. Mutually exclusive with the loading skeleton, and reset on session switch (the `React.memo`'d instance is reused, rendered without a `key`).

No `packages/shared/` change — no wire-schema change.

## Tests

- `subscription-handler.test.ts` — E5/E6 (empty warm delta, cold empty session terminate) + E7/E8/E9 (batch-size boundaries do not double-terminate). Written first; E5/E6 red before the fix.
- `useMessageHandler.replay-in-flight.test.tsx` (new) — E1–E4 flag divergence, X1–X5/X7/X8 failure + safety-net edges, P1 render-count guard. Written first; 6 red before the implementation.
- `ChatView.replay-in-flight-pill.test.tsx` (new) — F1–F8 handle/a11y, show-delay boundaries, skeleton exclusivity, pending-timer cancellation, session-switch leak.
- `replay-in-flight-pill.spec.ts` (new L3) — F9/F10/F11/X6.

## Deferred

- **9.4 / 9.5 → CI.** Both are blocked locally by pre-existing environment failures unrelated to this diff: the parent `node_modules` is missing the `apple-tools`/`blackhole` workspace links (client build + `plugin-registry` tests fail on a stash-control run too), and in the docker harness the UNTOUCHED baseline `tests/e2e/faux-text.spec.ts` fails (the faux model never replies) while `smoke.spec.ts` passes — so every replay spec, including the pre-existing `replay-delta-on-reload.spec.ts`, is unrunnable there. CI installs fresh; the new L3 spec is validated there.
- **10.1** — per-theme visual inspection of the pill (manual-only in the test plan), post-merge.


robertcsakany:
Addressed the CodeRabbit review. Applied 6, disputed 2.

**Applied**

- **`ChatView.tsx` — reset the pill before the new session paints.** Valid and real. The visible bit is now keyed by session id (`pillForSession`) and compared at render time, so B's very first committed render is already pill-free without reaching for `useLayoutEffect`. Added the requested session-switch test (`F8b`), plus `F8d` proving no A-armed timer resurrects it later.
- **X8 test overclaimed.** Correct — `setup()` always wires the new handler. Renamed to what it actually pins: the pre-existing `loadingHistory` path still handles `event_replay { events: [], isLast: true }` with no in-flight flag armed. A genuine old-client artifact can't be exercised from this tree; noted inline.
- **`tests/e2e/AGENTS.md` ordering.** Row moved after `replay-delta-on-reload.spec.ts`.
- **E2E: create and serialize a real multi-batch replay.** Both halves were right. Now uses `[[faux:long-transcript]]` (~120 turns, comfortably over `REPLAY_BATCH_SIZE`) instead of two `plain-text` turns, and releases server→client frames through a **sequential queue** — the previous per-frame `await` started every delay concurrently, so content and terminal frames could arrive together and the pill was never proven visible *mid*-transfer. The test now also asserts content has painted while the pill is still up.
- **E2E F10: deterministic pill observation.** Interval polling replaced with a `MutationObserver` installed via `addInitScript` before any navigation, so it survives the reload and cannot miss a short-lived pill.
- **E2E F11: geometry.** Added `height`.

**Disputed, with reasons**

- **Mermaid diagrams in the archived OpenSpec artifacts.** These are planning-phase records under `openspec/changes/archive/`, authored before implementation and included here only by the archive move. Rewriting an archived historical artifact isn't in this change's scope; the diagram-style rule applies to artifacts going forward.
- **Condense the `App.tsx` / `subscription-handler.ts` AGENTS.md rows.** Agreed in principle — both rows are long. But they were already long before this change; my edit appends one sentence each. Condensing them means rewriting accumulated change history for files this PR barely touches, which the repo's surgical-changes rule pushes against. Better as a dedicated docs pass (the `kb dox lint` `over-threshold` category already tracks it) than smuggled into a feature PR.

**Not disputed but worth flagging:** `y` is deliberately excluded from the F11 comparison. Replay rows are still landing between the two samples, so a `y` delta would measure content growth rather than layout perturbation by the pill; `x`/`width`/`height` are what an in-flow insertion would move.

PR #460, opened 2026-08-11
feat(server): collapse superseded tool_execution_update events at retention

robertcsakany:
Implements OpenSpec change `collapse-superseded-tool-execution-updates`.

## What

The event store retained **every** `tool_execution_update` tick a subagent produced. Each tick carries the *cumulative* subagent timeline, so a sustained run inflated the session buffer with near-duplicate payloads.

Now the store retains at most **2** updates per `toolCallId`: the pinned creating tick (supplying the reducer's first-wins `type`/`description`) and the newest tail. A predecessor is dropped only when the successor **subsumes** it — a superset gate over `partialResult.details` mirroring the client reducer's accumulative merge, including its `entries` empty-array overwrite guard and *both* sources of the rendered `result`.

**Collapse is retention-only — it never suppresses a live broadcast.**

## Measured (docker harness A/B, 4 sessions x 4 sustained subagent rounds)

| | before | after |
|---|---|---|
| retained `tool_execution_update` / buffer | 36 | **2** |
| share of buffer | 18.4 % | **1.2 %** |
| `storeTrim.collapsedUpdates` | 0 | 136 |

136 = exactly the difference — the counter neither double-counts nor over-reports. Full numbers + honest scope limits: `heap-evidence.md` in the archived change.

## Notes for review

- **Index pruning.** The seq-keyed collapse index rides on the `SessionBuffer` (released on evict/delete), and is now also **pruned on trim** — without that, a long-lived session grew one permanent entry per `toolCallId`, an uncapped map inside a memory-bounding change. Caught in the `review-code` pass; test `X8` fails (500 entries vs a 21-event buffer) without the fix.
- **F4 was vacuous and is rewritten.** It sampled `body.innerText()` for "2 distinct states"; a probe proved the only text changing during a sustained run is the elapsed-time counter (`2s` to `13s`). It now counts received `tool_execution_update` **WebSocket frames** *and* requires a collapse delta in the same window.
- **Harness gained the producer.** `pi-dashboard-subagents` is the only thing that drives a tool's `onProgress`; without it baked in, P2/F4 had no workload at all. Registered so tool precedence is untouched (bridge stays pinned first).
- **Fails-closed proof.** With `dropIfSuperseded` neutered, P2 and F4 both go red.
- **Unrelated flake fixed:** `doctor-route` asserted `elapsed < 3000` and hit 3098 ms under load. Wall-clock cannot distinguish a deadlock from a loaded-but-healthy run; the deadlock is detected by the status row asserted below it, so the bound was widened rather than the goalpost moved.

## Follow-ups filed

`reduce-subagent-details-payload`, `reduce-bridge-tick-bandwidth`, `drop-final-update-on-tool-execution-end`.

## Verification

- `npm test` — **13 214 passing**, 0 failures.
- `npm run build` — green.
- L3 (docker harness): F3, P2, F4 green; verified non-vacuous.

Only remaining task was 9.1 (`test-plan: manual-only`), deferred to post-merge verification.


robertcsakany:
Addressed the actionable findings in a08cdd228 — thank you, the subsumption one was a real defect:

**Applied**
- **`subsumes()` early-return gap** — correct and load-bearing. A plain-string `partialResult` (sets `result`) and a structured one with neither `details` nor `content` (sets nothing) both resolve to "no details", so the short-circuit could drop a result-setting predecessor. Removed it so the gate falls through to the existing rendered-result rule; applied the same fix to the equivalence-test mirror so F1/F2 stay pinned to the real policy. New test `X9` is red before / green after.
- **heap-probe hang** — pending CDP requests now reject on socket close/error and each request is bounded by `--request-timeout-ms`.
- **heap-probe `--top` NaN** — now rejected loudly instead of `slice(0, NaN)` reporting a false "no buffers found".
- **Follow-up specs** — numeric thresholds for cadence (≥ 1 rendered update / 2 s) and payload growth (≤ 2x across 10x entry growth); added the late-joiner empty-state fold scenario; fixed the `heap-evidence.md` path for the archived location.

**Disputed, with evidence**

1. **`tasks.md` 9.1 checkbox** — this is the documented repo convention, not an oversight. `ship-change` step 1 states: *"Flip each deferrable `- [ ]` → `- [x]` (validated post-merge)"* for tasks mapping to a `manual-only` manifest row, which 9.1 is (`(test-plan: manual-only)`). It is also a hard precondition of the archive step — `openspec archive` gates on `Task status: ✓ Complete`, so unchecking it would make the archived change invalid. The deferral is recorded in the commit message, the PR body, and the task's own inline tag.

2. **`memory-event-store.ts.AGENTS.md` verbosity** — this file *is* the pull-only sidecar the concision rule prescribes, not a directory tree row. Per `AGENTS.md`: *"rows >200 chars promote to a pull-only `<File>.AGENTS.md` sidecar; dir row keeps a summary + `→ see`"*. The directory row in `packages/server/src/persistence/AGENTS.md` is already just a summary plus the `→ see` pointer, and the sidecar is 6.7 KB against a 30 KB `AGENTS_BYTE_CAP`. Detail living here is the intended design; moving it into the source would duplicate what the sidecar exists to hold.

Verification: `npm test` 13 215 passing, 0 failures; `npm run build` green.

Issue #459, opened 2026-08-10
Idle reaper and caps reclaim never reclaim a non-headless session (same root cause as #452)


robertcsakany:
Split out of `fix-tmux-session-shutdown-leak` (task 3.8), which fixed the *shutdown* half of the same root cause.

## The defect

`createIdleReaper` and the caps reclaim path both terminate through an injected `killBySessionId`, wired in `packages/server/src/server.ts` to `browserGateway.headlessPidRegistry.killBySessionId` (lines 1355, 1386).

That registry has **no entry for a session spawned under `PI_SPAWN_STRATEGY=tmux` / `wsl-tmux`**, so:

- gear 1 `idle` — never reclaims a tmux session
- gear 3 `phantom` — never reclaims a tmux session
- caps reclaim — never evicts a tmux session

The reaper reports the reap and moves on, exactly as `handleShutdown` used to: the session record goes, the ~127 MB `pi` stays. Blast radius is wider than shutdown's, because reaping is unattended.

## The fix shape (already proven for shutdown)

`handleShutdown` now escalates to `session.pid` with the shared `killProcess(pid, { timeoutMs: 2000 })` ladder — strategy-agnostic by construction, because the server already stores the PID from `session_register`. See `openspec/changes/fix-tmux-session-shutdown-leak/design.md` D6.

The reaper needs the same treatment at its injection site, plus its own tests (reaper gears + caps reclaim), which is why it was not folded into the shutdown change.

## Why it was deferred

`fix-tmux-session-shutdown-leak` is scoped to the shutdown path and is gated on the E2E harness. Widening it to the unattended reclaim paths would have mixed two independently testable behaviours in one change.

Related: #452.

Issue #433, opened 2026-08-08
Browser E2E baseline is red on develop (10 specs) and the harness dies mid-run


robertcsakany:
The browser E2E suite (`tests/e2e/`, docker harness) is **not run in CI** — only `ci-e2e-electron.yml` is wired. As a result `develop` has drifted red without anyone noticing.

## Evidence

While shipping #432 I ran specs against two harnesses: one built from the change, one built from its base commit `05c3af33c`. These fail **identically on both**, so they are pre-existing, not regressions from that change:

- `bus-client-goal-plugin-action.spec.ts` — `client.plugin("flows", …)` expected to throw `NoPluginHandlerError`, does not throw
- `change-summary-table.spec.ts` — a "Pi session spawned" toast intercepts the `changed-files-chip` click (`eml-preview.spec.ts` already carries a `dismissToasts` helper for exactly this)
- `editor-pane.spec.ts` ×3 — `getByTitle('Open README.md')` never appears after the faux tool round-trip
- `file-preview-survives-churn.spec.ts`
- `openspec-artifact-dialog.spec.ts` ×2
- `project-init-button.spec.ts`
- `roles-custom.spec.ts`

## Second problem: the harness does not survive a full run

A full 80-spec run could not be completed locally. The container reaches `(unhealthy)` mid-run and its daemon restarts (`daemon restarted: pid 140 -> 656`), after which every remaining spec fails in ~400 ms — a cascade that looks like mass regression but is just the harness being down. Chunking specs (~30 at a time) with a harness restart between chunks works around it.

## Suggested work

1. Triage the list above — each is either a real product bug or spec drift.
2. Decide whether browser E2E belongs in CI (nightly, or on a label) so this cannot drift silently again.
3. Investigate the container's stability under sustained load — likely session accumulation across specs in one shared container.

robertcsakany:
Root cause of part 3 (the harness dying mid-run) is measured and the first half is fixed. Posting the detail since parts 1 and 2 have been waiting on this.

## What was actually happening

The container is shared by all specs with a 4 GiB cap, and a spawned `pi` session costs 150-280 MB. Specs spawned from **138 call sites** and ended essentially nothing, so sessions accumulated until the cgroup hit its ceiling and the daemon died. Every remaining spec then failed in ~400 ms, which read like mass regression but was just the harness being down.

Measured pre-fix, one chunk of 30 specs:

| metric | at start | after 114 tests |
|---|---|---|
[...]

PR #462, opened 2026-08-11
fix(server): terminate the session process for every spawn strategy, and reclaim spawns that never register

### LINKED ISSUE #449: REST session shutdown omits setLiveness(closedReason:"manual"), leaving REST-closed sessions as recovery candidates

### LINKED ISSUE #452: Shutting down a tmux-spawned session leaks the pi process (tmux is the docker default)

robertcsakany:
Closes #452. Closes #449.

Shutting down a session left its `pi` process running whenever the session was not spawned headless. The record was released and `session_removed` broadcast regardless, so the UI reported a clean shutdown while ~127 MB of `pi` survived — measured mid-run in the E2E harness as **21 tmux panes = 21 resident `pi` = 0 session records**, memory 2550 MiB and climbing.

## What was actually wrong — three layers, only the first was in the plan

**1. `handleShutdown` forgot the PID it already had.** It reached only for headless paths (`headlessPidRegistry`, and `killHeadlessBySessionId`, whose `findPidByMarker(sessionId)` scan returns `[]` because a tmux pane runs `cd <cwd> && pi` with no session id on its command line), then unregistered and broadcast unconditionally. `handleForceKill` 140 lines away had always keyed on `session.pid`. Shutdown now escalates the same way after a bounded 1.5 s grace, and logs an `ORPHANED` diagnostic instead of claiming a clean removal when a process outlives the ladder.

**2. …but the server had no PID for a tmux session at all.** `bridge.ts`'s `session_start` register — a fresh session's FIRST register — omitted `pid`, while the two registers in `session-sync.ts` send it. For a non-headless spawn that message is the only channel by which the server learns pi's pid (`tmux new-window` returns tmux's pid). **Every L1 test passed while the product stayed broken**, because each hand-fed `session_register` a pid the real bridge never sends. Only the L3 harness gate could see it.

**3. A `pi` that never registers was owned by nobody.** With shutdown fixed, the memory gate still failed at +29 %. The survivors were three tmux panes sitting forever on pi's interactive prompt:

```
Trust project folder?
/fixtures/kb-parent
 → Trust
```

Zero open sockets, so no `session_register`, so no session record — unreachable by shutdown, reap **and** idle-reclaim, until the host restarts. `SpawnRegisterWatchdog` saw every one of them and only reported. It now reclaims what it reports, keyed on the spawn token in the process environment.

Three prerequisites for that, each of which broke a working-looking implementation:

- **the token was not unique.** `execSync(cmd, { env })` sets the tmux *client's* env; once a `pi-dashboard` server is running, `new-window` inherits the tmux *server's*, so every window after the first carried the FIRST spawn's token (3 panes, 1 token, measured). Now passed per-window with `tmux -e`.
- **the lookup must narrow to leaf `pi`.** The token is an ordinary env var, so the tmux server, the dashboard's own node process and intervening shells inherit it. An un-narrowed lookup returned **five** pids for one token, and handing that set to the kill path took the whole container down. Narrowed to `comm == "pi"`, plus an unconditional never-self/ppid guard.
- **the `/proc` scan exited 1** whenever the last entry did not match (a `for` loop's status is the last iteration's), so `execSync` threw and the catch returned `[]` for *every* lookup — a watchdog that fired and reclaimed nothing, silently.

Plus: `arm()` disarmed the prior entry for the same cwd, so three concurrent spawns produced ONE diagnostic for THREE leaks.

## Also folded in

- **#449 / task 7.4** — `POST /api/session/:id/shutdown` was a parallel implementation, and this change had *widened* the divergence: WS terminated any strategy while REST still killed headless-only, and REST still omitted the `closedReason:"manual"` liveness write (so a REST-closed session returned as a cold-start recovery candidate). Both entry points now call one `shutdownSession()`; the duplicate is deleted rather than re-synchronised.
- Two bugs in the memory gate itself: `mapfile` (bash 4+) aborted `qa/tests/16-e2e-memory-bound.sh` on a macOS host, and `live_sessions()` read `j.data?.sessions` while `/api/sessions` returns `data` as an **array** — so P3 always compared against 0 and *structurally could not observe this fix*.

## Verification

```
baseline               panes=0  resident pi=0  server records=0  memory.current=1033 MiB
after 5 spawns         panes=5  resident pi=5  server records=5  memory.current=1597 MiB
after shutdown         panes=0  resident pi=0  server records=0  memory.current=1044 MiB

untrusted cwd, 3 spawns → 3 diagnostics → 3 reclaims → panes 0 / pi 0, container healthy
trusted   cwd, 3 spawns → 3 registers, 0 false timeouts, 0 kills
```

`qa/tests/16-e2e-memory-bound.sh`: **P1 935 → 1020 MiB (limit 1029), P3 divergence 0** (was 21 vs 0), **P4 healthy**.

L3 `tests/e2e/tmux-session-shutdown.spec.ts` asserts the container's own process table (`/proc` + `tmux list-panes` via `docker exec`) rather than the session list, because a record disappearing is exactly what the bug already did. It was RED before the bridge fix.

Full unit suite: 13190 passed. The single `doctor-route` timing failure is a pre-existing flake that reproduces identically on `origin/develop`.

Full evidence: `openspec/changes/archive/2026-08-11-fix-tmux-session-shutdown-leak/measurements/tmux-leak-evidence.txt`.

## Deferred / filed

- **#459** — the idle reaper and caps reclaim terminate through the same headless-only `killBySessionId`, so they never reclaim a tmux session either. Same root cause, wider blast radius (reaping is unattended), needs its own gear-level tests.
- **#451** — tasks 6.3/6.4 (full-suite acceptance run + its spec-level results for #433 part 1) need exclusive use of the Docker VM. Marked deferred; P4 passed on every chunked run.


robertcsakany:
Addressed in 21dbfafc7, with three disputes.

**Fixed**

- *Do not remove a session while its PID is still alive* — real catch, and a genuine spec/impl divergence. Resolved toward **telling clients** rather than withholding the removal: a new `session_orphaned` message (session id + surviving pid) is now broadcast **alongside** `session_removed`. The record is still released, because retaining it wedges the session in the UI with no way to clear it but force-kill and stalls consumers that await `session_removed` per session (the E2E reap) — design D3 rules that out explicitly. The requirement text now states that contract instead of contradicting it. Tested by stubbing the two OS primitives (a process that survives SIGKILL needs uninterruptible sleep, which no test can arrange) and running `shutdownSession` for real.
- *Reconcile the no-PID fallback* — added a scenario: with no recorded pid there is no process identity, so the universal-termination guarantee cannot apply; the server degrades to closing the bridge and releasing the record, and the **spawn-register watchdog** is the mechanism that reclaims such a process, since it holds the correlation token the session record lacks.
- *Fail when the session API response is invalid* — agreed and fixed: `curl -fsS`, and the parser exits non-zero rather than printing a count it did not derive from a real session array. A silent `0` is indistinguishable from a clean container and would record a false P3 divergence — precisely the bug class this gate exists to catch (it previously read `j.data?.sessions` against an array response and *always* reported 0).
- *AGENTS.md records* — rows trimmed and `tmux-session-shutdown.spec.ts` moved into path order.
- *MD040* — language tag added.

**Disputed**

- *Place OpenSpec artifacts outside `archive`* — the change **is** archived (`openspec archive` ran as part of shipping, moving `openspec/changes/fix-tmux-session-shutdown-leak/` → `openspec/changes/archive/2026-08-11-…/` and syncing the deltas into `openspec/specs/`). Its measurements move with it by design. `README-superseded.md` is deliberately placed inside the *other*, already-archived change: its whole job is to label those retracted numbers where a reader will find them.
- *Use one canonical teardown design* — the superseded tmux-window plan is retained on purpose, and its status line says so. It is not stale text: D5 was first declared superseded by D6, then **reinstated** in part, because a spawn that never registers has no session record and therefore no PID, so the correlation token really is the only handle for it. Deleting the wrong turn would delete the reason the final design has two mechanisms instead of one.
- *Correlate the spawned session with `requestId`* — exact `spawn_result` correlation needs the server's headless strategy, which the harness build does not use; the existing exemplar (`tests/e2e/session-reap.spec.ts`) resolves by snapshot difference for that reason, and this spec follows it. Playwright runs the harness with `workers: 1` + `fullyParallel: false`, so there is no concurrent worker whose session could be selected.

robertcsakany:
Round 2 addressed in 87bfa8585.

- *Handle `session_orphaned` in the browser* — correct, the signal existed only in the protocol. `useMessageHandler` now toasts it (error variant) naming the surviving pid, with a test pinning the ordering: the orphan warning arrives first and the following `session_removed` must not suppress it, while a clean `session_removed` alone shows nothing.
- *Make `session_removed` semantics consistent* — agreed; the heading contradicted its own body. The requirement is now *"A removal SHALL never imply a termination the server did not verify"*, and it states outright that `session_removed` means the session RECORD was released and is not an assertion about the process; the orphan signal is what distinguishes an unconfirmed termination. Added a scenario obliging the client to surface it.
- *Assert the escalation and record-release calls* — done: the test now asserts `killProcess(pid, { timeoutMs: 2000 })` (the shared ladder) actually ran and that `unregister` was called, rather than trusting `session_removed` on the wire as proof of either.

Full suite: 13294 passed, 0 unhandled errors, `tsc --noEmit` clean.

The three earlier disputes stand as explained in the previous comment (archived-artifact placement, the deliberately-retained D5, and `requestId` correlation being unavailable under the harness's tmux spawn strategy).

PR #463, opened 2026-08-11
feat(fix-ended-session-missing-endedat): evidence-based endedAt for every ended session

robertcsakany:
Implements OpenSpec change `fix-ended-session-missing-endedat`. Archives + syncs specs.

## What

A session record with `status: "ended"` now **always** carries an `endedAt`, and the timestamp comes from **evidence** rather than reconstruction time.

- **New** `packages/server/src/session/derive-ended-at.ts` — the single evidence rule. Precedence is fixed: recorded last activity → transcript last-write time → `startedAt`. It never returns `Date.now()`. `readJsonlMtime` moved here so there is exactly one transcript-stat path (`session-scanner` imports it).
- **`memory-session-manager`** — `ensureEndedAt` enforces the invariant at every entry point into the map, including `restore()` (a bare `sessions.set` that `update()`/`unregister()` could never see). Fills only when absent, so an explicitly supplied `endedAt` is always preserved.
- **`unregister(id, { witnessed })`** — the value rule is keyed on **witnessed vs reconstructed**, not on which function ran. The default stays `Date.now()` for an observed ending; `{ witnessed: false }` routes heartbeat/grace/ping expiry and history register→unregister through evidence derivation.
- **Disk-derived writers** — `session-bootstrap` stats its own `sessionFile` (it sets no `lastActivityAt`), `session-scanner` supplies `endedAt` on both the `sessionFromMeta` and no-meta branches, and `server.ts`'s boot normalisation uses the same rule instead of `Date.now()`.

This is **not** a liveness change. Liveness is `status` everywhere; no code treats a missing `endedAt` as live, and no liveness site was touched.

## Design constraints honoured

- **D1a** — `restore()` enforcement does **not** fire `onChange`. Emitting there would `moveToFront` + broadcast `sessions_reordered` for ~3,300 records at boot, churning the stored order this change exists to protect. Proven fails-on-revert at L1 (E12b).
- **Hot-path cost** — the `status`/`endedAt` conditional short-circuits before any filesystem stat, because the same enforcement sits on `update()`, which runs on every activity event (E12).
- **No overwrite** — derivation only fills an absent value.
- Out of scope and untouched: `terminal-manager` (different type, no `endedAt`), any client/wire-protocol change, any bulk migration.

## Tests

18/18 automated scenarios from `test-plan.md`, folded 1:1:

- **L1** (`packages/server/src/__tests__/ended-session-endedat.test.ts`) — E1–E12, X1–X4 (+E12b for D1a). Teeth verified: **10 of these fail** when the enforcement is reverted.
- **L3** (`tests/e2e/ended-session-endedat.spec.ts`) — F1, F2 against the docker harness. Port and compose project read from `.pi-test-harness.json`; nothing hardcoded.

**Known limitation, stated rather than implied:** test-plan #F2's "zero `sessions_reordered` frames during the restore loop" is *structurally unobservable* at L3 — the boot restore loop runs at `server.ts:364`, before `sessionManager.onChange` is assigned at `:391` and before the WS server accepts connections, and `moveToFront` is idempotent for an id already at the front. F2 therefore ships as an honest end-to-end guard on the persisted ended-tier order (with `completedFirst` armed so the `moveToFront` path is live); D1a's actual guarantee is the L1 E12b test. The limitation is documented in the spec file itself.

## Deferred

Left unchecked deliberately, per maintainer direction:

- **§8 Manual verification** (8.1, 8.2) — must run against the **live** instance after deploy: confirm the count of `status:"ended"` records without `endedAt` goes 34 → 0, and that the live session count (which is `status`-derived) does not move.
- **§9 Follow-up** (9.1, 9.2, 9.3) — open questions explicitly out of scope here: whether transcript mtime or an explicit last-entry timestamp is the better evidence source; that `extractTimestamp` returns `Date.now()` for an unparseable filename so the `startedAt` fallback can itself be reconstruction time (design.md D2a); and that the `live`/`liveEpoch` marker remains unexamined.

## Notes for review

`npm test` on this branch shows 12 pre-existing failures — 11 in `resource-activation-{toggle,routes}` and 1 in `skill-frontmatter` (caused by a local untracked skill dir, absent in CI). Verified pre-existing: reverting every source file this change touches to `origin/develop` reproduces the identical 11 resource-activation failures. No file in those suites is touched by this change.


robertcsakany:
Thanks — 4 of 7 applied in 6d674c82d, 1 disputed, 2 declined with reasons.

## Applied

- **Preserve an existing `endedAt` during repeated unregister** (Major) — agreed, and it was a real inconsistency with the `ensureEndedAt` preservation rule. `unregister()` now only stamps when `endedAt === undefined`. Added a fails-on-revert L1 test (`a duplicate unregister does not move an existing timestamp`) covering both the witnessed and `{ witnessed: false }` paths; verified it goes red when the guard is removed.
- **Make F2 self-contained + handle the deferred restart promise** — applied both halves. F2 now calls `ensurePinned()` itself, and the restart promise captures its error and rethrows after `await` so it can no longer mask the real failure.
- **Cache the container id and bound the `docker` calls** — applied, plus `timeout` on both `docker ps` and `docker exec`.
- **Shorten the `tests/e2e/AGENTS.md` row** — applied; promoted to `tests/e2e/ended-session-endedat.spec.ts.AGENTS.md` with a one-line dir row, per the repo's >200-char sidecar rule.

## Disputed

- **"Move the OpenSpec change out of the forbidden `archive` directory"** (Major) — this is a misreading of the convention. `AGENTS.md` forbids **authoring** a new change under `active/`/`archive/`; it does not forbid **archiving a completed one**. The documented ship pipeline (`ship-change` step 3, and `openspec archive` itself) moves a finished change to `openspec/changes/archive/YYYY-MM-DD-<name>/`, and the step 8.5 gate explicitly *requires* the source dir to be gone and the archive dir to exist before merge. Every previously shipped change in this repo lives there. Reverting this would break the archive+sync gate, so it is not actionable.

## Declined

- **"Align heartbeat and grace-period expiry with the inferred-ending rule"** (Major) — the **code is already correct**: every heartbeat/grace/sleep-recovery/ping-timeout site passes `{ witnessed: false }` and therefore uses the full `lastActivityAt → transcript mtime → startedAt` precedence, and there is a dedicated test (X1) asserting a timeout-inferred ending records `T − 5min`, not the detection time `T`.

  You did however catch a genuine **documentation** contradiction: `proposal.md` lists "a heartbeat expiry" among the witnessed examples, which contradicts `design.md` D1 and the spec (both of which classify it as inferred). The implementation follows design.md/spec. That file is now an **archived** planning artifact — a historical record of what was proposed — so it is deliberately not being rewritten post-archive. Flagged to the maintainer instead.

  On `spec.md#L248-L251`: the timeout scenario says "derived from the session's last activity" while the parent requirement immediately above it already states the full three-step precedence for *all* non-witnessed endings. The scenario is a specialization of that rule, not a competing one, so it is not incorrect — noted as a wording improvement for a future change rather than a defect.

- **Wording nits on `proposal.md` / `design.md`** ("describe the rule as filling missing timestamps", "limit the impact statement", "add a language tag to the fenced block") — same reason: these are archived planning artifacts. Rewriting a shipped proposal to read better after the fact would falsify the record of what was actually proposed and reviewed.

robertcsakany:
Not actionable — `See change: <name>` is this repo's established, documented convention, not a grammatical slip.

- **5730** occurrences of `See change:` across `packages/`, `tests/`, `docs/` and `AGENTS.md`.
- **0** occurrences of the suggested `See the change:`.
- `AGENTS.md` codifies it in the Documentation Update Protocol: the per-file purpose row "carries everything per-file: one-line summary, key exports, contracts, `See change:`", and `kb agents <path>` is documented as returning "purpose + exports + `See change:`".

It is a structured marker token (grep-able, machine-read by the `kb` doc tooling), not prose. Changing it in this one file would make the new code the only inconsistent site in the repo and would break the marker's uniformity for tooling. Leaving as-is.

PR #465, opened 2026-08-12
feat(repair-tool-error-surfaces): tokenize the tool-result error surfaces

molnar-botond:
Implements OpenSpec change `repair-tool-error-surfaces`.

## What

Reported as *"the bash tool result card error is completly red and unreadable"*,
then *"Split into sections, is the markdown there generated?"*. Three stacked defects
on the ctx error card, two of them shared with five sibling surfaces.

1. **Contrast.** The six governed tool-result error surfaces shipped dark-mode-tuned
   raw red literals. All 7 measured cells failed the 3:1 floor in light mode; 0 failed
   in dark. They now derive from `--severity-error-{bg,fg,border}` — the token set
   `unify-message-severity-colors` already established. No new token, no new gate.
2. **Flat red wash.** Severity now sits on the *chrome* (border, fill, label); the body
   renders in `--text-secondary` on `--bg-code` — the rule the sibling `receivedArgs`
   block six lines below the bug already followed.
3. **Raw markdown never rendered.** The ctx runtime error body is a context-mode
   execution result, emitted through `LinkifiedText` (URLs/paths only), so the fence
   markers and `Exit code:` / `stdout:` / `stderr:` structure showed as literal text —
   while the **success** path of the same component renders fenced code through
   `CodeBlock`. The parser now extracts `command` / `language` / `exitCode` / `stdout` /
   `stderr` / `preamble`, and the card renders the command through that same `CodeBlock`,
   an `exit <n>` badge, and each stream as a labelled section.

Unstructured runtime errors degrade to the flat message; `preamble` keeps prose above
the dump so no text is dropped. Parser stays pure and React-free.

## Regression guard

`scripts/__tests__/severity-literal-guard.test.mjs` fails when a raw `red-<NNN>`
literal re-enters a governed file (verified red at the parent commit: 10 hits).
Escape hatch: a `severity-exempt: <reason>` marker, used by the destructive stop
button in `ToolCallStep`.

Design **Q1 resolved**: the guard lives in `npm test`, not `check-conventions.mjs` —
that file caps itself at four rules and runs only in the ship gate, whereas `npm test`
is what CI, publish and `quality:changed` all invoke. Same shape as the existing
`scripts/__tests__/repo-hygiene.test.mjs`.

## Verification status — please read

- Client test project: **0 failures**. Client build green. Biome findings unchanged
  (18 before, 18 after — no new ones). `tsc` + `lint:e2e` clean in scope.
- The local `npm test` gate is **red for 15 pre-existing environmental files**
  (missing `@earendil-works/pi` in node_modules, `Jimp is not a constructor`,
  `spawnSync` ENOENT, libreoffice-dependent office-preview). None of them import
  anything this PR touches. **CI is the authoritative gate for those.**
- **Task 5.1 (e2e contrast probes) is written but never executed.** The docker
  harness restart-loops on the dev host (`cannot mount overlay read-only` at
  test-entrypoint), so no spec could run. The 180 contrast cells are asserted by
  construction only. This is the one thing worth a second look before merge.

QA/manual tasks (§6) deferred to post-merge verification.


molnar-botond:
@coderabbitai full review

molnar-botond:
@coderabbitai full review

molnar-botond:
Applied 6 of the 12 findings (see 5a408ce3). Explicit dispositions for the rest:

**Applied**
- `CtxToolRenderer.tsx:231` — exit badge double-tint. Real defect, and it also invalidated the e2e model; fixed both the badge (now on `--bg-code`) and the `TOOL_SURFACES` entry.
- `severity-literal-guard.test.mjs:36` — `red-\d{3}` → `red-\d{2,3}`; header + `scripts/AGENTS.md` row updated to match.
- `severity-literal-guard.test.mjs:83` — correct, the test was vacuous. Rewritten to actually drive `rawRedViolations` over every tracked client `.tsx` with a non-empty-sample assertion.
- `severity-contrast.spec.ts:249` — `parseColor` now throws instead of silently measuring black.
- `severity-contrast.spec.ts.AGENTS.md:5` — row corrected; the floor is 3:1 for `accent` pairs and theme-relative for `base` pairs.
- archived `tasks.md:9` — correct and important. 5.2 claimed `npm test` was green; it was not. The real red list and its provenance are now recorded.

**Not applied**
- *"Move the OpenSpec artifacts out of `archive/`"* — false positive. `openspec archive` places a **completed** change at `openspec/changes/archive/YYYY-MM-DD-<name>/` by design, and the ship pipeline requires that move before merge. The guideline being cited ("never under `active/`/`archive/`") governs where a change is **authored**, not where a finished one is archived.
- `mockups/retry-visibility/index.html`, `openspec/changes/unify-retry-visibility/{design,specs,tasks}` (4 comments) — not part of this change. They arrived in this branch via the `origin/develop` merge and belong to `unify-retry-visibility`. Editing them here would be scope creep.

PR #466, opened 2026-08-12
feat(upgrade-model-selector-primitives): surface provider refresh failures + shell-bound selector primitives

molnar-botond:
Implements OpenSpec change `upgrade-model-selector-primitives`.

## What changes

- **Extension — refresh failures are surfaced, not discarded.** `request_models` now keeps the `reportRefresh()` result and attaches per-provider failures to `models_list` as `refreshErrors`. Omitted (never `[]`) on a clean refresh, and never populated by a bare abort, so the happy-path message is byte-identical to before. A failed refresh stays degraded, not fatal — the last-known catalogue is still served.
- **Shared — additive protocol + contract changes.** New `ProviderRefreshError`; optional `refreshErrors` on `models_list` (extension→server and server→browser); the server forwards it verbatim. New `ui:thinking-level-selector` primitive key + contract, and an optional `placeholder?` on `UiModelSelectorProps`.
- **Client — the manual footer refresh button is retired.** It fired the same request the open transition already sends, and its busy indicator implied staleness right after a refresh. The open transition is now the sole trigger; the `refreshing` state, the `models`-identity clear effect, and the 10 s safety timeout are removed with it. The vacated footer slot renders a non-blocking notice naming each failing provider (never a toast — the refresh fires on every open).
- **Client — `OpenSpecRunConfigContext` → `ModelConfigContext`**, no compatibility alias; every consumer and the test harness migrated.
- **Client — shell-bound primitive registrations.** `ui:model-selector` and `ui:thinking-level-selector` register wrappers (new `lib/plugins/shell-primitives.tsx`) that inject `favorites` / `onToggleFavorite` / `onRefresh` from `ModelConfigContext`, so plugin surfaces gain favorites + refresh-on-open with no plugin-side wiring. With no selected session the context is absent: no favorites, no refresh, caller's list still renders.

Every added prop and protocol field is optional — an older bridge simply never sets `refreshErrors` and the footer stays silent.

## Notes for review

- The spec delta for the refresh requirement is **REMOVED + ADDED**, not MODIFIED: the old requirement's premise (a user-activated footer control with a busy state) is deleted, and the surviving behaviour is restated with the open transition as its trigger.
- The design assumed an open-transition auto-refresh effect already existed in `ModelSelector`; it did not. It was added here, which is what the requirement specifies.
- The registration wrappers were extracted out of `main.tsx` into `lib/plugins/shell-primitives.tsx` so the shell-binding is testable without booting the app.

## Verification

- Unit suites green; the 15 failing test files on this machine are pre-existing/environmental (Jimp missing, worktree without `node_modules`) and identical on `develop`.
- Playwright E2E run against the docker harness: 55 passed, 2 failed — both failures (`bus-client-goal-plugin-action`, `change-summary-table`) reproduce identically on a clean `origin/develop` baseline harness, so they are pre-existing and unrelated.
- QA/manual tasks (7.3–7.5) deferred to post-merge verification.


molnar-botond:
@coderabbitai full review

molnar-botond:
Round-1 review fixes pushed in 026e914c. Triage of the remaining threads:

**Applied**
- `command-handler.ts` — an aborted refresh no longer populates `refreshErrors` even when the result also carries provider errors, plus a test for the both-flags case. Good catch: the requirement makes abort log-only unconditionally, and my mapping read `errors` without checking `aborted`.
- `request-models-refresh-errors.test.ts` — now asserts the whole last-known catalogue via a stable `{provider,id,name}` projection instead of `length > 0`.
- `useMessageHandler.ts` — bridge-supplied `refreshErrors` entries are validated at the trust boundary; malformed entries are dropped.
- `App.tsx` — `modelRefreshErrorsMap` is cleared on server switch.
- `packages/client/src/AGENTS.md` — `thinkingLevelSelector` (and `logBlock`) added to the main.tsx registration list so the exhaustive claim holds.

**Skipped, with reasons**
- *"Move this artifact out of `archive/`"* (3 threads on `openspec/changes/archive/2026-08-12-upgrade-model-selector-primitives/**`) — false positive. The convention forbids *authoring* change artifacts under `archive/`; moving a completed change to `openspec/changes/archive/YYYY-MM-DD-<name>/` is the prescribed terminal step of the ship pipeline. CodeRabbit tagged these `[skip_comment]` itself.
- `openspec/changes/repair-tool-error-surfaces/**` (4 threads) — not this PR's work. Those artifacts arrived via the `origin/develop` merge and belong to change `repair-tool-error-surfaces`.
- `packages/client/src/AGENTS.md#L7` `oute` typo — pre-existing (commit e20cd5ae, PR #220), outside this change's scope; left untouched rather than mixed into an unrelated diff.
- `openspec/specs/plugin-ui-primitive-registry/spec.md#L342` (heavy lift) — an IntentRenderer-parity scenario for `ui:thinking-level-selector`. Deferred: the spec is already archived/synced, and adding a requirement scenario is new spec scope rather than a fix to this change.

molnar-botond:
@coderabbitai full review

molnar-botond:
Round 2 (the review-loop's hard cap) — pushed 6ad1402b.

**Applied**
- `request-models-refresh-errors.test.ts` — documented the precondition the full-catalogue assertion rests on: `request_models` pipes the list through `filterByEnabledModels`, which reads `<HOME>/.pi/agent/settings.json`, and the repo pins `HOME` to an ephemeral tmp dir for the whole run (`packages/shared/src/test-support/setup-home.ts` + the `npm test` HOME override). So no host `enabledModels` pattern can filter `LAST_KNOWN`. The coupling is real; it is neutralised by the harness rather than by mocking.

**Deferred, with reasons**
- *"Make the peer range match the refresh API"* (`command-handler.ts`, Major/heavy-lift) — correct observation, deliberately **not** fixed here, on two grounds. (1) **Scope/safety:** raising a `peerDependencies` floor is a dependency/packaging change affecting every consumer of the extension; the ship pipeline explicitly forbids auto-applying dependency changes, so it needs its own change rather than a drive-by bump. (2) **The old-peer behaviour is specified, not broken:** on a pre-0.84 registry `refresh()` yields nothing to inspect, `reportRefresh` returns `undefined`, and the handler returns a normal `models_list` with `refreshErrors` omitted and no warning — that is the requirement's "Refresh outcome is unavailable" scenario, and it is covered by a test. You lose refresh-error reporting on an old peer; you do not get a broken model list. Note the async-refresh dependency predates this PR (it arrived with change `update-pi-core-0-84-adopt-apis`, which introduced `model-refresh.ts`), so the peer-range skew is pre-existing rather than introduced here. Filed as follow-on.
- `packages/client/src/AGENTS.md#L7` `oute` typo — pre-existing (commit e20cd5ae, PR #220); left out of an unrelated diff.
- `openspec/changes/archive/**` "move out of archive/" (3) — false positive; archiving to `openspec/changes/archive/YYYY-MM-DD-<name>/` is the prescribed terminal step. CodeRabbit tagged these `[skip_comment]` itself.
- `openspec/changes/repair-tool-error-surfaces/**` — arrived via the `origin/develop` merge; not this PR's work.
- `openspec/specs/plugin-ui-primitive-registry.md#L342` — IntentRenderer-parity scenario for `ui:thinking-level-selector`; new spec scope, not a fix to this change.

PR #467, opened 2026-08-12
fix(optimistic-prompt): always settle the optimistic prompt

robertcsakany:
Implements OpenSpec change `fix-optimistic-prompt-stuck-sending`.

## Why

The composer's optimistic prompt bubble could hang at `sending` with the
composer disabled, and the faux browser-E2E specs were red.

## Root cause (proven per-leg in the docker harness)

The bridge's serialized inbound pump awaits each handler to completion.
`request_models` awaits a provider catalogue refresh that never settled, so the
pump wedged: every later `send_prompt` was queued and never dispatched — no
`prompt_received` ack, no `message_start`, nothing reached pi. The same session
answered fine over REST (which does not go through the queue). Full evidence
table in the archived change's `root-cause.md`.

## What changed

- `request_models` moved to the immediate (non-serialized) lane; `reportRefresh`
  bounds its wait (`REFRESH_TIMEOUT_MS` 10s) so no refresh can wedge a caller.
- `CommandInput` takes the pending-prompt **status**, not a boolean — only
  `sending` gates the composer, so an acked prompt re-enables it immediately.
- `PendingPrompt.status` gains `failed`: the existing 30s safety timeout marks
  the bubble failed (text preserved) + keeps `lastError`, instead of silently
  dropping the user's text. New `ChatView` render arm; the timer arms only on
  `sending`, so a failed bubble is never wiped.
- `applyPromptReceived` treats every settled status as terminal.
- Reset/replay no longer restores a `sending` bubble; `sent`/`failed` carry as
  before (`preserve-pending-prompt-across-replay` preserved).

## Tests

- 14 new/updated L1 rows (composer gate, reducer settlement, timeout, carry
  rules, failed render arm), verified RED before the fix.
- `tests/e2e/faux-text.spec.ts` + `faux-ask.spec.ts` green against the harness,
  asserting ACK-driven settlement inside 15s (`TIMEOUT_MS` is 30s) and that the
  failed arm never appears.

QA/manual task 4.6 (visual read of the failed bubble + banner) deferred to
post-merge verification.


robertcsakany:
Review round 1 triage — applied 7, disputing 2.

**Applied** (commit `a44f0661`):
- `applyPromptReceived` now checks the terminal-status guard *before* the `fresh:false` race-drop, so a late unfresh ack cannot clear a settled bubble (+ test).
- `openspec/specs/optimistic-prompt/spec.md` synced with the shipped behaviour: `failed` in the status union, the input gate keyed on `sending` rather than mere existence, the reset/replay carry rule limited to settled prompts, the timeout settling instead of clearing, and the third render arm.
- AGENTS rows: `usePendingPromptTimeout` arming rule, `request_models` bypassing the serialized lane, and the `oute` → `route` typo.

**Disputed:**
1. *"Move the complete OpenSpec change directory out of `archive/`"* — intentional. This repo archives a change as part of shipping it (`ship-change` step 3: sync delta specs into `openspec/specs/`, then move the change to `openspec/changes/archive/YYYY-MM-DD-<name>/`). The change is implemented and merged here, not in flight.
2. *"Make the acceptance checks prove ACK settlement" via a `sent` status assertion* — the transient `sent` tick is not a safe observable: the user `message_start` clears the bubble outright, and against the faux model that can land in the same frame as the ack, so asserting the tick would be racy. The composer gate is the stable equivalent — it re-enables **only** when the prompt is no longer `sending`, inside 15s while `TIMEOUT_MS` is 30s, plus an assertion that the failed arm never appears. Rationale is now a comment in both specs.
3. The design-doc contradictions (failure-surface, carry rule) are in the **archived** `design.md`, which records the deliberation; both were resolved explicitly under its "Open Questions" (two surfaces; narrow `sending`-only carry rule). The authoritative spec is the one synced above.

PR #469, opened 2026-08-13
fix(gateway): enforce one live bridge per session id

robertcsakany:
## Why

Two live bridges could register the **same** `sessionId`. The gateway's connection map was last-writer-wins, so a newcomer silently displaced the incumbent and every server→extension message — prompts included — was delivered to the loser of that race.

Observed on the live instance: a duplicate keeper resumed the same `.jsonl` 14 minutes later, registered second, and owned the slot. Three `POST /api/session/:id/prompt` calls returned `{"success":true}` and were written into the idle duplicate; the real session's transcript never grew. Every diagnostic said "healthy" — `activeBridgeCount`, fresh `processMetrics`, `status: "active"`, live `lsof`/`ps` — because telemetry is *outbound* and the displaced socket keeps sending it. Only transcript mtime disproved it.

Root cause: `pi-gateway.ts` claimed the routing entry unconditionally in the **first-message identity block**, which runs *before* the `session_register` dispatch. A contention check placed in the register branch would read `connections.get(S) === ws` — the newcomer itself — and see no contention.

## What changed

- **D0 — one contention-checked claim point.** The decision moves to the real claim point and is hoisted above every register side effect (watchdog clear, placeholder cleanup, `resetHeartbeat`, callbacks, `onEvent`). Messages from a socket that does not own an id are dropped. Messages are serialized per socket, since the decision is async.
- **D1 — bounded two-factor probe (5 s).** pong → incumbent keeps; no pong but TCP still writable → *busy, not dead* → incumbent keeps; neither → terminate and accept the newcomer. Uses the ping reaper's own liveness rule, so a busy-but-live bridge is never sacrificed. Same-pid reconnect and placeholder incumbents skip the probe.
- **D2 — refusal is terminal.** New `register_rejected` message sent *before* the close; the bridge stops retrying for that id and surfaces the reason. The refused register leaves the spawn-register watchdog armed so the duplicate's pi is reclaimed by its server-minted token. `armSpawnWatchdog` now covers **every** spawn entry point (REST resume — the incident's mint point — WS resume, zombie reopen, headless reload, cold-start recovery, plugin/automation, bridge-initiated); the browser transport is optional.
- **D3 — identity-scoped cleanup.** Every id-keyed teardown is guarded on `connections.get(id) === ws`, and `stop()` terminates `wss.clients` rather than `connections.values()` — the half that let the incident survive two restarts.
- **D4 — prompt honesty.** `POST /api/session/:id/prompt` annotates rather than fails while a contention record is live, and reports `delivered: true` so callers don't retry and double-send.
- **D5 — session-file resume guard.** Both guard sites refuse a `continue` whose target `sessionFile` a live bridge already serves under **any** id — the actual mint point of the duplicate. Fork exempt.
- **D6 — observability.** `/api/health` exposes `bridgeContentionCount`, `contendedSessionIds` and `piGatewayPort`. The refusal log line names both pids and is deliberately distinguishable from the ordinary registration line; log + health entry rate-limited to 1 per id per 5 s.

## Verification

- **~90 new tests** across L1 (vitest), L2 (`qa/tests/17-bridge-contention.{sh,ps1}`) and L3 (`tests/e2e/bridge-contention-health.spec.ts`), covering test-plan E1–E19, X1–X18, F1–F8, L1a–L3a, P1–P3.
- **Revert-check (task 9.2):** restoring the unconditional `connections.set(...)` turns **9** tests red — the suite has teeth, it is not vacuous.
- **Docker harness (production mode):** the L2 scenario passed end-to-end — duplicate received `register_rejected`, was closed, incumbent retained — and the real `server.log` emitted:
  `[gateway] contention refused: qa-contention-65334 incumbentPid=111111 newcomerPid=222222`
  which is exactly the signal that was missing during the incident. All four F6 health-lifecycle assertions verified against the live container.
- Full suite: **13,415 passing**, with the failure set identical to the pre-change baseline.
- Two rounds of isolated `@review` found and fixed a blocking gap (unarmed REST-resume spawn) plus a mid-probe dead-newcomer claim hole.

## Notes

- P3's 5-minute soak is opt-in (`RUN_CONTENTION_SOAK=1`); its invariant is also asserted directly on every run.
- Three main specs carried a delta header (`## ADDED Requirements`) that made their requirements invisible to `archive`/`validate`; repaired only those three. The same pre-existing defect exists in ~60 other specs and is left for a dedicated cleanup.
- **Accepted residual (per design D1):** a truly half-open incumbent reads as "busy" and strands the id until the OS TCP timeout. Deliberate — the alternative rule sacrifices busy-but-live sessions. TCP keepalive is the named follow-up.


robertcsakany:
Addressed all 9 review threads in 6447397ed.

**Major**
- **Concurrent claim race** (`pi-gateway.ts`) — confirmed and fixed. Two newcomers with independent message queues could probe the same incumbent; the first displaced it and took the entry, and the second saw only that the holder had changed and accepted, overwriting the winner. That is last-writer-wins reintroduced, which is exactly what this change exists to prevent. `claim()` now **re-decides against the current holder** rather than accepting on a stale probe, bounded to 3 attempts — each retry requires an *observed* holder change, so it cannot spin against a stable incumbent. Added an E19 concurrent-newcomer test asserting at most one socket is ever routable for the id.
- **PowerShell `require('ws')`** — confirmed. `GetTempFileName()` put the probe in the system temp dir, so the bare specifier never resolved and the test hard-failed on every Windows run without exercising contention. The probe is now written inside the repo and executed from there; the `.sh` twin also `cd`s to the repo root so it no longer depends on the caller's cwd.

**Minor** — all applied: the self-contradicting ownership-gate paragraph and the inaccurate half-open claim in `docs/architecture.md` (a true half-open incumbent *does* block a resume — that is D1's accepted residual, now stated as such), the third guard site documented, `17-bridge-contention.sh` wired into `run-all.sh`, X14 made deterministic (named holder, polls the server, asserts the positive outcome instead of the absence of a string), `tasks.md` 9.3/9.4 relabelled **DEFERRED — NOT VERIFIED**, `tests/e2e/AGENTS.md` re-sorted path-alphabetically, and the E2E spec switched off the phantom `ws` dependency to the Node global `WebSocket`.

Full suite: 13,415 passing, failure set identical to the pre-change baseline.

robertcsakany:
Two of the three new threads fixed in 169b288ac; one disputed.

**Fixed**
- `tests/e2e/AGENTS.md:93` (MD038) — good catch, and the root cause is worse than spacing: the row was truncated *inside* a code span, so markdown parsed ``echo… → see `` as a single span with a trailing space. Closed the span before the ellipsis.
- `bridge-contention-health.spec.ts` timer leak — correct, each connection left a live 5 s timer. Cleared on both settle paths, listeners registered `{ once: true }`.

**Disputed — `docs/architecture.md.AGENTS.md:3` ("split into one fact per line")**

The one-fact-per-line caveman rule applies to `docs/` **prose**, which is where I applied it — see the new `### Bridge Connection Contention` section in `docs/architecture.md`, written as short declarative fragments.

`architecture.md.AGENTS.md` is not prose. It is a pull-only **sidecar index**, and the whole file is three lines:

```
1: # architecture.md — index
2:
3: Full architecture reference. 3 components: … <every section, semicolon-delimited>
```

Line 3 is a single dense index paragraph covering *every* section of `architecture.md` — Server-Keypair Device Pairing, Knowledge Base (KB), and ~40 others — all in the same semicolon-delimited form, and that format predates this change. Reformatting only my contention fragment into one-fact-per-line would make this file internally inconsistent without making it more useful, since its purpose is a single greppable blob the `kb` tooling pulls on demand.

`node scripts/dox-byte-gate.mjs` and `kb dox lint` both pass on the file as written. If the sidecar format should change, that is a repo-wide reformat of this index, not something to do to one fragment inside an unrelated PR.

PR #470, opened 2026-08-13
feat: stateless MCP endpoint at POST /mcp (2026-07-28)

robertcsakany:
Implements `add-dashboard-mcp-server`: a stateless MCP endpoint at `POST /mcp` on protocol revision `2026-07-28`.

## What

New headless plugin `packages/mcp-server-plugin/` (no client entry, `claims: []`), mounting `POST /mcp` on the shared Fastify instance.

- **Protocol** — `2026-07-28` only. No `initialize` handshake, no `Mcp-Session-Id`, no `Last-Event-ID` resumption. `MCP-Protocol-Version` required per request and must agree with `params._meta`.
- **Auth** — every request needs a bearer credential, **including loopback**. `createNetworkGuard` is per-route, so `/mcp` sits outside it and self-guards.
- **Tokens** — per-session opaque 256-bit tokens, SHA-256 at rest, in-memory only (all die on restart), minted over the session's own bridge socket.
- **Self-target guard** — refuses a session driving itself. Direct self-targeting only; the indirect loop is a documented scope limit.
- **Tools** — curated allowlist, 5 of 19 `ServerPluginContext` members. `abort` maps to `abortSession` (soft-only, reports `false` honestly on a disconnected bridge).
- **Streaming** — `subscriptions/listen` over a hijacked reply. `params.sessionIds[]` required; there is no subscribe-to-all.
- **Provisioning** — writes `~/.pi/agent/mcp.json` key `pi-dashboard`, merge-only + atomic + refuse-unparseable.

## Decisions resolved

The change arrived with an unresolved decision gate: 7 test-plan clarifications (C1–C7) plus a recorded token contradiction and the force-kill question. All were resolved and ratified before any code, and recorded in `design.md` Decisions 4a and 6–13.

## Two artifact corrections

- Plugin-route precedent is **seven**, not the "nine → eight" the task directed (`subagents-plugin` only adds a hook; the original prose also named eight while claiming nine). Enumeration command recorded in `design.md`.
- `ServerPluginContext` is **19** members, split **5 allowlisted / 14 denied**.

## Seam change (additive)

`RegisterPiHandlerFn` widened to `(msg, sessionId)`. Implementation proved Decision 6 was otherwise unimplementable: the seam passed only `msg`, so a plugin could learn the caller's session only from the message body — the spoofable source the guard exists to refuse. `(msg)`-only handlers remain valid.

## Security fix found in review

`plugin_pi_message.sessionId` is a **required** protocol field, and `pi-gateway.ts` preferred it over the connection. Session A could send `{sessionId:"B"}`, receive B's token, then self-target A while the guard saw caller B — **the guard was theatre**. `plugin_pi_message` is now attributed to `currentSessionId`; other message types keep prior precedence.

The test asserting this previously exercised a hand-written stand-in that ignored the body by construction, so it passed regardless. Replaced with an end-to-end test over real sockets, verified to fail against the old code.

Review also caught device-token auth being dead (`host.verifyDeviceToken` consumed but never registered) and `subscriptions/listen` being advertised while unwired. Both fixed.

**Guarantee stated precisely:** the attributed id is *"the session this connection registered as"* — not spoofable per-message, but **not** a claim about pi-gateway port authentication, since `currentSessionId` comes from the first `register` message. That pre-existing bridge trust model is out of scope.

## Tests

271 tests in the new package, plus an end-to-end gateway attribution test, real-filesystem provisioning tests, and L2 scenarios in `qa/tests/01-install.sh` and `02-server-start.sh`.

Fail-closed verified by reverting each fix and confirming the test goes red: the gateway attribution fix, `reply.hijack()`/`flushHeaders()`, and the stream teardown binding.

## Not done (deferred, manual-only per the test plan)

- Interop with a real external MCP client (Claude Desktop / Cursor)
- Local pi round-trip with adapter ≥ 2.20.0
- Command-palette duplicate check


robertcsakany:
Thanks — addressed most of these. Two responses.

**Disputed: "Keep OpenSpec artifacts in the active change directory"**

The artifacts are in `openspec/changes/archive/2026-08-13-add-dashboard-mcp-server/` because this PR *completes* the change. Archiving is the prescribed final step here, not a misplacement: `openspec archive` moves the change dir and syncs the delta into `openspec/specs/`, which is why this PR also adds `openspec/specs/dashboard-mcp-server/spec.md` (10 requirements). The repo's convention forbids `openspec/changes/active/`, which is a different path and is not used here. Leaving the change un-archived would be the actual violation.

**CodeQL `js/missing-rate-limiting` — will not clear, and I think that's acceptable**

The finding was valid and I acted on it: `packages/mcp-server-plugin/src/server/rate-limit.ts` adds `AuthFailureThrottle`, wired ahead of the credential comparison in `routes.ts`, returning `429` + `Retry-After`.

The alert persists because the query only recognises known middleware (`@fastify/rate-limit`, `express-rate-limit`) and cannot see a hand-rolled limiter. I chose not to add the dependency because the throttle is deliberately narrower than a general rate limiter:

- Only **failed** authentications count, and a success clears the counter — so a legitimate MCP session driving a fleet is never throttled by its own traffic. A blanket request-rate limit would risk exactly that self-inflicted outage.
- The tracking map is capped at 10k sources; an unbounded per-IP map would trade CPU exhaustion for memory exhaustion, where the attacker picks the key space.

It is defence in depth — the primary control remains the token's 256-bit entropy. Happy to swap in `@fastify/rate-limit` if the team prefers a recognised dependency over a targeted one.

**Fixed:** the 18-vs-19 member contradiction in the risk table; the legacy-protocol risk row (resolved by Decision 10); the `HEAD` self-contradiction in `docs/architecture.md` (only the derived-from-GET behaviour is real); the caveman-style rewrite of the MCP section; the `http` fence identifier.

**Tightened beyond the report:** the slow-consumer spec scenario permitted *dropping*; the implementation terminates the subscription. I tightened the spec to match the code, since a drop policy would silently lose subscribed events.

PR #479, opened 2026-08-13
fix(gateway): gate plugin_pi_message on the connection, not the body

robertcsakany:
## What

Fixes the red `packages/server/src/__tests__/plugin-pi-message-attribution.test.ts` on `develop` (2 failures).

## Root cause

`add-dashboard-mcp-server` correctly made `plugin_pi_message` **attribution** use the socket's registered session:

```ts
const eventSessionId = msg.type === "plugin_pi_message" ? currentSessionId : (msg as any).sessionId;
```

…but the **ownership gate** ~140 lines above it still trusted the body field:

```ts
const named = "sessionId" in msg ? (msg as any).sessionId : undefined;
if (named && connections.get(named) !== ws) return;   // ← drops it here
```

`sessionId` is a required field on the envelope, so a message naming any other session was dropped at the gate and **never reached the attribution logic**. Two mechanisms, opposite assumptions about the same field — which is why the test failed with an empty `seen` (`expected undefined to be defined`) rather than a wrong id.

That test's own docstring anticipated exactly this: it exists because an earlier version "passed while the real gateway preferred `msg.sessionId`… making the whole guard spoofable."

## Fix

Gate `plugin_pi_message` on `currentSessionId`, so the gate and the attribution agree and the body field is untrusted end to end.

## Severity

**Not a security hole.** The old behaviour failed *closed* — spoofed envelopes were dropped, never attributed to the victim, so no credential could be minted for another session. This restores the *specified delivery guarantee* (`dashboard-mcp-server` spec: "A body-supplied session id does not influence attribution") and turns CI green.

## Verification

- `plugin-pi-message-attribution.test.ts`: red on `e389cb794` → **3/3 green**
- Full server suite: **404 files / 4192 tests green** (two independent runs)
- `tsc --noEmit`: clean

> Note: one intermediate full-suite run showed 3 unrelated timing failures (`attach-proposal-replay`, `cli-signal-forwarding`, `event-wiring-resume-clear`) under machine load; all three pass in isolation with the patch and the repeat full run was 100% green.


robertcsakany:
Superseded by 0c1eafae9 on develop — closing.

Both changes fix the same defect (the #469 ownership gate keying `plugin_pi_message` off the untrusted body `sessionId`, so a forged frame was DROPPED rather than re-attributed via `currentSessionId`, and never reached `onEvent`).

`0c1eafae9` is the better of the two and should be the one that lands. Mine folded `plugin_pi_message` into the existing `named` expression:

```ts
const named = msg.type === "plugin_pi_message" ? currentSessionId : ...
if (named && connections.get(named) !== ws) return;
```

That leaves a gap: when `currentSessionId` is `undefined` — an unregistered socket — `named` is falsy, the gate is skipped entirely, and the frame proceeds. `0c1eafae9` takes an explicit branch that requires `currentSessionId && connections.get(currentSessionId) === ws`, so an unregistered socket is refused instead. Strictly tighter for the same test outcome.

No unresolved review threads here and nothing unique to salvage. Closing in favour of develop's fix.

PR #480, opened 2026-08-13
fix(worktree-server-autostart-leak): stop worktree sessions hijacking the shared dashboard ports

robertcsakany:
Implements OpenSpec change `fix-worktree-server-autostart-leak`.

A pi session running inside a git worktree resolves its dashboard server CLI from that worktree and auto-starts it on the **shared default ports**, hijacking the host dashboard's pi-gateway. When that server's startup then failed or hung after `piGateway.start()`, it stayed resident holding the gateway port forever — captured in the wild as PID 78379 (gateway port held 5h52m, dashboard port never bound, live event loop, clean exit on SIGTERM).

## What changed

**1. `server-launch` — bounded startup + teardown** (`packages/server/src/lifecycle/bounded-startup.ts`)
`start()` is now a wrapper around `_startCore()` run through `runBoundedStartup`. A step that throws **or never settles** tears down the listeners — gateway first, so its ping interval stops holding the event loop — and rethrows the **original** error; a teardown error never masks it. Recovery-path exit code `2` is reclassified as a port conflict (`isPortConflictExitCode`).

**2. `bridge-auto-start-lifecycle` — single-flight lock** (`packages/extension/src/autostart-lock.ts`)
Advisory lockfile at `~/.pi/dashboard/autostart-<port>.lock`, atomic `open(..., 'wx')`. Staleness considers the **detached child** (not just the session pid), **pid reuse** (recorded `startedAt` vs process start time) and the readiness budget. Released in a `finally` covering ready/failed/timed-out spawns. An unwritable directory **degrades** to the old behaviour rather than blocking auto-start.

**3. `bridge-auto-start-lifecycle` — worktree refusal** (`packages/extension/src/autostart-guard.ts`)
Refuses when the resolved **cliPath** (never cwd) has a `.worktrees` segment AND either shared default port is in play. Evaluated **before** lock acquisition and **before** `onLaunchStart`, so a refusing session neither contends for the lock nor leaks a spinner. Refusal and lock-loss are appended to the dashboard server log — the refusal path creates the file itself, since it skips the launch primitive that normally does.

## Clarifications resolved

| id | decision |
|----|----------|
| C1 | spawn readiness budget = `HEALTH_CHECK_TIMEOUT_MS × 3` (30s), one constant |
| C2 | no normal-path conflict exit code — probe-based `PortConflictError` stays the sole normal-path mechanism |
| C3 | worktree predicate matches **both** the pre-realpath and post-realpath spelling |
| C4 | server startup deadline derived from the C1 constant, so the two cannot drift |

## Tests

35 new specs across L1 + L2: bounded-startup (E2/E3/E20), gateway timer teardown (E21), `start()` wiring, lock staleness (E5–E9, X6, X7), worktree predicate (E13–E18), auto-start flow (E4, E19, F1, F2, X1–X5, P2), port-conflict classification (E11/E12), plus `qa/tests/18-server-port-hygiene.sh` for E1/E22 — server B takes its **own** gateway port plus A's **occupied** dashboard port, so the gateway binds and the *later* `fastify.listen` fails: the shape only the teardown can clear.

## Notes

- Constants live in `shared/config.ts`, **not** a new module — a brand-new shared file is unresolvable from a git worktree and drops the server into recovery mode (caught by `cli-signal-forwarding` during this change).
- Isolated worktree dashboards must now move **both** ports off their defaults; `isolated-ui-verification` already allocates both.
- QA/manual tasks (`test-plan` #M1 + the operator VM re-runs) deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

PR #482, opened 2026-08-13
feat(client): scope replay-cache entries to their server + purge on reset paths

robertcsakany:
OpenSpec change: `purge-replay-cache-on-reset-paths` (archived as `2026-08-13-purge-replay-cache-on-reset-paths`).

## Problem

The client's durable IndexedDB replay cache was keyed by `sessionId` alone. Session ids are **not unique across dashboard servers**, and a server switch repoints the WebSocket in place — same document, same origin, one shared store accumulating entries from every server.

A surviving entry is not inert, it is **authoritative**: `rehydrateSession` returns its `maxSeq` as the subscribe cursor, so the server delta-replays only the tail onto a stale base. Several reset paths (Refresh Chat, server switch, `session_state_reset`) left the durable entry alive.

## Change

- **Entries are server-scoped.** `ReplayCacheEntry` gains `serverKey` (`host:port`). `get(sessionId, serverKey)` misses on mismatch; `put` stamps it. New `deriveServerKey(wsUrl)` normalizes the scheme-default port so `ws://box/ws` and `ws://box:80/ws` cannot key differently.
- **Schema 2 → 3.** v2 entries carry no server identity, are unattributable, and are purged once. The schema check deliberately runs *before* the key check — a pre-v3 entry has no `serverKey`, so key-first ordering would classify it foreign, decline to delete, and strand a zombie.
- **`delete(sessionId)` stays unkeyed** by design: the schema purge, over-cap branch, poisoned-entry catch and `session_state_reset` have no current-server notion.
- **`resetBuffers()`** discards in-memory buffers on switch — explicitly *not* an invalidation, since the previous server's durable entries stay valid for a switch back.
- **`refreshChat` extracted** from `App.tsx` (it was duplicated behind the header and mobile `onRefresh` props) so the ordering guarantee — durable delete **before** in-memory reset — is unit-testable. Delete-first means an interrupted refresh leaves *both* layers untouched rather than a stale base plus a fresh tail.

## Verification

- **L3 harness: 4/4 green** in `tests/e2e/replay-delta-on-reload.spec.ts`. This change was previously blocked here; the blocker was `fix-optimistic-prompt-stuck-sending` (landed on develop as a1d0da794), not the `pi-ai` version skew that was originally suspected — that hypothesis was falsified directly.
- Full suite: 14011 passed. `tsc --noEmit` clean. All four step-4.4 enforcers green (`check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity`).
- Two rounds of isolated `@review`. Verdict: ship.

## Notable fixes surfaced during the ship loop

1. **The L3 spec threw at runtime**: it called `byTestId(page, "refresh-chat")`, passing the raw app testid as the `TESTIDS` *key*. `TESTIDS[key]` was `undefined`, so `getByTestId(undefined)` threw `Cannot read properties of undefined (reading 'unicode')`. Added `TESTIDS.refreshChat`. The spec predates develop's `tests/` typecheck gate, which does catch the old spelling (verified).
2. **Test #E3 was vacuous.** It built its "v2" fixture via `createReplayCache({schemaVersion:2}).put()`, but `put()` stamps `serverKey` unconditionally — so the fixture carried a *matching* key and passed even under the key-first ordering it claimed to rule out. Now writes a genuine pre-v3 record (raw store, no `serverKey`) and asserts via a **raw store read**, because with `serverKey` absent, deletion is unobservable through `get()`. Verified red-under-injected-bug, green as-is.

## Deferred (manifest-sanctioned)

Tasks 7.1 and 7.2 are `manual-only` rows: subjective repaint feel, and a two-dashboard cross-server check that needs harness infra which does not exist.

## Not in scope

Also drops a stale duplicate `openspec/changes/gate-notify-rows-by-level/` plan dir that this branch carried — develop already shipped and archived that change (`2026-08-10-...`, 64/70 tasks done vs this copy's 0/55.
EOF
)


robertcsakany:
Addressed in eaa18fbcb.

**1. `deriveServerKey` uppercase scheme — fixed, good catch.** Confirmed `deriveServerKey("WSS://box/ws")` returned `box:80`: `secure` came from a case-sensitive `startsWith("wss://")` while `URL` normalizes the scheme, so it fell through to the `ws:` default. That is precisely the "one endpoint, two keys" failure this change exists to prevent. Since `ws:`/`wss:` are WHATWG special schemes, `URL` parses them directly — so `secure` now reads from the parsed `u.protocol` and the `http://` rewrite is gone. Regression test added asserting `WSS://` ≡ `wss://` and `WS://` ≡ `ws://`.

**2. Stale v2 record in `replay-cache.ts.AGENTS.md` — fixed.** Correct, the first record still said `SCHEMA_VERSION = 2` with a `serverKey`-less entry shape, contradicting the record below it. Merged into a single v3 description.

**3. e2e #F1 vacuity — fixed, and you were right that it proved nothing.** The spec only *assumed* persistence via a 1.8 s sleep. It now reloads first and asserts a **delta** subscribe (`lastSeq > 0`), which is reachable only from a cache hit, then clears the captured frames and asserts the post-refresh reload is a full replay. Verified 4/4 green against the docker harness.

**4. Composite `(serverKey, sessionId)` record identity — acknowledged, not taken here.** The mechanism is real: the store keyPath is `sessionId`, so a write for a colliding id on server B replaces A's record and switching back full-replays that session.

Declining it in this PR for a specific reason: it is a **hit-rate** cost, not a data-integrity one. The read-side key check (`entry.serverKey !== serverKey → miss`) still governs what may be served, so a replaced or foreign record can never yield the wrong server's history — the property the change is actually for. Making the record identity composite means a keyPath migration plus an index to preserve the deliberately unkeyed invalidation paths (schema purge, over-cap evict, poisoned-entry catch, `session_state_reset`), which is a larger schema change than the one under review and would land unreviewed.

What I did instead: the spec previously overclaimed ("entries ... remain usable if the user switches back, so switching is not penalised with a full replay in each direction"). That is true per distinct session id but false for a colliding one. The requirement now states the limit explicitly and names the composite-identity change as the follow-up, rather than asserting a guarantee the implementation does not provide.

PR #484, opened 2026-08-13
feat(surface-provider-health-in-settings): health pill + verbatim error in Settings → Providers

molnar-botond:
Implements OpenSpec change `surface-provider-health-in-settings`.

## What
- Per-provider health pill in Settings → Providers with four registers: connected (green + model count), auth-error (yellow + HTTP status), unreachable (red), not-tested (neutral).
- Verbatim `error` string on a monospace line beneath the pill on failure.
- Server caches the probe result `{ok,status,error,modelCount,testedAt}` per provider — written on save (`PUT /api/providers` awaits `probeProvider`) and by the Test button (`POST /api/providers/test`). `GET /api/providers` folds in the credential-free cached health with **no probe on read** and **no background poll**.
- Client refetches health after save; edited-but-unsaved rows drop to not-tested (no stale pill).

## Notes
- Repairs the pre-existing structurally-invalid `provider-connection-test` main spec (leaked `## ADDED Requirements` delta header + missing `## Purpose`) so archive/sync could proceed.
- Two isolated code reviews (gpt-5.6-sol + gpt-5.6-terra) run; all four blocking findings fixed and re-verified.
- QA/manual deploy+smoke task deferred to post-merge verification.


molnar-botond:
CodeRabbit triage: spec + SettingsPanel findings addressed in 2831f55e (now outdated). Deferred: injecting a providers-file path into `registerProviderRoutes` for test isolation — the backup/restore-against-real-`providers.json` pattern is the existing repo convention (see `provider-routes-masked-key.test.ts`); changing the route signature is an out-of-scope refactor better handled repo-wide. The two remaining threads are on archived planning docs (proposal/tasks).

PR #501, opened 2026-08-15
fix(chat): replay indicator no longer occludes scroll-to-bottom, and meets SC 1.4.11

robertcsakany:
## Why

The replay-in-flight pill shipped in #458 had two defects a post-ship UX review **measured**, both invisible to the 22 existing assertions (which pin presence and ARIA, never position or contrast).

1. **It hid a control on mobile.** Pill `absolute bottom-4 right-4 z-10`, scroll-to-bottom `absolute bottom-4 left-1/2 -translate-x-1/2 z-10`. At equal `z-10` the pill painted later and **completely occluded** the button — the user could not return to the bottom of the transcript for as long as the replay ran, exactly when new content is arriving.
2. **No perceivable boundary (WCAG 2.1 SC 1.4.11).** `--bg-tertiary` over `--bg-primary` = **1.19:1** dark / **1.14:1** light; `--border-subtle` ~**1.42:1**, against a 3:1 floor. `shadow-lg` contributes nothing over near-black.

A third rode along: `animate-spin` honoured no `prefers-reduced-motion` branch while `index.css` already ships six such blocks.

Text contrast was **not** a defect (7.69:1 / 8.55:1) and is untouched.

## What changed

- **Tail treatment** replaces the corner chip: a `pointer-events-none` `aria-hidden` scrim pinned to the bottom of the transcript, plus the label centred at `bottom-16 z-20` above the scroll controls — which **keep their resting position**, so nothing moves when a replay ends. Clearance is by **layout, not paint order**: label at 64px, button spanning ~16..51px, both centred, so the ~13px gap is identical at every viewport width.
- **New `--border-strong` token** in both theme blocks (`#808080` / `#777777`) → **5.01:1** dark / **4.48:1** light. No existing border token qualifies (`--border-secondary` 1.57:1) and a fill alone cannot reach 3:1 on `#0a0a0a` without becoming a light-grey blob.
- Spinner suppressed under `prefers-reduced-motion`; the label stays rendered, so the status is conveyed without motion.
- Redundant `aria-label` dropped — it duplicated the visible text verbatim. `data-testid` / `role` / `aria-busy` survive byte-for-byte.

## The red-first gate (design D6)

The design flags "a vacuous non-occlusion test" as its own risk, so the new Playwright case was **run against the unfixed indicator first** and confirmed to fail on **box intersection** — not on a missing selector or a timeout, either of which would mean it was not exercising the defect:

```
Error: indicator {"x":170.2,"y":630.5,"width":188.8,"height":26.5}
       intersects scroll-to-bottom {"x":168.9,"y":619.8,"width":37.2,"height":37.2}
```

After the fix the same unmodified test passes.

## Coverage

| Level | What it pins |
|---|---|
| Playwright `G5` | at 375px the label's box does not intersect scroll-to-bottom, and the button is still **clickable** (occlusion, not absence, was the defect) |
| Playwright `F11` | **rewritten** — see below |
| vitest `G1-G4` | scrim is `pointer-events-none` + `aria-hidden`; scrim and label arm/clear together; neither renders beside the skeleton; label carries its position/stacking/token classes; scroll-to-bottom keeps `bottom-4`/`z-10` |
| vitest `themes` | `--border-strong` in **both** theme blocks, with **computed** WCAG luminance clearing 3:1 in each — not a hardcoded restatement of the claim |

No geometry is asserted in vitest: jsdom has no layout engine, so every `getBoundingClientRect()` returns zeros and an overlap assertion there would pass vacuously.

Local `@review` found `G4` vacuous (`if (bottomBtn)` never fired — the button only mounts on a real scroll event). It now stubs the scroll geometry, asserts presence **before** class, and was verified to **fail closed** by moving the button to `bottom-24`.

## Drive-by fix: `F11` was comparing different rows

`F11` compared `lastRowBox()` at "pill appear" vs "pill clear" and failed with x 516→533. Reproduced identically against the **unfixed** build, so it predates this change — fixed here because it lives in this spec. Two independent reasons it was meaningless:

1. **Different elements.** The tail marker rides the terminal batch, which is also what clears the flag — so at the appear sample it did not exist and `lastRowBox` fell through to its `[data-index]` fallback. The x delta measured row alignment, not the pill.
2. **A legitimately moving target.** With the bottom-pin armed every batch moves the rendered rows, which is why `y` had to be left unpinned — giving up the one axis an in-flow insertion would actually shift.

The anchor is now pinned by `data-index` with the bottom-pin released first, so new content lands *below* it and it must not move at all. The assertion is strictly **stronger**: full box equality including `y`, at two instants — while the indicator is up with batches still landing, and across its removal.

## Verification

- `replay-in-flight-pill.spec.ts` — **3 passed** against the docker harness (rebuilt from local source).
- `npm test -- replay-in-flight themes.test` — **49 passed**.
- `check-conventions --base origin/develop`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity` — all clean.
- `origin/develop` merged before the harness ran, so the gate validated the integrated tree.

## Not verified here

Tasks 5.5 (visual clearance sweep at 375/768/1440 in both themes) and 5.6 (the scrim reads as dimmed-but-legible, not hidden) are **manual QA, to be confirmed later**. 5.5's mechanism is covered by `G5` at the tightest width and is viewport-independent by construction; 5.6 is the one subjective call where the design's fallback is variant B in `mockups/ui-plan.md`.

Pre-existing unit failures on `develop` (biome `noFloatingPromises` in vendored `pi-forms-bpmn` bundles, plus load-sensitive integration specs) are untouched and unrelated.

Client-only. No protocol, server, schema, or persisted-state change. Rollback is a plain revert.


robertcsakany:
Addressed the `AGENTS.md` finding in f9f5943 — that row genuinely carried implementation history on an already-oversized cell.

Disputing this one, with evidence.

The checkbox is not standing in as verification evidence: both lines carry an inline `(manual QA — tested later)` marker immediately after the number, which is the repo's documented convention for this exact case. From `.pi/skills/ship-change/SKILL.md`:

> When only QA/manual tasks remain, marks them done (tested later), archives + syncs specs…

`openspec archive` refuses to run a clean sync while tasks are open, so the alternative is either not archiving (which blocks the merge) or archiving with `--yes` over a "30 incomplete tasks" warning — which loses the *reason* they are open. The inline marker keeps that reason attached to the line, and the PR body states the same thing under "Not verified here". So the outstanding work is recorded in three places, not concealed.

The lines as they stand:

```
- [x] 5.5 (manual QA — tested later) Confirm at 375/768/1440 in both themes …
- [x] 5.6 (manual QA — tested later) Manually confirm the scrim's cost …
```

Worth noting 5.5's *mechanism* is not actually unverified — Playwright `G5` asserts zero box intersection at 375px, the tightest width, and the clearance is viewport-independent by construction (both controls are centred, so the separation is purely vertical). What remains manual there is the visual sweep across the other two widths and both themes. 5.6 is a genuinely subjective call, which is why the design keeps variant B in `mockups/ui-plan.md` as the named fallback rather than allowing the gradient to be quietly tuned.

PR #504, opened 2026-08-18
feat(fix-tmux-cwd-command-injection): close command injection via workspace path in tmux spawn

robertcsakany:
Implements OpenSpec change `fix-tmux-cwd-command-injection`.

Closes a command-injection defect on the tmux session-spawn path: the workspace path was interpolated into a double-quoted pane segment (`cd <cwd> && pi`), where `/bin/sh` still expands `$(…)`/backticks/`$VAR`. `buildTmuxCommand` now returns an argv (`string[]`) with `cwd` as a literal `-c <cwd>` element, and `spawnTmux`/`spawnWslTmux` invoke via `buildSafeArgv` + `execFileSync` (`shell:false`). `spawnWslTmux` routes `wsl.exe --exec <tmux argv>` (bypasses cmd.exe + WSL's default shell). `shellEscape` is retained for the pane command (tmux's own pane shell remains by design).


robertcsakany:
## CodeRabbit triage

Addressed in the follow-up commit:
- **TBD spec Purpose** → filled with the session-spawn dispatch scope.
- **WSL-tmux path contract** → added a note scoping the requirement: the argv/no-shell guarantee holds, while Windows→Linux `wslpath` translation is a pre-existing gap tracked separately (disclosed in design.md Risks/non-goals).
- **Fixed 1.5s sleep in the e2e spec** → replaced with a bounded deadline poll.

Disputed / out of scope:
- **"OpenSpec artifacts sit under a forbidden `archive/` parent"** — this is the correct, expected `openspec archive` destination (`openspec/changes/archive/YYYY-MM-DD-<change>/`). The AGENTS.md "never under `active/`/`archive/`" rule governs where changes are *created*, not the post-merge archive location. Not a defect.
- **"Probe the tmux session inside WSL before selecting the command"** (`spawnWslTmux` always passes `false` → always `new-session`) — pre-existing behavior, unchanged by this change (the prior `wsl ${buildTmuxCommand(cwd, false, ...)}` also always emitted `new-session`). This change's scope is the command-injection fix, not WSL session-probe parity; filed as a follow-up.

Minor style nits (caveman re-route, token in doc snippet, fenced-code language tags on an archived design doc, AGENTS row length) — not gating; left as-is.

PR #509, opened 2026-08-18
feat(providers): warn when the anthropic-messages bridge peer is unresolved

robertcsakany:
Implements OpenSpec change `warn-missing-anthropic-messages-peer` (archived to `openspec/changes/archive/2026-08-18-warn-missing-anthropic-messages-peer`).

## What

When the Anthropic OAuth row in Provider Authentication shows **Connected** but the `flows-anthropic-bridge` probe reports the `@pi/anthropic-messages` peer unresolved, render an inline advisory under that row with a one-click **Install peer**.

Client-only — the bridge probe already ships end to end.

## How

- **`useAnthropicPeerProbe`** — derives `peerMissing` STRICTLY from `/api/health.plugins[flows-anthropic-bridge].lastProbe.peers["@pi/anthropic-messages"].ok === false`. Every other shape (no response, non-OK, no row, no `lastProbe`, malformed payload, still loading) fails **open**. Re-reads on mount, a 60s poll while mounted, window focus, and a successful `pi-package-event`.
- **`AnthropicPeerHint`** — `InlineMessage severity=warning` + a single `Install peer` pill via `usePackageOperations`. An `import failed:` reason withholds the install control (the package IS installed) and switches the copy. The post-install "applies next session start" state is an explicit latch, not the auto-clearing queue success.
- **E10 contract test** pins the bridge's `import failed:` reason prefix so a rename fails a test.
- A **test-only vitest alias** stubs the uninstalled optional peers so `bridge/index.ts` transforms under test.

## Tests

- 26 automated scenarios (E1–E9, F1–F9, X1–X7 in `ProviderAuthSection.peer-hint.test.tsx`; E10 in the bridge). All green: 32 client + 21 bridge.
- Enforcers clean (conventions, dox-byte, i18n lint+parity, knip-config).
- Isolated `@review` pass, no blocking findings.

## Deferred

- **F10 / task 4.1** (visual fit under the green Connected marker, dark + light) — `test-plan: manual-only`, verified post-merge.

## Note on CI

The base `develop` currently carries pre-existing failures unrelated to this change (10× `packages/extension` autoStartServer/mDNS, 1× `scripts/skill-frontmatter` budget). `git diff develop -- packages/extension packages/server scripts` is empty for this branch — those files are identical to `develop`.


robertcsakany:
Thanks — dispositions for the 7 review threads (all 🟡 Minor):

**Applied (commit 829fcd7f9):**
- **Published `spec.md` `TBD` purpose** → replaced with the capability's stable purpose. This was the one live-repo artifact affected; good catch.
- **Stale-response guard in `useAnthropicPeerProbe`** → added a generation-id guard so an overlapping slower read (focus/poll/package-event) can't overwrite a newer response. Fail-open semantics preserved.
- **Archived `proposal.md` stale refs** → corrected the peer-hint test filename (`ProviderAuthSection.peer-hint.test.tsx`) and the import description (probe key = `PEER_AM_LEGACY`; install source = `RECOMMENDED_EXTENSIONS`, not both `PEER_AM`/`PEER_AM_LEGACY`).

**Already correct in the pushed tree:**
- **`vitest.config.ts` "static-imports" → "dynamic-imports"** — the comment already reads "dynamic-imports both anthropic-messages peer specifiers" in the shipped commit; no change needed.

**Intentional / disputed:**
- **Task 4.1 marked `[x]`** — this is the project's `ship-change` convention: manual-only test-plan rows are flipped to `[x]` as *validated post-merge*, not left `[ ]`. The PR body and test-plan both keep F10 explicitly deferred. Leaving it `[ ]` would flag the change as incomplete to the archive tooling.
- **`test-plan.md` L2/L3 rationale wording** and **archive-layout analysis** — archived historical artifacts; the L2/L3 note already states every observable is client-side derivation reachable at L1 and an L3 run can't produce an authenticated Claude row, so no new process/multi-OS behaviour is introduced (the install delegates to the existing package queue). Left as the historical record.

PR #512, opened 2026-08-19
feat(worktree): filter the worktree list + session-less cleanup surface

robertcsakany:
## What

Makes the worktree list usable and gives it a cleanup surface.

**Filter (client-side, no protocol change).** Default predicate `isMain || (!detached && inTree)` hides the noise rows; every hidden row is revealable by a chip that states its action (`+ detached 5`), and the header reads `N of M shown` as a **union**, not a sum of chip counts. `inTree` normalises `\` → `/` — without that every Windows row classifies out-of-tree and the default view collapses to the main row alone.

**Shared `WorktreeList`** in two modes. `spawn` keeps the whole-row `<button>` (the spawn dialog's §1 now delegates to it); `manage` uses a non-button container because a checkbox and a `✕` cannot legally nest inside a button.

**New manage surface** (`Dialog size="lg"` from the folder actions menu) with per-row removal, bulk `remove-batch`, and repo-global prune. The menu gate is on the folder being a git repo — deliberately **independent of live sessions**, since the surface exists to clean up worktrees that have none.

## Server

| Endpoint | Change |
|---|---|
| `POST /api/git/worktree/remove` | `deleteBranch?` → `git branch -d` (never `-D`) after a successful removal; new `is_main_worktree` 400 (was `git_failed` → 500) |
| `POST /api/git/worktree/remove-batch` | **new** — per-item results in input order, 50 cap enforced before any git, never aborts on first failure |
| `POST /api/git/worktree/prune` | **new** — repo-global, returns `{ pruned }` |
| `GET /api/git/worktrees` | entries gain `exists` |

Two traps the design called out, both now provably avoided:

- `BranchDeleteCode` is **disjoint** from `RemoveCode` — the generic failure is `delete_failed`, not `git_failed`, and it never reuses `branch_not_merged` (that code makes `CloseWorktreeDialog` auto-tick `--force` and retry). Both unions are *derived from runtime const arrays*, so the disjointness test evaluates a real set intersection instead of erasing at compile time.
- `exists` is tested `=== false`, never falsy. `undefined` means "unknown, treat as present" — a falsy test would mark every row missing and disable every remove control the moment a new client pairs with an older server.

## Testing

- **Unit/component**: 216 tests across the 7 touched suites. Covers the filter truth table, chip-coverage invariant, union count, path-suppression truth table, elision integrity, exists tri-state, tab-order/nesting, and the branch-delete + batch + prune server paths.
- **E2E**: 6 new Playwright scenarios green **twice** against the docker harness (menu gate, session-less removal + list convergence, escalation inheritance, TOCTOU "already gone", repo-global prune, and contrast ≥ 4.5:1 measured in **both** themes).
- **QA**: `qa/tests/22-worktree-separator.ps1` asserts in-tree classification on Windows separators at the process level.
- Two independent review rounds; no blocking findings. Round-1 fixes landed: stale selection pruning, wired batch retry preserving the original `deleteBranch` intent, fetch-error surface, and a reworked containment test that had been passing vacuously.

## Notes

- Row text moves to `--text-primary` / `--text-secondary`: `--text-muted` fails AA on dark and `--text-tertiary` fails on light, so neither may carry row text. This is a deliberate visible restyle.
- Path elision is JS segment-wise, never CSS `direction:rtl` — bidi reordering relocates leading punctuation and renders `.worktrees/x` as `worktrees/x.`.
- Vanished-directory rows route to `prune`, never `remove` (`validateCwd` 400s a nonexistent path before git runs), so they convert to prune candidates rather than dead-ending on a raw 400.

Deferred (manual-only): chip-affordance wording read, 375px two-line row read, and final isolated-environment QA — each has automated coverage of its mechanical half.


robertcsakany:
Thanks — 5 of 8 addressed in f079a83. Disputing 3 with evidence:

**Fixed**
- **git-operations.ts:981 (Critical, shell injection)** — correct and taken. `git branch -d` now runs via `execFileSync("git", ["branch", "-d", branch], …)`. Added a regression test that asserts the argv shape for a branch named `feat/x&echo_pwned`; reverting to the shell form turns it red (verified). Also widened the E8 command spy to watch `execFileSync` — otherwise the "never invokes `git branch`" assertion would have silently gone blind after this change.
- **ManageWorktreesDialog batch recovery** — taken, though for a different reason than stated. `force` *does* bypass the server guard (`if (activeIds.length > 0 && !force)` in both the single and batch handlers), so the retry would not "repeat the same failure" — it would succeed, and rip the worktree out from under live pi sessions without ending them. That is worse. `active_sessions` rows now escalate to `CloseWorktreeDialog`, which owns the awaited shutdown-then-remove flow. `branch_not_merged` added to the force-retryable set as suggested.
- **X5 contract** — test-plan row now matches tasks.md and the E2E: the registration survives, the row converts to a prune candidate.
- **folder-actions-menu spec** — git-repository precondition added, plus an explicit non-Git clause.
- **`LifecycleSuccess.data`** — fixed in 74b5bf3 (explicit `expect(...).toBeDefined()` + narrowing guard, not optional chaining, so a missing payload cannot pass silently). CI lint is green on these files.

**Disputed**

1. **C1/P1 "resolves a batch availability risk without a guard"** — dropping P1 was an explicit, recorded human decision on clarification C1 (option (c): "no budget — batches are rare, blocking is accepted"), not an oversight. The test plan's own rule was that if C1 resolved to "no budget", P1 is *dropped rather than weakened to an exit-0 assertion* — which is what happened. Inventing a threshold now would be exactly the weakened assertion that rule forbids. The synchronous-Git concern is real and is precisely why the 50-item cap exists and is enforced before any git runs; making removal asynchronous is a larger change than this one and belongs in its own proposal.

2. **WorktreeList: intersect the text query with reveal state** — this would break a specified scenario. Test-plan **E4** requires that typing a substring present only in a hidden `{branch: null}` detached row makes *that row render*; under the proposed intersection it would stay hidden until the user also toggled the right chip. A filter box that cannot find a row you know exists is the worse failure mode, and "search overrides the default view" is the conventional behaviour. The override is deliberate and commented at the call site.

3. **Move sidecar records into the directory `AGENTS.md`** — the sidecars *are* the documented convention here, not a deviation. Root `AGENTS.md`: "Flat dirs with many files split file-based: rows >200 chars promote to a pull-only `<File>.AGENTS.md` sidecar; dir row keeps a summary + `→ see`." `packages/server/src/git-worktree/AGENTS.md` already carries the summary rows with `→ see` pointers, and these sidecars pre-date this PR — it only appends to them. Collapsing them back would breach the 30 KB `AGENTS_BYTE_CAP` that motivated the split; `node scripts/dox-byte-gate.mjs` passes as-is.

robertcsakany:
Second pass — the **critical** CodeQL alert is now gone (PR head is down to 1 open alert, a `high` inside the pre-existing `validateCwd`, same rule class already open on `develop`).

**Fixed in 06d8b1d + 67f694d**
- `removeWorktree` runs `git worktree remove` via `execFileSync("git", […])`. This was the actual live vector: `cwd` is caller-supplied and `remove-batch` accepts up to 50 of them per request.
- `pruneWorktrees` → `spawnSync`, which also let me drop the `2>&1` that was the only reason it needed a shell (both streams are now captured and scanned directly).
- `branchOfWorktree`'s porcelain read → argv form.
- Both new paths have regression tests asserting the argv shape for a `&`-bearing branch *and* a `&`-bearing worktree path; reverting either to the shell form turns them red (verified locally, not assumed).

**Scope-disputed → filed as #514**: migrating the *remaining* `execSync`/`tryRun` paths (`addWorktree`, `addWorktreeFromPr`, merge, push, PR, fetch, diff-stat). These are pre-existing and untouched by this PR, they span most of the worktree lifecycle surface, and each needs its own behavioural test pass. Folding that into a feature PR at merge time is exactly the kind of unreviewed blast radius the repo's surgical-change rule exists to prevent. Exposure is bounded by `networkGuard` (localhost-only) on every affected route.

**Already addressed (stale thread)**: the X5 contract — `test-plan.md:73` was updated in f079a83 and now matches `tasks.md` and the E2E assertion (registration survives, row converts to a prune candidate).

**Still disputed**, with the reasoning unchanged from my previous comment: the C1/P1 threshold (an explicit recorded human decision; the plan's own rule was "drop rather than weaken to an exit-0 assertion") and the text-query/reveal intersection (it would break specified scenario **E4**, which requires a query to surface a hidden `{branch: null}` row without first toggling a chip).

PR #522, opened 2026-08-20
fix(spawn): derive correlation TTLs from the timeout that armed the watchdog

robertcsakany:
Implements OpenSpec change `fix-spawn-correlation-ttl-coupling` (archived + specs synced in this PR).

## Why

`spawnRegisterTimeoutMs` is user-configurable (clamp 5s–120s). The two TTLs that decide what happens **after** it fires were hardcoded 60s. Raising the timeout — the intuitive response to a slow machine — therefore **guaranteed** the correlation token was dead by the time the watchdog fired.

```
 t=0 ──────── 60s ──────── 90s ─────────────── 150s ────────▶
  spawn        │            │                    │
               │            │                    └─ recovery window ENDS
               │            └─ watchdog fires → timeout banner
               └─ correlation token DIES
```

Reported symptom: *"banner clears itself, but no card appears until refresh"* — a register landing after 60s finds the correlation dead **before the watchdog has even fired**, so `session_added` ships without `spawnRequestId` and the client's Tier-1 auto-select never runs; a register landing in 90–150s is still inside the recovery window, so the banner clears anyway. Both halves fall out of the constants.

## What changed

| Design | Change |
|---|---|
| **D1** | Every downstream TTL derives from the **same** value that armed that spawn's watchdog: `clamp(timeout) + RECOVERY_GRACE_MS(60s) + ORDERING_MARGIN_MS(5s)`, in one shared module. `armSpawnWatchdog` now **returns** the timeout it armed with, so the three correlation-recording paths (spawn / resume+fork / degrade) and the fork registry share one config read and survive a mid-spawn Settings change. |
| **D1a** | Records which registries are deliberately **not** re-derived — `pendingAttachRegistry` (anti-strand) and `pendingResumeIntentRegistry` (anti-poison) bound the damage of a *failed* spawn; widening them is a regression, not a fix. |
| **D2** | The fire no longer competes for the correlation; the register broadcast consumes it. `spawn_register_recovered` carries no `requestId` by design — `session_added` is what carries it. |
| **D3** | `hidden` reads a plumbed, strict-boolean `dashboardSpawned` instead of `params.source` — the bridge's **pre-decision self-report**, which can never be `"dashboard"` at register time, so a dashboard spawn reporting `hasUI:false` was filtered out of its own sidebar. |
| **D4** | `recentlyFired` keyed by token when present (a second same-cwd fire used to overwrite the first); cwds normalized via `realpath` with raw-string fallback (`/tmp` vs `/private/tmp`); the pi-gateway clear cascade is **tier-aware**, so registering one spawn stops disarming a concurrent same-cwd spawn's watchdog. The pid tier is left unrepaired **deliberately** (see design D4). |
| **D5** | Fire and recovery are logged with the timeout that **actually applied**; `REGISTER_TIMEOUT` carries `spawnToken` and gains a token-joined `REGISTER_RECOVERED` companion — a recovered timeout is finally distinguishable from one that never recovered. |
| **D6** | A dropped inbound bridge message is reported over the socket and recorded server-side; its `console.error` goes to `/dev/null` under the default `capturePiOutput:false`. Live-socket-gated, never buffered, bounded 10/session/60s; the dropped id is **payload, never the routing field**. |
| **D7** | `POST /api/session/:id/prompt` stops asserting a `delivered:true` it cannot know (it only ever proved a socket write, on exactly the contended branch least able to know). It reports `transmitted` plus a `promptId` handle that rides out on `send_prompt` and returns on `prompt_received`. |

## Verification

- **L1/L2:** full suite green — 1385 files / 15117 tests, 0 failures. New specs cover the TTL boundary (`89_999ms` resolves, `155_001ms` evicts), one-read-per-spawn across a live Settings change, recovery keying, cwd normalization incl. a real EACCES path, tier-aware clears, the `hidden` matrix, drop-report bounds, and the prompt-ack registry.
- **L3 (docker harness):** `tests/e2e/spawn-correlation-recovery.spec.ts` — 5/5 green.
- **Live:** a spawn at `spawnRegisterTimeoutMs: 90_000` arms and fires at `90000ms` (not the constructor default); `server.log` carries `[watchdog] FIRE`/`RECOVERED` with cwd/pid/token/tier and `[bridge-drop]`; `spawn-failures.log` carries the token-joined recovery companion.
- **Gates:** `tsc --noEmit`, `lint:e2e`, check-conventions, dox-byte-gate, i18n-lint `--strict`, i18n-parity, knip-config, knip-ratchet (+ `--check-baseline-diff origin/develop`) all pass.
- **Review:** an isolated `@review` subagent reviewed the full diff. No blocking findings. Three fixes landed from it, each pinned by a test verified to fail on revert: the `recentlyFired` evict timer now guards on **entry identity** (two token-less same-cwd fires shared a key, so the first fire's timer truncated the second's recovery window); `pendingPromptAcks` gained a **per-session in-flight cap** (the TTL bounded each entry's lifetime, not the count); `REGISTER_RECOVERED` names the real mechanism.

## Notes for the reviewer

- **Protocol additions are optional** (`promptId`, `inbound_drop_report`), so an older bridge degrades cleanly: no drop record, prompts stay `transmitted`. `delivered: true` disappearing from the contended prompt response is the one deliberate response-shape change — it corrects a false claim; `success` is untouched. One caller was updated with it (`contention-resume-guard-api.test.ts`).
- **F1/X12 are L1-pinned, not re-run live** — the >60s boundary costs a 70s wall-clock wait to re-assert arithmetic that fake timers pin deterministically. Rationale is recorded in the spec file, not silently checked off.
- **Archive pre-renames:** `openspec archive` compares scenario **name sets** and refuses to drop a scenario the delta supersedes, so five main-spec scenario headings were pre-renamed before the sync. Each is a no-op on the final text (the sync replaces the whole requirement block) and all five are itemized in the archive commit message.
- **Out of scope, deliberately:** five inherited dead exports were de-exported (`autostart-guard`, `ngrok`, `tailscale`) to bring `knip-ratchet` back to baseline. That ratchet was **already red on `origin/develop`** before this branch (verified by stashing the entire diff and re-running); this change does not raise any ceiling.


robertcsakany:
Declining the archived-`tasks.md` suppression finding, with reasons:

1. **Archived changes are immutable history in this repo.** `openspec/changes/archive/**` records what the change said when it shipped; `scripts/check-conventions.mjs` skips it for exactly that reason. Editing a shipped checklist to read better would falsify the record.

2. **The normative text already carries the qualifier the finding asks for.** The binding statement lives in `openspec/specs/bridge-message-pump/spec.md`:

   > The contract is therefore that a drop is recorded **when the channel permits** — not that every drop, nor every suppression, is guaranteed to appear. No requirement SHALL promise a summary that the same saturation can discard.

   So no requirement demands an unconditional suppression summary.

3. **The test the checklist refers to asserts the bound, not the summary.** `connection-inbound-drop-report.test.ts` pins "at most 10 reports in the window" as the hard assertion; the `suppressed` count is asserted only on a report that the bound itself permitted to be sent — i.e. exactly the "when the channel permits" case.

The other three findings from this round are fixed in the next push: `armSpawnWatchdog`'s return value is now in the documented signature, `promptId` is threaded through the slash handoff (bridge-owned route and the non-bridge fallback that reaches `pi.sendUserMessage`), and the P1 budget now measures the added `normalizeCwdKey` cost directly instead of diffing against a "baseline" that also called `realpathSync`.

Issue #521, opened 2026-08-20
[Feature Request]: lazy load session history when open long session


Philogag:
I'm using the Pi extension [ranxianglei/billion-context-pi](https://github.com/ranxianglei/billion-context-pi), which keeps a session in a highly intelligent state and available for a long time, eventually creating a very large session.

When I reopen an old session, loading and replay the entire session onto the browser takes a long time. But I don't actually monitor the history continuously. If I need to find the history, I scroll up to view it. 

Therefore, a better approach would be to load only the most recent messages by default, and then gradually load messages to the client in chunks as the user scrolls up.

robertcsakany:
There is already a server-side seam for exactly this, but it is currently dead code.

### `MAX_REPLAY_EVENTS` — `packages/server/src/browser-handlers/subscription-handler.ts:27`

```ts
/** Max events to replay per session subscription (0 = unlimited) */
const MAX_REPLAY_EVENTS = 0;
```

It is applied at all three replay entry points, always as "keep the last N":

| Line | Path |
|---|---|
| `:242` | stale `lastSeq` → full re-replay |
| `:260` | normal warm subscribe (delta from `lastSeq + 1`) |
| `:346` | **cold hydration** — after `loadSessionEvents` parses the JSONL from disk |

All three do `events.slice(events.length - MAX_REPLAY_EVENTS)`. It is a hard-coded constant, not wired to config (unlike `memoryLimits.maxEventsPerSession`), and set to `0`, so it never fires.

### Does it solve this request?

Partly — it is the server half, and only the naive half.

- ✅ It would cut wire payload and client reduce cost for a huge reopened session, which is the complaint here.
- ❌ It is a one-way truncation with no way back. There is no protocol message to request older events, so once truncated the history is simply gone from the client. That is why it is `0`. This request needs progressive backfill on scroll-up: a range-request message plus a client scroll trigger.
- ❌ Truncating the tail drops the chat **head** (first user message / session intent), which this repo deliberately protects elsewhere — see the `preserve-chat-head-on-event-trim` change (why `maxEventsPerSession` went 5000 → 20000).

### What already mitigates the same pain

- `compactEventsForReplay` (`packages/server/src/session/replay-compaction.ts`) — drops superseded `message_update` snapshots, roughly 20k → 1k events (change: `compact-warm-replay-stream`).
[...]

PR #523, opened 2026-08-20
feat(replay): bound session replay with a head/tail window + on-demand gap backfill

robertcsakany:
Implements OpenSpec change `lazy-load-session-history` (issue #521).

Reopening a long session replayed the **entire** stored stream to the browser.
This bounds it, by user configuration, without losing access to what was elided.

## What changed

- **`memoryLimits.maxReplayEvents`** (default `0` = unlimited, requires server restart).
  Positive values below `MIN_REPLAY_WINDOW` (100) clamp up; `0` is never clamped.
- **Head/tail window** on FULL-stream subscribes only. Keyed on **content**
  (`lastSeq === 0 || lastSeq > maxSeq`), never on call site — the `:260` branch is
  dual-purpose and serves warm reloads, the dominant reopen path.
  A genuine delta is never windowed and never emits `history_window`.
- **Applied AFTER compaction** (~20:1), so the budget buys conversation rather than
  superseded snapshots. The returned high-water seq still comes from the
  pre-compaction **full input**, so `clearReplaying` can never re-send.
- **Both cut edges snap** (tail forward, head backward, bounded by `SNAP_LOOKUP`), and
  both **shrink** — the configured budget stays a hard cap, never a soft floor.
- **Three additive protocol messages**: `history_window` discloses the gap,
  `history_backfill` requests an explicit range, `history_backfill_result` answers —
  exactly once per request, on every path including refusals.
- **`EventStore.getEventsRange`**: binary search on both bounds + one slice,
  O(log n + k), not a linear filter.
- **Click-to-load gap divider** in chat. States: idle (exact count) / loading /
  refused / unavailable / removed once filled. Every protocol error code collapses to
  one plain sentence — codes never reach the user.
- A windowed replay is **not** written to the client replay cache, so the gap can
  never become permanently invisible behind a cache hit.

## Safety

`maxReplayEvents` defaults to `0` and the feature is fully inert at that value —
`computeReplayWindow` returns `null`, and `compactEventsForReplay(arr)` is
byte-identical to its previous behaviour (pinned by a regression test).

## Verification

- 15,126 unit tests pass, including 56 new scenarios from `test-plan.md`.
- 2 Playwright specs green against the docker harness.
- All ship enforcers green, incl. `knip-ratchet --check-baseline-diff origin/develop`.
- Independent `@review` pass: no blocking findings. One real bug it surfaced (a stale
  scroll anchor when a backfill splices nothing) is fixed; four non-blocking residuals
  are recorded in the archived `design.md`.

## Deviations from the plan

- Test-plan rows **F4–F11** were catalogued L3 but are covered at jsdom level. Windowing
  only fires when the server runs with a non-zero `maxReplayEvents` — a restart-only
  field on the docker harness every other e2e spec shares — and the new messages arrive
  over the WebSocket, which `page.route()` cannot intercept. F12/F13 remain L3.
  Recorded in `tasks.md` § 16.
- QA/manual task 17.1 (divider copy at the singular/plural boundary) deferred to
  post-merge verification; its copy is asserted at unit level.


robertcsakany:
Addressed in bbbd71817 — thanks, three of these were real.

**Fixed**

- **Non-head-adjacent request orphans part of the gap** (`subscription-handler.ts`) — correct, and the server should not have trusted the client to only ever walk forward from the edge. `headMaxSeq` now advances *only* when the served range is genuinely head-adjacent (`from === headMaxSeq + 1`); `remainingGapCount` is computed from the (possibly unmoved) head edge, so it stays truthful either way. Covered by a new test that asserts an out-of-order request is served but leaves the head and the count untouched.
- **Restore the anchor only for the matching backfill splice** (`ChatView.tsx`) — correct on both interleavings, including the one I had missed: the final splice inserts rows *and* removes the divider, so the net row count is unchanged and a length-keyed effect never fires. Replaced the `messages.length` heuristic with a monotonic splice revision bumped once per successful splice in `useMessageHandler`. Three new tests cover: bump on the final zero-net splice, no bump on a live event, no bump on a refusal/empty response.
- **Restore the shared configuration after the test** (`max-replay-events-setting.spec.ts`) — correct, and this one would have been a genuine cross-spec flake source: a leftover non-zero `maxReplayEvents` would window *other* specs' replays. The original `memoryLimits` are now restored in a `finally`.
- **Conflicting zero-gap `history_window` contract** — real contradiction, and the spec text was the wrong half. Resolved in favour of the implementation: no window applied ⇒ **no message at all**, never a `gapCount: 0` announcement. Same rationale as excluding deltas — a zero-gap announcement would silently reset the bookkeeping of a client that is mid-gap-browsing. Fixed in both the synced and archived spec, and documented on the wire type.

**Disputed, with evidence**

- **Backfill stops at trimmed holes** and **compaction context across chunk boundaries** — both are accurate observations, but they are challenges to *documented decisions* (D6's response-keyed stop rule and D7's per-slice boundary), not implementation defects: the code matches the design, and an independent review pass reached the same two conclusions. Rather than headlessly rewriting the design mid-ship, both are recorded verbatim as known residuals in the archived `design.md` § *Residuals found by the independent review* — which is the file you anchored the comments on. The blast radius is bounded to zero for existing users: `maxReplayEvents` defaults to `0` and the feature is fully inert at that value. The hole-skipping fix is explicitly noted as a prerequisite for any non-zero default, along with the observation that `oldestGapSeq` is already on the wire as its seed.
- **Invalidate old replay work on unsubscribe** — I implemented a subscription-generation guard for this and then removed it, because it is **unfalsifiable in this code path**: `sendEventBatches` computes and emits `history_window` *synchronously* before its first `await`, so the generation cannot change between the call and the emission, and no test can drive the interleaving. Shipping a guard that can never fire, behind a test that can only pass vacuously, is worse than not shipping it. The announcement is computed from the very events being delivered on that call, so it is never stale relative to them, and a client that has since re-subscribed drops its gap state on the `session_state_reset` that precedes every windowed full-stream replay. The gap *bookkeeping* is already invalidated on unsubscribe via `clearGapState`. Happy to revisit if you can point at an interleaving I have missed.

robertcsakany:
Round 2 — one fix, one false positive. This is the second and final review round for this change (the ship loop caps at review → fix → re-review).

**Fixed**

- **Clear history-gap state when the server changes** (`App.tsx`) — correct. Gap bookkeeping is scoped to one server's replay window, and server B may hold the same `sessionId` and replay it unwindowed, so a surviving `tailMinSeq`/`gapCount` would splice a false divider into a transcript that has no gap. `setHistoryGaps(new Map())` and `setHistorySpliceRev(0)` now sit alongside the seven sibling map clears in `clearInMemoryState`. Untested by construction, exactly as those seven siblings are — `server-switch.test.ts` mocks `clearInMemoryState` wholesale, so the closure's contents are not reachable from it.

**False positive**

- **Move OpenSpec artifacts out of `archive/`** — this repo's `AGENTS.md` documents this exact review outcome as a known false positive:

  > `ship-change` MOVES a completed change into `openspec/changes/archive/<date>-<name>/`, which `scripts/check-conventions.mjs` skips as immutable history; **a review asking to move an archived change back is a false positive.**

  The archive move is the intended terminal state of the ship pipeline, performed by `openspec archive` (which also synced the five delta specs into `openspec/specs/`). `openspec validate --specs` passes 549/549 and `check-conventions --base origin/develop` reports no violations.

**Still standing from round 1** (unchanged, answered in [the previous comment](https://github.com/BlackBeltTechnology/pi-agent-dashboard/pull/523#issuecomment-5363333420)): the two `design.md` threads are recorded design residuals rather than implementation defects — bounded to zero blast radius by the default `maxReplayEvents: 0` — and the unsubscribe-invalidation guard is unfalsifiable in this code path because the announcement is emitted synchronously before the first `await`.

PR #524, opened 2026-08-20
feat(headless-reload): one honest reload entry point; respawn stays the default

robertcsakany:
Implements OpenSpec change `fix-out-of-band-reload`.

## What was broken

Reload has six triggers. Four of them fanned out with a raw
`piGateway.sendToSession(sid, {type:"send_prompt", text:"/reload"})`, which goes straight to the
bridge and never passes the server-side interception. In the bridge, `/reload` calls
`options.reload()`, which reads `globalThis[RELOAD_KEY]` — populated only when a human once typed
`/__dashboard_reload` in pi's TUI. For every dashboard-spawned session that key is absent, so the
reload logged one line to stderr and did nothing.

Then `command-handler.ts` emitted `command_feedback {status:"completed"}` **unconditionally**.
An operator saved settings, installed a package or updated pi-core, saw success, and nothing had
changed. That false success was the core defect.

Two compounding gaps: the fan-outs iterated `getConnectedSessionIds()` only, so a headless session
whose bridge WS died — alive and respawnable — was never targeted; and nothing checked whether a
session was mid-run before delivering a reload.

## What this does

One entry point, `dispatchReload(sessionId)`:

1. **Busy** (compacting, or streaming with a live bridge) → refuse, mirroring pi's own TUI wording.
2. **Headless PID** → kill-and-respawn.
3. **No PID, live bridge** → forward `/reload` (terminal-hosted), gated on `sendToSession`'s
   return value, not the connection probe.
4. **Neither** → an honest terminal error.

Plus: exactly one truthful terminal `command_feedback` per reload, always keyed `/reload`
(`BridgeCommandOptions.reload` now returns a `ReloadOutcome`, including the synchronous-throw case
of a single-use captured `RELOAD_KEY`); fan-outs target connected ∪ registry; a
`DashboardSession.compacting` signal derived from events the bridge already forwards; a
stale-`streaming` carve-out so a bridge-dead session stays respawnable; an explicit
never-respawn-a-PID-less-session guard; and pi-core update split out as `respawnForRuntimeSwap`
(a binary swap cannot be deferred by the busy check).

## The measurement worth reading

An earlier revision of this change dispatched `/__dashboard_reload` to the session's RPC keeper,
on the claim that pi's RPC mode runs the line through `session.prompt()` **with** command
handling. The docker harness (with `keeperLog.capturePiOutput = true`) falsified it:

```json
{"type":"message_start","message":{"role":"user","content":[{"type":"text","text":"/__dashboard_reload"}]}}
{"type":"message_update","assistantMessageEvent":{"type":"text_end","content":"faux: no scenario (id=unset)"}}
{"type":"agent_end", ...}
```

pi sent the literal text **to the model**. Control: pi's built-in `/help` written to the same
socket behaved identically — not the `__` prefix, not our registration. **pi's RPC
`{type:"prompt"}` performs no slash-command dispatch.** That path was removed; it would have
injected junk into the operator's transcript, burned a model round-trip, and reported `completed`
because the socket write succeeded.

> **Follow-up worth filing:** `rpc-keeper/dispatch-router.ts` (`dispatch_extension_command`, from
> `add-rpc-stdin-dispatch-with-keeper-sidecar`) uses the same `writeRpc` + `{type:"prompt"}`
> mechanism and therefore has the same defect for every slash command it forwards. Live on
> `develop`, out of scope here. `docs/architecture.md` § RPC keeper sidecar still asserts it
> dispatches slash commands.

## Verification

- 22 L1 scenarios (E1–E12, X1–X8, plus X9/X10/P1).
- L3: `tests/e2e/headless-reload-dispatch.spec.ts` passes against the docker harness — one
  terminal pill, a new pid after the respawn, status converging off `ended`, tokens surviving the
  re-register, and the session answering a follow-up prompt.
- `npm test` + `npm run build` green (3 parallel-run flakes triaged, all pass in isolation, none in
  this change's area).
- Enforcers pass: `check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `knip-config`,
  `knip-ratchet --check-baseline-diff`.

## Deliberately left open

`9.4`/`9.5` (F4 streaming-refusal e2e, F5 toast coalescing e2e) and `12.3`/`12.4` (the two
follow-up changes above) are tracked **unchecked** in the archived `tasks.md` rather than marked
done. QA/manual rows `10.1`/`10.2` deferred to post-merge.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

robertcsakany:
Addressed. Summary of triage — 9 of 10 applied, 1 disputed with a citation.

**Functional (applied):**
- **Await the reload result** — correct and the most important one. `reload` is now `async` and awaited; a rejection after `completed` was exactly the false success this PR removes. Handler also try/catches the callback, and there is now a test for the rejected-promise path.
- **Concurrent respawn guard** — correct, and it was a process *leak*: both callers saw the old PID, both killed it, both spawned, and the registry only remembers the last registration. Added a per-session in-flight guard. Note it had to be checked **before** the PID gate, not after — the kill half of a respawn clears the PID, so a guard placed after it never fires (the first version of this fix failed its own concurrency test that way).
- **Fan-out target filter** — correct. The filter's only effect was to drop a connected session with status `ended` and no PID, which is still forwardable. Removed entirely.
- **Count only accepted reloads** — correct, applied.
- **Local-scope registry union** — correct, and the same class of bug as the global fan-out. `listSessions()` now carries `cwd` and local scope unions registry entries the scope governs, with the containment rule restated next to it. New test covers it.

**Tests/docs (applied):** mock now throws instead of pre-converting; E8 rewritten as genuinely concurrent plus a guard-clears test; archived test-plan canonical rows updated in place for rev 2 with counts recomputed; e2e AGENTS.md row shortened; the Retry bullet in `docs/architecture.md` split into one-fact-per-line form.

**Disputed — "Move these OpenSpec artifacts out of `archive/`":** false positive. `AGENTS.md` line 133 states the rule is *creation-time only* and names this exact case: "`ship-change` MOVES a completed change into `openspec/changes/archive/<date>-<name>/`, which `scripts/check-conventions.mjs` skips as immutable history; **a review asking to move an archived change back is a false positive.**" `check-conventions --base origin/develop` passes. Sibling archived changes (e.g. `2026-08-20-fix-spawn-correlation-ttl-coupling`) live there too.

**Partially applied — caveman style in `docs/architecture.md`:** the `/reload` Flow section (1121–1178) was written by DocScribe in caveman style already. The line-350 bullet I touched was prose; that one is now split one-fact-per-line.

Server + extension suites: 6168 passed. Enforcers green (`check-conventions`, `knip-ratchet --check-baseline-diff`, `i18n-lint --strict`, `dox-byte-gate`).

PR #525, opened 2026-08-21
fix(folder-header): stop a worktree branch leaking into its parent folder header

robertcsakany:
Implements OpenSpec change `fix-folder-header-worktree-branch-leak`.

## Problem

Creating a git worktree made the **main folder's** sidebar header display the **worktree's** branch (`os/foo` instead of `develop`), and it only corrected itself by accident. Two independent defects combined:

1. **The fallback answered a question it cannot know.** `GroupGitInfo` resolved the folder branch positionally (`sessions.find(s => s.gitBranch)`). `maybeRekeyOrder` moves a worktree session to the **front** of its parent folder's order the moment `git_info_update` establishes its worktree identity, so a fresh worktree session deterministically won that fallback — and supplied the branch URL and PR pill with it. A folder whose only children are worktree sessions had no correct answer available at all.
2. **The authoritative value was missing far more often than the spec assumed.** `git_head_update` broadcasts only on first-seen-or-change and there was no connect-time snapshot, so a browser connecting after the server cached a folder never learned its HEAD — the positional fallback was permanently load-bearing.

## What changed

- **Eligible-child resolution (D1).** A child session supplies the folder header's git identity only when its own `cwd` **is** the folder group's `cwd` under the shared `pathKey` canonicalization. The whole tuple — branch, branch URL, PR number, PR URL — comes from that one session, never mixed. Expressed as *cwd identity* rather than "exclude worktree sessions", which keeps the pinned-worktree-folder case correct.
- **Folder-HEAD snapshot on connect (D2/D3).** The server unicasts its cached folder-HEAD map to a newly connected browser, reusing the existing `git_head_update` message (zero protocol surface, zero client change), behind a `typeof ... === "function"` guard so hand-built `DirectoryService` fakes keep working.
- **Refresh on entry (D4).** `directory-service` owns THE single group-key recompute path plus a debounced `refreshFolderHeadsForEnteringKeys()`, invoked from `maybeRekeyOrder`, from `session_register` **ungated** by `isNewCwd` (that check is false whenever an ended session already carries the cwd, while the poll set skips ended sessions), and from `onDirectoryAdded`.
- **Cache retained on set-leave (D5).** The poll set is not the set of folders the client renders (workspace-only and ended-only folders render without being in it), so eviction would strip the connect snapshot for folders the user is still looking at.

## Bonus: a root-cause bug this surfaced

`readHeadDisplayAsync` used the `platform/exec` wrapper's `execFileAsync`, which lacks node's `util.promisify.custom` symbol — so `promisify` falls back to its generic rule and resolves with the **bare stdout string**. Destructuring `{ stdout }` off a string yields `undefined`, and `String(undefined)` is the literal `"undefined"`.

**Every folder-HEAD read has been caching a branch named `"undefined"`** since `attribute-openspec-poll-eventloop-stalls`. It stayed invisible precisely because no browser ever received the cached value — the connect snapshot is what exposed it. Fixed at the call site with real node `execFile` + `promisify`, mirroring the documented precedent in `lib/path-containment.ts` and `pi/pi-core-checker.ts`. The regression test fails on revert (`Received: "undefined"`).

## Verification

- **L1 (vitest):** 26 new tests — 8 client (`SessionCard.test.tsx`), 4 poll + 1 reader (`folder-head-poll.test.ts`), 8 directory-service (`folder-head-integration.test.ts`), 5 gateway (`browser-gateway-snapshot-on-connect.test.ts`). Covers test-plan #E1-#E21, #X1-#X3.
- **L3 (Playwright, docker harness):** 5 scenarios (#F1-#F5) — green twice consecutively against the all-in-one harness. #F1 was verified non-vacuous: the spawned worktree session really did fold into the parent group (`gitBranch: wtleak-*`, `mainPath: /fixtures/sample-git`).
- **Enforcers:** `check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity`, `knip-config`, `knip-ratchet --check-baseline-diff` all pass.
- **Local `@review` pass:** 0 blocking findings; 2 actionable nits applied.

## Notes

- `knip-ratchet` reports `types 194 > baseline 193 (+1)`. Verified **pre-existing on `develop`**, not introduced here: substituting `origin/develop`'s version of every file this PR touches still yields 194. `--check-baseline-diff origin/develop` passes (no baseline was raised).
- Manual task 6.1 (eyeball the worktree-only header layout after the PR pill and branch link are gone) is deferred to post-merge verification per its `(test-plan: manual-only)` disposition.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

PR #530, opened 2026-08-22
fix(repair-tool-error-surfaces): fold the scenario manifest, fix 3 defects it surfaced

robertcsakany:
## Summary

OpenSpec change `repair-tool-error-surfaces`. The user-visible fix shipped in #465; the change was then **reopened** with a scenario manifest (`test-plan.md`, 24 automated + 3 manual rows). This PR folds those rows — and fixes **three real defects the new tests uncovered**.

## Defects found by the folded scenarios

| # | Defect | Fix |
|---|---|---|
| `#E9` | A line reading `stderr:` **inside stdout's own output** opened a real section, stranding the rest of stdout under a bogus label | A stream header counts only when preceded by a blank line (or start-of-body) and not already opened — how context-mode actually emits the dump |
| `#X3` | Stream sections were **unbounded** — a failing command emitting thousands of lines dumped all of them into the DOM | Reuse `truncateOutputForDisplay` (last 200 lines + a hidden-lines marker), the same helper `BashOutputCard` uses. No second cap introduced |
| `#F8` | `UrlLink`/`FileLink` hardcoded `text-blue-400`, tuned on a dark ground. On `--bg-code` it measured **1.80–2.52:1 across all NINE light themes** — the same one-sided defect this change exists to fix | Use the mode-aware `--link` token (also the 3 hand-written `<a>` in `CtxToolRenderer`) |

After the `#F8` fix only `tokyo-night/light` remains, at 2.77:1 (from 1.80), taking the carve-out design.md D3 already inherits — keyed `tokyo-night/light/link`, deliberately **not** borrowed from the `info` cell so fixing that cell cannot silently re-raise this floor. Every other cell clears 3:1 outright, **where previously nothing gated any of them**.

## Scenarios folded

- **Parser** `#E1`–`#E10`, `#X1` — falsy-zero exit code, non-numeric exit, banner-strip ordering, no-language fence, delimiter injection, multi-fence, malformed-input fuzz
- **Renderer** `#F1`–`#F6`, `#X2`, `#X3` — chrome-vs-content split, three-channel error-signal redundancy, five single-line surfaces, partial-extraction degradation
- **Contrast** `#F7`–`#F9` — 18 theme·mode sweep, links on `--bg-code`, anti-vacuity token-resolution guard
- `AgentToolRenderer` gains its **first** unit test (`#F4`'s agent cell)

`#E11`/`#E12` needed no work: the guard already ships as `scripts/__tests__/severity-literal-guard.test.mjs`. `design.md` Q1 is corrected to record that shipped resolution — it had been left open, and a duplicate rule added to `check-conventions.mjs` during this pass was **reverted** after review caught the divergence.

## Review notes

Two review rounds (`@review`, isolated subagent). Round 1 caught the duplicate guard; round 2 caught **`#F5` passing vacuously** — it asserted on the message `<pre>` because the `receivedArgs` `Collapsible` renders only when open, *and* the error message text also contains the word "language". Rewritten to scope to the collapsible's own subtree, and **proven to fail on revert** by mutating the source.

## Verification

- 6/6 `severity-contrast.spec.ts` e2e rows green against the docker harness
- Touched unit suites green; the full `npm test` red list is pre-existing/environmental only (missing `apple-tools` workspace link, perf smokes, pi version pins) — none in touched files
- `check-conventions`, `dox-byte-gate`, `i18n-lint`, `i18n-parity`, `knip-config`, `knip-ratchet --check-baseline-diff` all pass
- `openspec validate --specs`: 552 passed, 0 failed

Remaining open tasks are the 3 `manual-only` manifest rows (10.1–10.3), which defer legitimately.

## Archive notes

Second archive of this name. Both documented traps hit and worked through (ADDED-already-exists lift; dropped-scenario refusal → 16 pre-existing scenarios carried into the delta). The lift initially **regressed** `message-severity-tokens` by overwriting a later refinement with older delta text; caught by the content-equality proof and restored in both the live spec and the archived delta.


robertcsakany:
Re the `openspec/specs/in-memory-event-buffer/spec.md` depth-limit finding — **not actionable in this PR**, and now marked outdated.

That file belongs to `fix-pasted-image-message-vanishes`, not to this change. It appeared in the diff only because `develop` had two unpushed commits sitting underneath this branch, so the PR was diffing against a stale base. `develop` has since been pushed and the branch re-merged; the PR is now 20 files, all belonging to `repair-tool-error-surfaces`.

The observation itself may well be valid — it just needs to be raised against the change that owns that capability.

All other findings were fixed in 5b5116e2c and ae648b070:
- `#F4` assertions now cover the element's own `className`, not just `innerHTML` (a raw red on the line itself would have slipped through)
- `AgentToolRenderer.test.tsx` gained a completed-agent case and a missing-`toolDetails` case, and now uses the correct `AgentDetails` `"completed"` status (the earlier `"complete"` landed on the stopped fallback, so it asserted without exercising the completed branch)
- `test-plan` C1/F8 and task 5.2 now record the documented `tokyo-night/light` link carve-out instead of claiming >= 3:1 in all 18 cells

Issue #531, opened 2026-08-22
Electron-E2E: zombie-adoption spec hangs on macOS (pre-existing, darwin path never runs in CI)


robertcsakany:
## Summary

`tests/e2e-electron/zombie-adoption.electron.spec.ts` hangs on **macOS**. Every test in the
file exceeds the 90 s test timeout and then the worker teardown times out too, so the app is
never cleanly closed.

## This is NOT an Electron-43 regression

Found while landing `upgrade-electron-runtime` (Electron 32.3.3 → 43.4.1). A control run was
done to attribute it:

| Electron | zombie-adoption on macOS |
|---|---|
| 43.4.1 | hangs (2/2 run tests, worker teardown timeout) |
| **32.3.3** (previous pin) | **hangs identically** |

So the behaviour predates the bump.

## Why it went unnoticed

`.github/workflows/ci-e2e-electron.yml` runs a `[ubuntu-latest, windows-latest]` matrix —
there is **no macOS leg**, so the darwin path of this suite has never executed in CI. The same
blind spot hid a second, simpler bug that `upgrade-electron-runtime` did fix:
`resolvePackagedBinary()` derived the darwin binary name from the `.app` bundle basename
(`PI-Dashboard`) instead of forge's `executableName` (`pi-dashboard`), so it ENOENT'd before
ever launching.

## Suspected cause (not yet confirmed)

`stubDialog()` monkeypatches `dialog.showMessageBox` in the main process *after* launch, racing
the app's health fetch. On macOS a real `dialog.showMessageBox` is a native modal that blocks
the main-process run loop, so if the real modal opens before the stub installs, everything
afterwards — including `app.close()` — blocks forever. Note the `--no-zombie-prompt` test hangs
too, which does not obviously fit that story and should be explained before a fix lands.

## Impact

Scenario F2 ("native surfaces survive 11 majors") is only partly covered on darwin:
`tray-ownership.electron.spec.ts` passes locally on macOS with Electron 43, but the
dialog-driven flows in this spec cannot be verified there.
`doctor-version-skew.electron.spec.ts` is separately unrunnable next to a live dashboard because
the app hard-codes `:8000`.

## Suggested work

- Install the dialog stub before the app can reach the adoption arm (or gate the arm behind a
  test-only switch) rather than racing it post-launch.
- Explain the `--no-zombie-prompt` hang, which the modal-race theory does not cover.
[...]

PR #532, opened 2026-08-22
feat(electron): upgrade runtime 32.3.3 → 43.4.1, raise macOS floor to 12.0

### LINKED ISSUE #529: Electron pinned to 32.3.3 (EOL) — Chromium 128, 11 majors behind, no security backports
## Summary

`packages/electron/package.json:32` pins `electron: 32.3.3`. That release line is **end-of-life** and receives no security fixes. Because the desktop app ships a full Chromium to end users, this is security debt in a released artifact, not just dependency drift.

## Facts

| | |
|---|---|
| Current pin | `electron@32.3.3` → **Chromium 128**, published `2025-03-03` |
| Latest stable | `electron@43.4.1` → **Chromium 150**, published `2026-08-19` |
| Gap | **11 Electron majors / ~22 Chromium majors / 533 days** |
| Electron support policy | latest **3** stable majors — currently 43, 42, 41 |
| Status of 32.x | **unsupported — no security backports** |

Electron's [version support policy](https://www.electronjs.org/docs/latest/tutorial/electron-timelines): *"The latest three stable major versions are supported by the Electron team… The oldest supported release line will receive only security fixes directly."*

Electron tracks even-numbered Chromium releases on an 8-week cadence (E26=Chromium 116, E27=118, …), which puts our pin at Chromium 128.

## Why it matters here

- The Electron shell renders the dashboard via `mainWindow.loadURL(serverUrl)` (`packages/electron/src/main.ts:432`), so any Chromium renderer CVE since 128 applies to a surface that loads server-rendered content.
- The shell can expose CDP on a TCP port (`--debug-cdp` / `PI_DEBUG_CDP`, default `9222` — `packages/electron/src/lib/resolve-cdp-activation.ts`). Opt-in and default OFF, but it raises the value of keeping the engine current.
- Auto-update ships this Chromium to every installed user.

## Good news: the upgrade surface looks small

The Electron API surface in `packages/electron/src/main.ts` is narrow and entirely stable APIs:

```
app.quit / app.on / app.whenReady / app.commandLine / app.isPackaged
app.requestSingleInstanceLock / app.name / app.disableHardwareAcceleration
BrowserWindow  dialog.showMessageBox  shell.openExternal
ipcMain.handle / removeHandler / removeAllListeners / on
```

No `remote` module. `contextIsolation: true` and `nodeIntegration: false` already set (`main.ts:407-409`). No native runtime dependencies requiring an ABI rebuild — `electron-updater` is pure JS, and `sharp` is a devDependency used only for icon generation.

So "11 majors of breaking changes" likely overstates the real work. The risk concentrates in toolchain and packaging rather than app code.

## Risk areas to check

- `@electron-forge/*` (`^7.6.0`) and `electron-builder` (`^26.8.1`) compatibility with the target Electron
- `forge.config.ts`, `electron-builder.yml`, `electron-builder-nsis.json`, `entitlements.plist` — macOS signing/entitlements and Windows NSIS packaging
- Node major bump inside Electron between 32 and target
- Existing Electron e2e/smoke coverage (`playwright.electron.config.ts`) as the regression gate

## Sequencing

Two open OpenSpec changes already touch this package and should be considered for ordering:

- `openspec/changes/electron-platform-extraction`
- `openspec/changes/harden-electron-renderer-boundary`

## Suggested approach

Step to a supported line rather than jumping blind — verify build + sign + notarize + smoke at each stop, then land on the newest of 41/42/43 that packages cleanly.

---

*Filed from an explore-mode research session on browser providers; discovered while costing out a "fork a browser" proposal. No code changed.*

robertcsakany:
Closes #529. Refs #531.

OpenSpec change: `upgrade-electron-runtime`.

## Why

`packages/electron/package.json` pinned `electron@32.3.3`. That line is **end-of-life and receives no security backports**. The desktop app ships a full Chromium to end users and auto-updates it into every install, so this was ~22 Chromium majors of known renderer CVEs in a **released artifact** — security debt, not dependency drift. The renderer loads server-rendered content via `loadURL`, so every renderer CVE since Chromium 128 lands on a live surface.

## The app code is a no-op — the macOS floor is the real work

The Electron API surface was re-derived from the `import … from "electron"` sites (not a `Symbol.method` grep, which misses constructor uses like `new Tray(...)`). It is entirely long-stable APIs. **Nothing under `packages/electron/src/` changed except new tests** — the proposal called a diff there a signal to stop.

Every currently-supported Electron line requires macOS 12, so 10.15 and 11 are dropped and all three enforcement points move together:

| point | before | after |
|---|---|---|
| `forge.config.ts` `LSMinimumSystemVersion` | `10.15` | `12.0` |
| `_electron-build.yml` `MACOSX_DEPLOYMENT_TARGET` | `10.15` | `12.0` |
| CI verify step | plist `10.15`, otool `-gt`, per-arch `10`/`11` | plist `12.0`, otool **equality**, single expected major |

### Two defects in the otool leg that only became reachable at 12.0

- It compared **upward-only** (`-gt`), so a **below-floor** slice passed. Unreachable at 10.15 (x64 could not declare lower); reachable now. Now equality.
- The extractor did `print $2; exit` — it read only the **first** slice of a fat Mach-O.

Both moved out of inline shell into `packages/electron/scripts/macos-floor.mjs` so they are unit-testable at all (they previously existed only as untested shell). The diagnostic no longer blames `MACOSX_DEPLOYMENT_TARGET`: the otooled binary is the renamed Electron **prebuilt**, whose `minos` is baked upstream and copied verbatim, so the check is an **upstream-floor tripwire** and now says so.

## The update-stream gate (the highest-risk item)

Dropping an OS for shipped users needs more than release notes. Left ungated, a Catalina client reads `latest-mac.yml`, downloads the new artifact, launchd refuses it, and the recurring check retries **forever** — strictly worse than doing nothing.

`electron-updater` has the mechanism, but **both obvious wirings are silently inert**. Both were verified against the *installed* toolchain, not the upstream README:

- **The value must be the Darwin triple `21.0.0`.** `checkIfUpdateSupported` does `try { if (semver.lt(os.release(), min)) return false } catch { warn }; return true`. `"12.0"` (marketing) and `"21"` (bare major) make `semver.lt` **throw**, the throw is **caught**, and it falls through to *"supported"* — a malformed value disables the gate instead of failing loudly.
- **It must be injected into the emitted `latest-mac.yml`.** `electron-builder.yml`'s `mac.minimumSystemVersion` never reaches update metadata (the update-info builder never writes it; its only consumer `macPackager` is skipped under `--prepackaged`). The config key would have been a plausible-looking no-op.

The injection is **macOS-only** and the injector refuses any other filename: the gate has no platform guard, so the same field in `latest-linux.yml`/`latest.yml` compares a Linux kernel (`6.5.0`) against a Darwin value and denies updates to **every** client there — the one **fail-closed** direction. The arm64+x64 merge now **rejects** a leg-asymmetric value rather than inheriting from whichever file the glob yielded first.

## Verified locally on macOS — the leg with zero CI coverage

`dmg-build-launch.electron.spec.ts` is darwin-skipped and `ci-e2e-electron.yml` runs ubuntu+windows only, so the darwin path is exactly what this change risks and exactly what CI cannot see.

- packaged app: `plutil` → `LSMinimumSystemVersion 12.0`; `otool` → `minos 12.0` (matching the empirically observed 43.4.1 prebuilt floor)
- DMG built, mounted, launched, dashboard loaded (`dmg-build-launch` green)
- `tray-ownership` green on Electron 43
- **node-pty under the execpath fallback**: loaded directly under the packaged app's own binary with `ELECTRON_RUN_AS_NODE=1` → Node `24.18.1`, modules ABI `148`, N-API prebuild loaded, real PTY round-tripped. This is the load-bearing claim in `design.md` Decision 3, previously only an inference.
- the real emitted `latest-mac.yml` injected, idempotent on re-run, and the injector confirmed to refuse a non-mac filename

## Testing

20 automated scenarios from `test-plan.md`, plus new infra the plan called for:

- `macos-floor-check.test.ts` — E4 (equality: `11` fails, `12` passes, `13` fails), E5 (multi-slice: fails on the second slice, where a first-match extractor returns `12.0` and passes), X3 (diagnostic names the upstream floor, must **not** contain `MACOSX_DEPLOYMENT_TARGET`)
- `electron-version-pin.test.ts` — E1/E2/E3, each with fixture cases that **must** fail so the guards are not vacuous
- `update-min-system-version.test.ts` — U2/U3/U4 asserted **behaviourally** through the same comparator (a string-equality test would pass on `"12.0"` while the gate is dead), U7
- `scripts/__tests__/merge-latest-mac.test.mjs` — U6, including both glob orders of the asymmetric case
- CI steps for U1/U5 (`verify-update-gate-scope.mjs`) and X4 (Ubuntu 22.04 / glibc 2.35)

These are registered at the repo root via a second, explicitly-scoped vitest project: `packages/electron` is excluded from the root run for pre-existing orphaned tests, and **a gate no CI job runs is not a gate**.

## Things I did not treat as this change's problem — with evidence

- **`zombie-adoption.electron.spec.ts` hangs on macOS.** I reinstalled and repackaged on the **old** `32.3.3` pin and it hangs **identically**, so it is a pre-existing macOS-only harness limitation, not a regression. Filed as #531. Root cause of the blind spot: no macOS leg in `ci-e2e-electron.yml`.
- **Full-suite reds** are all `Hook timed out` / `Test timed out`, differ run-to-run, and every affected project is green in isolation (server 425 files/4609 tests, extension+bus-client 141 files/1710 tests). Machine load.

Also fixes a **pre-existing** darwin bug the same blind spot hid: `resolvePackagedBinary()` derived the binary name from the `.app` bundle (`PI-Dashboard`) instead of forge's `executableName` (`pi-dashboard`) and ENOENT'd — group 7 could not run at all until this was fixed.

## Reviewer attention

- **`_electron-build.yml` step order** is load-bearing: inject (darwin) → glibc smoke → floor verify → Windows NSIS → gate-scope verify **last**, because the Windows `latest.yml` only exists after the NSIS step.
- **Unproven until this PR's CI runs:** Windows NSIS, both macOS runner legs, the new Ubuntu 22.04 glibc step, and the published-artifact metadata. The Ubuntu step could not be validated locally at all.
- **`10.1` stays an accepted, unverified gap** — no macOS 11/10.15 VM or runner exists. Automated coverage stops at the *declared* floor and the *updater* gate; launch **refusal** on a below-floor OS is verified by nothing.

## Breaking change

macOS 10.15 / 11 users can no longer install or auto-update. They keep their current install and are **not** offered this or any later update — by design, via the gate above. Intel Macs are **not** dropped (`darwin/x64` retained). Windows and Linux unaffected. CHANGELOG entry added under `## [Unreleased]`.


robertcsakany:
Thanks — all three findings were valid and are fixed in c96832561.

**1. `MACOSX_DEPLOYMENT_TARGET` set after `pnpm install`** — correct, and it mattered more than the "quick win" label suggests. The spec clause this change is supposed to satisfy says *"even if a future change introduces a native module compiled from source on the runner, the produced bundle SHALL still launch on Monterey"*, and the install-time window made that clause false. Note the otool check cannot cover for it: it inspects only the main Electron binary (which is the renamed upstream prebuilt anyway), so a drifted native module would have been invisible. Now exported to `GITHUB_ENV` in a darwin-only step before install; kept on the make step too so that step stays self-documenting.

**2. `find … | head -1` under `set -euo pipefail`** — confirmed by reproduction rather than by reading:

```
$ bash -c 'set -euo pipefail
  MERGED=$(find electron-darwin-* -name "latest-mac.yml" 2>/dev/null | head -1)
  echo "REACHED GUARD"'
$ echo $?
1          # ← no "REACHED GUARD": the step dies before the skip branch
```

With `|| true` it prints the guard line and exits 0. This one was mine — the merge step above already had the guard and I didn't carry it over.

**3. Wording** — right, and it was a real inaccuracy, not just phrasing: the gate suppresses *newer* metadata, it does not migrate anyone *to* a final 0.7.x. Below-floor users remain on whatever they currently have. Fixed in `README.md`, `docs/electron-session.md` and `docs/electron-session.agent.md`. (`CHANGELOG.md` already said "you keep the version you have installed today".)

---

**On the CodeQL `actions/cache-poisoning/poisonable-step` alert** (the remaining red): this is a new *instance* of an accepted repo-wide pattern, not a new class of risk. There are currently **16 open alerts for this exact rule**, 8 of them already in `_electron-build.yml` itself (lines 140, 144, 159, 189, 193, 211, 237, 248) plus 4 in `publish.yml` and 4 in `_smoke.yml`. The rule fires on any executing step in a reusable workflow that checks out `inputs.ref`; my new Ubuntu-22.04 step is structurally identical to the existing ones. Fixing it properly means reworking the privileged-checkout/caching model across all three workflows, which is a separate security change — I did not want to smuggle that in here. Happy to file it as a follow-up.

**`job-object-windows`** is also red, and is **pre-existing**: `develop`'s own last run of that workflow (run `28725224248`, 2026-07-05) fails with the byte-identical `[job-object-smoke] INFRA: app never brought a server up on :8000`. The script itself classifies `INFRA` as *"the test could not set up (not an invariant violation)"*. The relevant positive signal for this PR is that **`electron-e2e (windows-latest)` passes**, which is what proves the Electron 43 app actually boots on Windows.

robertcsakany:
### Round-2 review closed

Code-level findings fixed in `f76ec76a3`:
- `packages/electron/AGENTS.md` — restored path-alphabetical ordering of the `scripts/*` rows.
- `verify-macos-floor.mjs` — validate the optional `expectedMajor` argument. A typo previously coerced to `NaN`, making every slice "mismatch", which reads as a real floor failure rather than a bad invocation.
- `electron-lifecycle.ts` — the darwin fallback picked `readdir`'s first file when `pi-dashboard` was absent, silently launching an arbitrary helper. Now errors with the candidate list.

Spec-level findings, decided explicitly:

- **"Make asymmetric architecture metadata a hard error"** — accepted, and the spec was the thing that was wrong. `merge-latest-mac.py` already rejects a leg-asymmetric or disagreeing `minimumSystemVersion`; the requirement only said the field "SHALL survive the merge". Spec text updated in `af6de54eb` to state the reject-on-asymmetry invariant, including the "one leg has it, the other doesn't, in either glob order" case. The archived delta is left as the historical record of what the change *proposed*.

- **"Do not allow unavailable Mach-O validation to pass publication"** — valid, and deliberately **not** changed here. This is pre-existing spec language (`SHALL emit a ::warning:: (not fail)`), and flipping it changes the shipped CI contract. Filed as **#533** with the full reasoning from both reviewers that raised it, plus three concrete options (the canary approach looks best: keep the tolerant path, but assert extraction works against the known installed prebuilt so a format change fails loudly).

### Merging with two red checks — the evidence

- **CodeQL `actions/cache-poisoning/poisonable-step`** — a new *instance* of an accepted repo-wide pattern, not a new class of risk. 16 open alerts for this exact rule today, 8 of them already in `_electron-build.yml` (lines 140, 144, 159, 189, 193, 211, 237, 248), 4 in `publish.yml`, 4 in `_smoke.yml`. The rule fires on any executing step in a reusable workflow that checks out `inputs.ref`; the new Ubuntu-22.04 step is structurally identical to the existing ones. Fixing it means reworking the privileged-checkout/caching model across three workflows — a separate security change, not something to smuggle in here.

- **`job-object-windows`** — pre-existing. `develop`'s own last run of that workflow (`28725224248`, 2026-07-05) fails with the byte-identical `[job-object-smoke] INFRA: app never brought a server up on :8000`, and the script itself classifies `INFRA` as *"the test could not set up (not an invariant violation)"*. The positive signal that matters for this PR is that **`electron-e2e (windows-latest)` passes**, which is what actually proves the Electron 43 app boots on Windows.

Everything else is green: `ci`, both `electron-e2e` legs, and all three CodeQL analyses (including the newly-added `python` one, triggered by `scripts/merge-latest-mac.py`).

Follow-ups filed: **#531** (pre-existing macOS zombie-adoption harness hang), **#533** (unextractable-`minos` fail-open).

PR #534, opened 2026-08-22
feat(gateway): HOME-derived transport, bridge auth, session move, remote transcripts

robertcsakany:
Implements OpenSpec change `add-pi-gateway-transport-identity`. **150 of 176 tasks**; the remainder need real Windows/container hardware or are QA/E2E arms (listed below).

## Why

A bridge chose its dashboard by network discovery, so a stale instance advertising a hostname it did not serve could capture sessions from a working one. This makes the local endpoint **derived from `$HOME`**, not from the network.

## What changed

**Transport.** The pi-gateway now serves a unix socket by default — `<configDir>/gateway-<piPort>.sock`, `0600` in a `0700` dir. TCP is opt-in via `PI_GATEWAY_TCP` (docker sets it to `1`, since bridges outside the container cannot reach an in-container socket). Windows stays on loopback.

**Endpoint precedence (D3).** One resolution site, highest first: `PI_DASHBOARD_SOCKET` → `PI_DASHBOARD_URL` → pinned-instance → rendezvous record → paired-remote. **mDNS can only suggest, never win.** Absence resolves to *unavailable* — never a fallback to discovery.

**Auth.** Unix = kernel file-mode (D5). Loopback TCP = `X-Pi-Local-Token`. Remote TCP = single-use, scope-bound ws ticket, mintable only by a paired-device bearer or a genuinely-local caller. The server's Ed25519 identity is pinned by fingerprint and verified by nonce challenge before registering.

**Session move.** `/dashboard-connect`, `/dashboard-list`, `/dashboard-where`. A provisional registration claims **no** routing; routing transfers only on `session_move_commit`, so a failed move is a no-op rather than an outage. The origin renders `movedTo` instead of a bare crash.

**Remote transcripts.** Sessions are addressed **by id only** — any path-bearing field on the wire is refused. Bounded resumable backfill (offset + length + hash cursor). Origin is derived from the authenticated bridge credential, never bridge-claimed; remote-origin sessions refuse local file reads and refuse resume.

## Verified against two live instances, not just unit tests

Two dashboards under two HOMEs with **mDNS enabled**: one session each, zero cross-talk; each binds exactly one TCP port (its HTTP port) and **no bridge TCP port at all**; a same-HOME second instance attaches without rewriting the record; a live session moved between two instances with the origin showing `movedTo`.

**That run found two defects the whole green suite missed:**

1. **Cross-instance moves were impossible.** The target refused any session it did not already know — which is every real move. The unit tests passed because they modelled origin and target as the *same* gateway. The refusal also leaked exactly what it was meant to hide: accept-vs-reject revealed whether a session lived on that instance.
2. **`createPiGateway` was never passed `instanceId`**, so every destination announced `""` and the mover's identity check compared empty strings — verification that always passes.

Both fixed, with a two-gateway regression arm.

## Known gaps (deliberate, not oversights)

- `identityVerified` is hardcoded `false` at the mDNS re-target site, so an mDNS-triggered re-target is refused *unconditionally*. Fail-closed by choice; documented in design.md D16.
- Windows (§5), the container `VOLUME` check (8.7), QA/E2E arms (12.39-12.52) and perf runs against 4 MB/44 MB transcripts are **not covered by CI** — E2E is opt-in and QA arms are per-OS.
- `fix-bridge-mdns-migration-hijack` remains open on purpose: it owns the *producer*-side fix (a loopback-bound server must not advertise a LAN hostname), which this change does not address. Reconciliation is recorded in design.md D16.


robertcsakany:
@coderabbitai review

Automated review was skipped (109 files > 100 limit). This PR touches bridge authentication, unix-socket binding, session routing/move, and a remote-transcript read boundary — please review at least these paths if a full pass is not possible:

- `packages/server/src/pi/bridge-upgrade-auth.ts`
- `packages/server/src/auth/bridge-ticket-eligibility.ts`
- `packages/server/src/pi/gateway-socket-bind.ts`
- `packages/server/src/pi/provisional-registration.ts`
- `packages/server/src/pi/pi-gateway.ts`
- `packages/server/src/session/session-origin.ts`
- `packages/extension/src/transcript-request-guard.ts`

robertcsakany:
@coderabbitai All three fixed — thank you, your P2 was the one my internal audit missed.

**P1 (commit token / wire sessionId)** — `6cc32e764`. Routing, materialisation, heartbeat and logging now use `verdict.sessionId`; a mismatched wire id is refused outright. Regression test commits A's token naming B and asserts B is untouched.

Rewriting that test across **two** gateways then exposed a second defect: the target's next frame hit the auto-placeholder branch and claimed the session *before* the commit, so "a provisional claims nothing" was defeated by the following message. Invisible on one gateway, where the origin's ownership gate absorbed it. Fixed in `ee5f9ad98`, which also refuses any commit that would displace a **live incumbent** — restoring the invariant that a move grants nothing a plain `session_register` does not.

**P2 (release before acknowledgement)** — `3ef16c3d5`. Correct, and my own fixes above made it worse by adding refusal paths. Added `session_move_committed`, emitted only after routing actually transfers; the coordinator completes on that, not on its send. Note the existing "releases the origin only AFTER the commit is acknowledged" test was passing **vacuously** — its fake target never answered and nothing required it to.

**P3 (origin gate inert)** — `556da2102`. Confirmed: `originDeviceId` was only ever read, never written. A bridge ticket now carries the paired device it was minted for, the upgrade verdict carries it to the connection, and the gateway derives origin from transport + peer + credential at register. An unattributable remote peer gets `UNATTRIBUTED_REMOTE`, not `undefined` — absent means *local*, which would have silently re-opened the paths this closes. Also applied at `/api/session-file`, whose cwd confinement does not travel across hosts.

Each fix is mutation-verified (reverting it reddens exactly the new test).

Deliberately **not** changed: the tokenless-loopback grace (`requireTicketOnLoopback: false`) is a recorded deprecation decision (design.md D10b), not an oversight — raising it is a product call, not a patch.

robertcsakany:
@coderabbitai full review

PR #536, opened 2026-08-23
feat(overlays): route-backed overlay dialogs for the core surfaces

robertcsakany:
Implements OpenSpec change `add-route-backed-overlay-dialogs`.

Seven core surfaces now render as dialogs over a pinned underlay while keeping their URLs **exactly**: `/settings`, `/folder/:cwd/settings`, the three preview routes, `/tunnel-setup`, and the deep-linked OpenSpec artifact. All six bundled plugin claims convert with them.

**No e2e `goto(...)` target changed** — that was the change's own falsification test (D1). The `tests/e2e` diff deletes zero lines; every added `goto` is in a spec this PR adds.

## Design

Model **C**: the URL stays canonical and the underlay renders from a path *frozen at navigation time* (nested wouter `Router` on `memoryLocation`), so the launching surface stays visible without a second URL-derived branch. Dismissal is an explicit navigate, never a bare `history.back()`.

## Beyond the conversion

- **Dismissal guard (risk R1).** A dialog adds three gestures a page never had (backdrop / Escape / ✕). A panel-level *opt-in* guard stops them silently discarding unsaved edits; plugin claims and clean surfaces are unaffected.
- **D1b fixed.** The discard confirm now returns to the launching route instead of evicting to the card list — the original defect this change exists to fix.
- **Resource dedupe.** One scope-switched surface replaces two hand-assembled call sites and their two byte-identical, mutually-unchecked `page→type` maps.
- **Shared `Dialog` z-layer.** It painted at raw `z-[60]` — the *toast* layer — so a confirm raised from inside a dialog rendered underneath it. Now the `z-dialog` token; raw-z debt 42 → 40, baseline ratcheted twice.

## Notable findings during the work

- **D2 was never delivered.** Group 4 built a container seam and deferred injection; group 5 covered only the core surfaces, so it fell between them. Every plugin claim still rendered as a full page and **nothing failed** — the seam's unit tests passed a container explicitly. Now wired, with an e2e canary; the seam itself is removed rather than left unreachable.
- **The underlay showed the onboarding page.** The folder-home renderer was gated on a live-URL cwd, null once the overlay owns the URL.
- **Guard clobbering.** A single last-write-wins slot let an inner surface disarm the outer guard — `SettingsPanel` + `InstructionsPage` is exactly that pairing. Now an identity-scoped stack.

The last three came from an independent audit of the diff, two of them in code written in this change.

## Verification

- Full unit suite green (**15600+**), `tsc` clean, build clean.
- Gates: `check-conventions`, `dox-byte-gate`, `z-layer-lint` (40/40), `knip-ratchet` (234/234 exports, 193/193 types), `openspec validate --strict`.
- Harness e2e run per conversion, not batched at the end — every regression in this PR was caught there and invisible to unit tests.
- **CI is the real gate for the full e2e suite**: `playwright.config.ts` sets `globalTimeout: 15 * 60_000`, so a local full run cannot finish. Ran it in shards and chased every failure to a baseline built from the exact merge base.

## Pre-existing red on develop (NOT from this PR)

Confirmed by rebuilding the merge-base (`22fe62c79`) and reproducing identically — **25 tests / 7 areas**: `openspec-board-drop*` (17), `openspec-artifact-dialog` F3/F4 (2), `editor-pane` (2), `file-preview-survives-churn`, `automation-fanout` F5, `bus-client-goal-plugin-action`, `change-summary-table`.

Session-spawning specs additionally fail under accumulated harness load and pass on a fresh harness, so a long sharded run is not a trustworthy signal.

## Known gaps (tracked, not hidden)

- **9.10** — five automated test-plan rows ship undone, each with the coverage that *does* exist: codegen for `presentation` (blocked — no bundled plugin declares it), mobile swipe-back, the two C5 perf budgets (**unmeasured**), and a cross-overlay pairing round-trip.
- **9.5–9.9** — audit findings not fixed here: `App.tsx` net-grew 185 lines with duplicate render sites gated rather than removed; desktop folder-editor underlay fidelity; body-level `aria-hidden`; module state read during render; chat subtree remount on open/close.
- Manual mobile pass deferred to post-merge.


robertcsakany:
**Re: the threads asking to move OpenSpec artifacts out of `openspec/changes/archive/` — disputing these as false positives.**

Archiving is what this step of the ship pipeline *does*. `AGENTS.md` (OpenSpec Conventions) states it explicitly:

> `ship-change` MOVES a completed change into `openspec/changes/archive/<date>-<name>/`, which `scripts/check-conventions.mjs` skips as immutable history; **a review asking to move an archived change back is a false positive**.

`node scripts/check-conventions.mjs --base origin/develop` reports no violations on this branch. The artifact-location rule targets *creation* time (never create under `active/` or `archive/`), not the archive move at ship time.

The same applies to the review comments on `design.md`, `tasks.md` and the delta `specs/` under that archived path: those files are immutable history now, and their delta specs were already synced into `openspec/specs/` by the archive step. Where those threads flagged something about **shipped behaviour** rather than the artifact text, I applied it against the live files instead:

- the folder-editor underlay concern is tracked as follow-up **9.6** (it is underlay *fidelity* — the editor already unmounted on that navigation before this change);
- the codegen-coverage concern is tracked as **9.10(a)**, blocked because no bundled plugin declares `presentation`;
- the OpenSpec-artifact mobile-container point is already the resolved D6 behaviour: the full-page route is retained for mobile (e2e E9 pins it), and only the desktop path moved to the overlay.

The five findings on live files were real and are fixed in `b5bfbb8a7` — including the one that mattered most: a frozen `/session/:id` underlay was rendering the overlay's own content. Thanks for that one; it was a genuine bug my own tests missed.

robertcsakany:
@coderabbitai full review

PR #537, opened 2026-08-24
fix(custom-provider): preserve endpoint-advertised model metadata

molnar-botond:
## Why

Custom providers that advertise rich model metadata were ignored. A live probe of the configured 9router provider returns, for `cc/claude-opus-5`, `context_length: 1000000, max_completion_tokens: 128000, capabilities: { reasoning: true, vision: true }` — yet the dashboard showed `ctx=200000, maxTok=64000, reasoning=false` for that model **and every other model in the provider, identically**. That triple is verbatim the `anthropic-messages` fallback floor: nothing from the endpoint reached the UI.

Both discovery paths dropped the response body one function too early, then stamped every id with a hardcoded floor. The wrong context window mis-sizes compaction, and `reasoning: false` erased the thinking levels the model actually supports. Name-matched catalog enrichment isn't a safe default either — the same probe shows `ag/claude-opus-4-6-thinking` advertising 200k while pi's catalog lists `claude-opus-4-6` at 1M, so it can *overstate* a route's real limit just as the floors understate it. Only the endpoint knows what a given proxy route serves.

## What changed

Discovery now preserves what the endpoint advertised and merges it as a distinct tier:

    native models.json  >  endpoint-advertised  >  api-typed floor   (per field)

- **`packages/shared/src/provider-model-metadata.ts`** *(new)* — one pure mapper both discovery surfaces use, keyed on **response shape** (not the configured `api`), so an OpenAI-shaped body from an `anthropic-messages` provider still maps. Fixed field allowlist; unknown fields (incl. echoed credentials) dropped. No upper sanity bound on advertised numerics — only the endpoint is authoritative for its route.
- **Per-field validation & merge** — malformed/zero/negative/string capacities are "not advertised" and fall through to the floor for that field alone; a partially-advertising model keeps its advertised values AND the catalog/floor values for the fields it omitted.
- **`metadataSource` gains `"endpoint"`** — reported only when every projected field was advertised; a mixed-tier model reports its weakest adopted tier, so a floor value is never shown as confirmed. The client treats `"endpoint"` as confirmed — only `"fallback"` stays uncertain.
- **Endpoint tier outranks the name-matched catalog probe**, which cannot resolve prefixed/hybrid ids and over-claims when it does match.
- **Reasoning adopted without synthesizing a `thinkingLevelMap`** — a guessed table would replace "no thinking levels" with "wrong thinking levels".
- **`refactor`** — `listProviderModelIds`/`listProviderModels` now share one `fetchProviderModelBody` helper (they differ only in the final map).

Ids-only helpers (`listProviderModelIds`, `probeProvider`) are asserted unchanged so the Test button cannot regress. Built-in pi-ai models are untouched, and discovery failures (500, non-JSON, timeout, unresolvable `\$ENV` key) still yield no models rather than throwing.

## Tests

L1: 5 suites extended + 2 new — ingestion/mapping, per-field merge/precedence, provenance, and graceful degradation across **both** discovery surfaces (server + extension). L3: 1 new rendered-DOM provenance spec (harness cannot seed a metadata-rich `/v1/models`, so it proves the drivable slices; limitation documented in the spec header). Pre-existing provider-probe suite (31 tests) stays green after the refactor.

> ⚠️ **Full suite + manual QA (tasks 10.x) not yet run** — each suite passed in isolation and post-merge subsets pass, but the whole-repo `npm test` and manual verification are still pending. See `openspec/changes/fix-custom-provider-model-metadata/tasks.md`.

OpenSpec change: `fix-custom-provider-model-metadata`


molnar-botond:
CodeRabbit round 1 addressed in `17b2c87`. Per-thread dispositions (each verified against the actual code):

**Applied**
- **Keep the timeout active during JSON parsing** (`provider-probe.ts`) — fixed exactly as proposed; `clearTimeout` moved to `finally` so a stalled response *body* is aborted. Covered by the existing X2 stalled-body test.
- **Preserve/align thinking metadata** (`provider-model-metadata.ts` / discovery spec) — resolved via option (b): the implementation deliberately never synthesizes a `thinkingLevelMap` from advertised capability (guessing risks *wrong* levels; `reasoning` alone restores availability; a native map wins). Narrowed the spec requirement + scenario to match, closing the spec/impl gap.
- **Align preservation vs filtering contract** (discovery spec) — reworded to 'every *supported* capability field', agreeing with the mapping requirement that ignores unrepresentable modalities.
- **Reconcile completed validation tasks** (tasks.md) — 10.3 (rebuild/deploy) and 10.4 (live 9router check) unchecked as deferred-to-post-merge, matching the PR status; 10.1/10.2 remain checked (scope-verified).
- **Language tag on precedence block** (design.md) — added `text`.
- Removed a stray active `openspec/changes/add-overlay-layering-system/` that rode in from a pre-existing branch commit (develop already archived it 2026-08-19) — this cleared the six overlay-layering threads, which were never part of this change.

**Skipped, with rationale**
- **'Keep the OpenSpec change under the permitted path'** — false positive: archiving to `openspec/changes/archive/<date>-<name>/` is the prescribed project convention (ship-change + openspec-archive skills), and the real `check-conventions` CI gate passed green. Moving it back under `changes/<name>/` would un-archive the change.
- **'Require at least one capability title'** (E2E) — the spawned session's model list is environment-dependent, so a hard non-zero count assertion would be flaky. The test intentionally asserts the *vocabulary* of any rendered capability titles; the exhaustive provenance coverage lives in the L1 `ModelSelector.test.tsx`.

PR #538, opened 2026-08-24
feat(surface-concurrent-ask-user-prompts): surface concurrent ask_user prompts

molnar-botond:
Implements OpenSpec change `surface-concurrent-ask-user-prompts`.

## What
- **Fix the drop (correctness):** narrow `addInteractiveRequest` dedup to `requestId` equality only, deleting the `method`+`params.title` content fallback that silently dropped a second concurrent prompt sharing a title (the `update_roles` case — two parallel confirms with a constant title collapsed to one, hanging the second tool to the 5-min PromptBus timeout).
- **Grouped render (UX):** concurrently-pending free-floating asks render in one `MultiAskPanel` — a vertical stack of independently-answerable cards, each resolving its own `requestId`. Tool-paired asks (with `toolCallId`) keep their inline placement; a `method:"batch"` ask renders as its BatchRenderer wizard in one slot.
- No server / bridge / protocol / PromptBus changes.

## Tests
- Reducer units U1–U5 + R-reconnect (`event-reducer.test.ts`).
- Panel/derivation component tests C1–C6 (`multi-ask-panel.test.tsx`).
- Updated the pre-existing reducer test that encoded the removed dedup behavior.

QA/manual task 5.2 (live two-concurrent-`update_roles` check) deferred to post-merge verification.


molnar-botond:
Thanks @coderabbitai. Triage of the review:

- **Cancellation-path test** — applied (new `C2b` in `multi-ask-panel.test.tsx`): asserts a card resolves its own `requestId` with `undefined, true`.
- **"Move the change out of `archive/`"** — intended, not a regression. `ship-change` archives a completed OpenSpec change into `openspec/changes/archive/<date>-<name>/`; `scripts/check-conventions.mjs` treats that path as immutable history and skips it. Per project conventions, a review asking to move an archived change back is a known false positive.
- **MD040 fence in `proposal.md`** — that file now lives under the archived (immutable) change dir; not rewriting archived history for a lint nit.
- **`view` field description in `chat/AGENTS.md`** — pre-existing row for `ChatView.tsx` that this PR does not author (the diff only adds the `MultiAskPanel.tsx` row). Out of scope for this change under the surgical-changes rule.

PR #541, opened 2026-08-24
fix(flush-dialog): bound flush Dialog content and stop the close-control collision

robertcsakany:
Implements OpenSpec change `fix-flush-dialog-scroll-and-close-collision`.

## The defect

`Dialog`'s `flush` mode was a bare `overflow-hidden` box carrying a `max-h` cap and **no flex context**, so a child sized `h-full` resolved `height:100%` against an indefinite parent, fell back to `auto`, and grew to its content. Its own `overflow-y-auto` was therefore never bounded and never became a scroller, while the panel clipped the surplus away unreachably.

Measured live at 1440x900, five of five non-plugin overlay routes clipped their content with **zero working scrollers** — worst case **18 353 px** of a README with no scrollbar, no affordance, and no indication the document continued. Separately, the container's absolutely-positioned close control painted into a corner `flush` never reserves, landing on top of the child's own header controls on three surfaces.

The suite stayed green because all 10 route-backed-overlay e2e tests assert routing semantics, and the three jsdom overlay suites have no layout engine at all: `toBeVisible()` fails only at zero height and passes happily on an 828 px element with 537 px clipped away.

## What changed

- **`flush` establishes a capped flex column** (`overflow-hidden flex flex-col min-h-0`). The non-flush branch is byte-identical — asserted, including "no flex token leaks in".
- **`flush` suppresses the built-in close control**; a new `showClose` opt-in restores it for a headerless flush child. Mirrored into the PUBLIC plugin API (`UiDialogProps`), where it also carries the layout contract a plugin author needs.
- **Three children adopt `flex-1 min-h-0`** (`SettingsPanel`, `DirectorySettings`, `ZrokInstallGuide`). `min-h-0` is the load-bearing token, not the removal of `h-full`.
- **`OpenSpecArtifactDialog`** gains its own back affordance on all three branches and drops the `h-[85vh]` wrapper; `closeInset` is deleted outright.
- **The definite-height PINS stay**: the `h-[92vh]` plugin slot (claim bodies are `absolute inset-0` and contribute zero intrinsic height — they would collapse to 0 under a height-indefinite panel) and both `h-[70vh]` popouts (a stable height as transcripts stream). Each gained a comment saying why, so the next reader does not "finish the job".
- **`useFocusTrap`** now hands focus to the first focusable that *appears* when a dialog opened with none. Not in the original design — found during implementation: the close control used to guarantee a non-empty focusable set, and suppressing it stranded focus on the container for `/settings/general` (measured 0 focusables at mount, 51 once its data lands). Spec'd in both delta specs.

## Verification

New generic browser gate `tests/e2e/overlay-layout.spec.ts` (50 tests). Its route table is **completeness-checked** against the router's static descriptors and the plugin `shell-overlay-route` manifests, with a drift tripwire on `isOverlayRoute`'s branch tokens — a new overlay route that skips the gate fails the gate.

**Proven falsifiable, not merely green**: re-run with D1's flex tokens reverted, it fails on exactly the five previously-broken routes (14 failures, all in the geometry family). The occlusion family reports 0 failures either way — reported separately and honestly, since a composite verdict would have hidden that.

Also covered: tall/short/at-cap/resize geometry, the plugin-collapse regression (highest severity here), mobile shells, the three `DirectorySettings` mount contexts, the deepest `h-full` in the chain (the instructions editor, never measured before), the unsaved-edits guard, and an empty-state anti-vacuity check.

Unit: 16 205 tests green. Two pre-existing tests that asserted the removed close gesture were **updated, not weakened** — each now asserts the control's *absence*, so the shrunk gesture list cannot be mistaken for lost coverage.

Reviewed in two adversarial rounds by an independent model (`@review`): no blocking findings; its non-blocking findings (attribute-blind MutationObserver, a completeness check sourced from the wrong table, a bare-`button` occlusion fallback, two vacuous plugin content probes) are all fixed in this branch.

## Known-unrelated

`automation-fanout`, `bus-client-goal-plugin-action`, `change-summary-table` and `file-preview-survives-churn` fail on this harness **independently of this change** — each reproduced on a baseline build with this change's source reverted.

QA/manual task 6.14 (visual check where the close control used to sit) deferred to post-merge verification.


robertcsakany:
Thanks — 8 of 11 findings are fixed in 5366de9. Three are disputed with evidence:

**Fixed**
- `useFocusTrap.ts` hidden-ancestor focus (Major): `getFocusable` now rejects any candidate under a `hidden` / `aria-hidden` **ancestor**, walking up to the container. Regression test added (late focusable buried under `aria-hidden` → focus stays on the container).
- `overlay-layout.spec.ts` close-control selector: bound to the exact `${route.overlay}-close`. The rewrite also surfaced a real arity bug — `locator.evaluate(fn, arg)` calls `fn(element, arg)` — which is fixed.
- Reachability (Major, "a working scroller can be unrelated to the clipped content"): the gate now drives the scroller to its end and asserts the **last rendered content** lands inside the panel's box, so "a scroller exists" is no longer the whole assertion.
- `useFocusTrap.test.tsx` observer-delivery boundary: node existence is not observer delivery. Now awaits a macrotask hop before asserting focus was not stolen.
- Markdown/table findings: `Dialog.tsx` row column count (pipes inside a code span), `MarkdownPreviewView.tsx` trailing pipe, `AgentToolRenderer.tsx` / `FlowAgentCard.tsx` rows trimmed to purpose + retained contract, `tests/e2e/AGENTS.md` restored to path-alphabetical order (the split-large-agents rewrite had reordered it).

**Bonus, found while verifying your reachability point:** the tall/short preview routes were gating on the preview *shell*, which mounts before its markdown arrives — so those geometry assertions could race an empty box. They now gate on the fixture's own rendered text, and the short-content route moved off `hello.txt` (which renders "can't preview this file" — short for the wrong reason) to `notes.md`.

**Disputed**

1. **"Move these OpenSpec artifacts out of `archive/`"** — false positive. This repo's `AGENTS.md` states that `ship-change` MOVES a completed change into `openspec/changes/archive/<date>-<name>/`, and `scripts/check-conventions.mjs` skips that directory as immutable history: *"a review asking to move an archived change back is a false positive."* `node scripts/check-conventions.mjs --base origin/develop` passes on this branch. The same reasoning declines the edits proposed to the archived `design.md` / `proposal.md` / delta spec.

2. **"Keep the topmost-layer Escape contract in both specs"** — not a regression. `openspec/specs/dialog-primitive/spec.md` **does** carry both the topmost-layer sentence and the stacked-overlay scenario (lines 30-34 and the "Escape over a stacked overlay does not close the dialog" scenario). `dialog-system`'s dismissal requirement never carried it: `git show origin/develop:openspec/specs/dialog-system/spec.md` shows the pre-change requirement listing three sources with no escape-stack wording. This change's MODIFIED block carries **all three** original scenarios plus three new ones — nothing was dropped. Adding the contract to `dialog-system` is a spec-coherence improvement about behaviour this change does not touch, and belongs in its own change.

3. **Strengthening the persistent `shell-overlay-route` scenario text** — declined *as a spec edit*, accepted *as a test change*. The requirement already states the intent ("so the overflowing content is reachable"); the gap was that the test asserted less than the spec said. That is now fixed in the test rather than by weakening or re-negotiating the written contract, and the archived delta stays immutable.

PR #543, opened 2026-08-24
fix(bridge): deliver image attachments on the follow-up drain

robertcsakany:
Implements OpenSpec change `fix-bridge-followup-image-drop`.

## The bug

A follow-up prompt sent while the agent is streaming lost its images. The bridge buffered the entry as a bare string (`bridgeFollowUp: string[]`), so `msg.images` was discarded at the buffer call site and the drain shipped text only. The failure was **silent**: the client accepted the image, the chip rendered the text, and the model never received the attachment.

The spec was internally contradictory on this point — it required "Drain preserves image attachments" while defining the buffer as `string[]`. This reconciles the two halves in favour of delivery.

## What changed

- **Buffer carries entries.** `bridgeFollowUp` becomes `{ text, images? }[]`, extracted into `packages/extension/src/followup-buffer.ts` so the admission arithmetic is reachable by tests and the byte ceiling is injectable.
- **Drain delivers them.** The drain sends `buildUserMessageContent(entry.text, entry.images)` with **no send options** — `{deliverAs:"followUp"}` there is a known-broken path (pi has already exited `getFollowUpMessages()`, so the entry would never drain). The extraction deliberately splits validation + content assembly from send options so the two cannot be confused.
- **Wire carries a count, not bytes.** `pendingQueues.followUp` becomes `{ text, imageCount }[]`; the chip shows an attachment indicator, never a thumbnail. Image bytes stay extension-side. **BREAKING** for `followUp` consumers; the client normalises legacy `string[]` for one release (client-first rollout).
- **Buffer is bounded in bytes.** A 32 MiB aggregate ceiling joins the existing 20-entry cap. The pre-change worst case was `20 x N x 10 MB`, unbounded in N — there is no per-send image-count cap anywhere. The total is recomputed from live entries at each admission, never accumulated, so every mutation site releases bytes correctly without knowing the budget exists. Enforcement is refusal, never eviction; entries are refused whole and images are never stripped to fit.
- **Refusals are visible.** All four refusal sites now emit `command_feedback` instead of a silent `console.warn`.
- **Shape-safe accessors.** Sizing and MIME validation read through `imageBlockData`/`imageBlockMime`. A direct `.data` read sizes a nested Anthropic block at **zero** — an unbounded hold that passes every ceiling check — and drops it as "invalid mime".
- `edit_followup_entry.images` retired (design D5): unpopulatable under the count-only wire.

## Verification

- **L1**: 4 new/extended suites against the real modules — byte-exact sizing, BVA on the ceiling (just below / exactly at / just above), both bounds proven independent, mutation semantics, drain payload shape. Falsifiability spot-checked: reverting the D3c accessor turns the nested-shape cases red.
- **L3**: 7 Playwright rows (F1-F3, F5-F8) green against a docker harness built from this branch. F8 asserts the drained row's **image slot**, not just its text — without that the assertion would have held before the fix too, leaving the drain call site unpinned.
- Enforcers green: check-conventions, dox-byte-gate, i18n-lint `--strict`, i18n-parity, knip-config, knip-ratchet + baseline-diff, z-layer-lint.
- Local review: 2 rounds, no blocking findings; all non-blocking findings fixed.

`origin/develop` was merged before the harness ran, so the gate validated the integrated tree.

## Known-unrelated red

`packages/server/src/__tests__/faux-session.integration.test.ts` (5 cases) fails on this branch **and on clean `origin/develop`** — verified in a throwaway probe worktree. Not introduced here.

QA/manual task 16.4 (confirm against a real model) deferred to post-merge verification.


robertcsakany:
Addressed 9 of the 10 threads in 39e0e07c3 — the validation ones (non-array `images` container, duplicate/fractional `clearIndices`, fractional/infinite `imageCount`) were real and are each now pinned by a test. Also fixed the contradictory abort heading, the `currentIndex`-on-refusal wording, the stale "capacity 1" note, the pipe characters that split the `command-handler.ts` row into 8 cells, and the e2e row's alphabetical position.

**Disputing one:** "Do not mark deferred real-model validation as complete" (task 16.4).

This is the repo's deliberate ship convention, not an oversight. The `ship-change` skill specifies: when only QA/manual tasks remain, *"flip each deferrable `- [ ]` → `- [x]` (validated post-merge)"* — a checked box in an **archived** `tasks.md` means "this change is shipped", not "a human ran this". The deferral is stated in three places that a reader reaches before the checkbox: the archive commit message, this PR's description, and the merge commit.

Leaving it unchecked would instead break the archive gate, which requires a change to be fully task-complete before it can be archived and its specs synced — so the change could not land at all without either this flip or a fabricated claim of manual execution.

Worth noting the automated evidence already covers the delivery path this task would re-confirm by hand: the L3 `F8` case drives an image-bearing follow-up through the real bridge in the docker harness and asserts the drained chat row renders an actual image slot, and the L1 suite pins the exact content-array payload with zero send options. Task 16.4 adds real-provider confirmation on top of that, which is why it is post-merge rather than blocking.

robertcsakany:
On the last open thread — "Document the full `npm test` result and relocate this artifact" — both halves are declined, with evidence.

**1. `npm test` is green.** The authoritative run is CI, and the `ci` job passes on this PR (26m59s, run `32781253398`) — `faux-session.integration.test.ts` included and passing there. The 5 failures I saw were on my local host only. I verified they are not mine by checking out clean `origin/develop` in a throwaway probe worktree and reproducing the identical 5 failures with none of my code present, so task 16.1 ("`npm test` green") is accurate as written. The local-host artifact is already disclosed in the PR description under "Known-unrelated red" rather than hidden.

**2. The archive location is correct and deliberate.** From `AGENTS.md`:

> `ship-change` MOVES a completed change into `openspec/changes/archive/<date>-<name>/`, which `scripts/check-conventions.mjs` skips as immutable history; **a review asking to move an archived change back is a false positive.**

`openspec archive` performed that move as the designed final step of the change lifecycle — it also synced the delta into `openspec/specs/mid-turn-prompt-queue/spec.md`, which is the durable artifact. Relocating it would break the convention check's contract and orphan the spec sync.

The other thread ("do not mark 16.4 complete") was addressed in my earlier comment: it is the repo's documented ship convention, and leaving it unchecked would block the archive gate that lets the specs sync at all.

Thanks for the review — the validation findings in this batch were genuinely good catches. The non-array `images` container in particular would have thrown and failed an entire prompt rather than reporting an unusable attachment; all three are now fixed and pinned by tests.

PR #548, opened 2026-08-24
test(verify-subagent-pull-under-load): verify the four scenarios the payload change shipped unverified

robertcsakany:
## Why

`reduce-subagent-details-payload` (archived 2026-08-15) shipped the push/pull split for subagent timelines, but **four of its manifest scenarios were never verified** — none was observable on the harness, whose subagents finish in ~600 ms, so there was never a mid-run window in which a timeline grows while somebody watches it. Carrying that gap as prose inside an archived change makes it invisible. This change makes it work.

| id | scenario | now |
|---|---|---|
| **F1** | a mounted inspector converges 5 → 30 rendered entries with no close/reopen | ✅ verified |
| **P4** | resync reply bytes/s ≤ the push bytes/s the strip removed | ✅ **PASS by ~36×** |
| **P5** | a representative inspector-open share (C4 abort > 50 %) | ✅ four watch patterns; **not triggered** |
| **X1** | pi killed mid-run, no terminal frame → replay is scalar, not corrupt | ✅ pinned |

**No production behaviour change.** Only `qa/fixtures/`, `tests/e2e/`, `docker/`, one L1 test, and the change's own openspec dir.

## The substrate, and why it isn't a real faux subagent

A nested faux subagent **cannot** be scripted here — its inner `createAgentSession` resolves a different faux core with an empty response queue, so it dies after ~2 no-op turns. That was root-caused and written off upstream in `reduce-bridge-tick-bandwidth`, which routed around it with a synthetic `Agent`-tick producer. This change extends that producer with two **sentinel-gated** behaviours:

- `[[entries:5..30@60]]` — the timeline grows then **PLATEAUS** for ~9 s. The plateau is load-bearing, not padding: terminal frames are never stripped, so without it the rendered count would converge on `subagents:completed` and F1 would pass while the pull path never ran.
- `[[bus:250]]` — real `subagents:*` frames at the real producer's `PROGRESS_THROTTLE_MS`. These are what make the pull path exist at all: the strip, the frame buffer and the resync responder all live on that channel.

Both are off without a sentinel, so the existing throttle rows stay byte-identical (pinned by E7).

## How the rows avoid being vacuous

Every classification keys on **`__resyncRequestId`**, never on eventType — a resync reply and a pushed frame are *both* `subagent_started`.

F1's convergence is attributed by **carrier exhaustion**, with all four carriers accounted for:

| carrier | excluded by |
|---|---|
| terminal frame (never stripped) | the assertion window closes while the agent is non-terminal |
| fat pushes | 0 of 102 non-terminal pushes carried a timeline |
| open-time resync (fires — the *rendered* timeline is empty at mount) | by CONTENT: the open-time reply carried **5** entries |
| **cadence** | positively: the converging reply's token **equals** a captured `reason:"cadence"` request id |

`F4` inverts `F2` on a `PI_DASHBOARD_SUBAGENT_STRIP=0` harness start — without it, "0 fat pushes" could equally mean "the fixture never produced a timeline".

`X1` runs **unwatched by construction** and asserts it: a single resync stores a *fat* reply, which would fail the row for a reason unrelated to the regression. Assertions are made on the post-kill replay stream, since `force_kill` closes the bridge WS before the signal.

## Measured

| arm | value |
|---|---|
| pull (reply bytes/s, per subscriber) | **1 059.8** (spread 9.7) |
| push removed (strip OFF − strip ON) | **37 654 B/s** |
| verdict | **PASS, ~36×** — flat across a 100/250/1000 ms bus-cadence sensitivity table |
| inspector-open share | 0 % / 17.7 % / 29.9 % / 99.6 % across four stated watch patterns |
| harness ceiling | lossless to N=5 subscribers (measured, not assumed) |

Full numbers + provenance: `openspec/changes/archive/2026-08-24-verify-subagent-pull-under-load/heap-evidence.md`, derived from `measurements.json` (24/24 values verified mechanically).

## Review

Two rounds through an isolated `@review` gate. Round 1 raised a blocking scope finding (a bundled production commit → split out as #546) plus a dead `tool_execution_end` branch that made X3's claim vacuous, a silent-zero reply rate that could have written `0` into the evidence, and an F6 check that asserted wall-clock ordering while claiming a content check. Round 2 caught 7 stale transcribed numbers. All fixed and re-verified green.

Deferred (manifest `manual-only`): 7.2 visual check of the replayed X1 card, 7.3 human read of the evidence file.


robertcsakany:
## CodeRabbit triage — 7 fixed, 1 disputed

Commit `review(coderabbit): …` addresses all 8 threads. Every code fix was re-verified against the docker harness (F1/F2/F3/F5/F6/F7 + X1/X2/X3 green; P2/P3 pull arm green, medians unchanged).

| # | Finding | Disposition |
|---|---|---|
| 1 | `requireReplies` dropped on retry | **Fixed** — a real hole. A pull arm whose first spawn failed could retry with reply validation off and record `replyBytesPerSec: 0` — exactly the silent zero the flag exists to prevent. |
| 2 | `SUBAGENT_STATUS_OF` misses `data.result.details.status` | **Fixed** — that is the slot `tool_execution_end` uses, and the one `ENTRY_COUNT_OF` already read; any status filter silently missed the terminal carrier. |
| 3 | Evidence says "0 of 27 pushes", table says 102 | **Fixed** — stale prose from before the numbers were re-derived. |
| 4 | `Generated: 2026-08-27` is in the future | **Fixed** → `2026-08-24`. |
| 5 | `tests/e2e/AGENTS.md` rows not path-alphabetical | **Fixed** — `subagent-pull-*` now precedes `subagent-thin-*`. |
| 6 | F6's `< 30` bound would pass a 29-entry open-time reply | **Fixed — the count bound was weak.** Added the token assertion: the reply that carried the converged timeline is asserted **not** to answer any open-time request. With F5 (its token *equals* a captured `reason:"cadence"` request id) the carrier attribution is now positive on both sides. |
| 7 | Per-file purpose records too verbose | **Fixed via the repo's own tooling** — `node scripts/split-large-agents.mjs --write` promoted long rows to pull-only sidecars per AGENTS.md's ">200 chars → sidecar" rule: `qa/fixtures` 12 084 → 1 047 B, `docker` 20 438 → 2 937 B, `tests/e2e` 23 478 → 21 022 B. Dir rows now carry a summary + `→ see`. |

### Disputed

**#8 — "Move synthetic-agent parameters into fixture configuration."**

Declining, with reasons:

1. **The sentinel values are not agent configuration — they are the test's workload, and belong at the call site.** `[[ticks:240@50]][[entries:5..30@60]][[bus:250]]` *is* the scenario: a 12 s run whose timeline grows 5 → 30 over the first ~3 s and then plateaus. Hiding it behind a profile name pushes the one thing a reader must check — *does this workload actually produce a mid-run window?* — one indirection away.
2. **The numbers are load-bearing and individually justified in the diff.** `250` matches the real producer's `PROGRESS_THROTTLE_MS`; the `60` growth-ticks exist so the ~9 s plateau is ≥ 3 × `CADENCE_BASE_MS`; the `-bus100` / `-bus1000` variants exist *precisely* to vary one of these numbers so the P4 verdict ships as a sensitivity table rather than arithmetic on a single constant. A shared profile store invites editing one of them without the reasoning attached.
3. **No such configuration mechanism exists**, and this repo's guidance is explicit that speculative abstraction is a defect ("No speculative features/abstractions"; "Would a senior engineer call this overcomplicated?"). Introducing a fixture-config layer for one fixture, inside a test-only change whose contract forbids production edits, is the wrong direction.

The sentinel mechanism (`[[ticks:N@Mms]]`) is the established pattern here, introduced by `reduce-bridge-tick-bandwidth`; these two additions follow it deliberately and are fully sentinel-gated, so no pre-existing row changes behaviour.

PR #555, opened 2026-08-25
feat(add-tail-only-replay-window): tail-only replay window

robertcsakany:
Implements OpenSpec change `add-tail-only-replay-window` (issue #521 — the transcript opens on recent messages and grows upward as you scroll).

## What changed

- **Server** — `memoryLimits.replayWindowMode` (`head-tail` | `tail-only`), threaded like `maxReplayEvents`. In `tail-only`, `computeReplayWindow` returns `headEnd 0`; `session_state_reset` moved into `sendEventBatches` so a windowed replay always resets first; `GapState.hasHead` stops a phantom head credit when `from === 1` against `headMaxSeq 0`.
- **Client** — backfill floors at `oldestGapSeq`; an empty response resolves to a *terminus* (session start / not retained) instead of "unservable"; a pure `shouldAutoLoadHistory` predicate drives the auto-load, with one shared programmatic-scroll suppression window.
- **Settings** — a "Replay window shape" select, disabled with a stated reason when `maxReplayEvents` is 0.

## Verification

- 16,731 tests passing, `tsc` clean, client build clean.
- All five enforcers green (`check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity`, `knip-ratchet`), the last verified inside a **freshly built** harness image — the cached image had been producing a false green.
- E2E against the docker harness: 5/5 trigger-suppression, 3/3 loading-head, 3/3 settings, 4/4 splice-anchor.

## Defects this change found and fixed

- **A pre-existing selection bug** (exposed, not caused): `selectionSpanRef` stores row *indices*, so a splice shifted them, unmounted the selected row and collapsed the range. Fixed by remapping during render — a layout effect is too late, since `rangeExtractor` runs while the virtualizer computes its range.
- **An a11y defect**: the auto-load live region lived inside a virtualized row, so the virtualizer unmounted it before the announcement — screen readers said nothing. Hosted outside the list now.
- **Three from a security/perf audit**: stale scroll intent causing an unbidden fetch; a stranded `pending` leaving a permanent spinner; and an O(gap) slice per walk step (`countEventsRange` now counts without allocating).

## Notes for review

- **Task 7.1 (#P2)** is an *advisory* opt-in probe (`PW_PERF=1`), following `chat-render-perf.spec.ts` — absolute frame budgets in a shared container flake as CI gates. Its first draft passed **vacuously** (`distancePx: 0`); it now asserts distance travelled and rendered-row growth so it cannot pass without exercising the walk.
- **The knip ratchet was red on `develop` independently** throughout. Measured per-item each round, this branch adds zero unused exports/types; a few unrelated module-private symbols were unexported here to hold the ratchet at baseline. The durable fix belongs on `develop`.
- **Scroll-to-turn is unavailable in `tail-only`** by construction (turn anchors sit in the gap). F19 pins that assumption and goes red if turn bars ever become clickable.
- QA/manual tasks 8.1–8.3 (`#M1`–`#M3`) are marked done for **post-merge** verification.


robertcsakany:
@coderabbitai review

PR #567, opened 2026-08-29
feat(collapse-pairing-into-gateway): one pairing surface on Gateway; Security links to it

robertcsakany:
Implements OpenSpec change `collapse-pairing-into-gateway`.

Collapses the duplicated pairing UI to the ONE Gateway "Connect a device" surface:
- `noSecureRoad` block keyed on the `no_reachable_endpoint` response (never an unloaded payload / endpoint count) — explain + Set-up-the-Gateway action + localhost escape hatch, per mockups/gateway-empty.html
- `onSetupRequested` prop (dialog → Setup tab; page → focus provider section; fallback navigates + scrolls `#connect-a-device`)
- full selectable fingerprint + payload `urls[]` in the pairing context panel
- Settings ▸ Security renders a link (`security-pair-link`), no pairing controls; `PairingView` and orphaned `QrCodeDialog` deleted; single encoder pinned by structural tests
- `guardPairingUrls` loopback-http exemption (mirrors the server test-origin rule); all other non-TLS entries stay fail-closed
- 15 i18n keys retired, 5 new keys in both catalogues

QA/manual tasks (F6 physical camera scan, F7 visual pass) deferred to post-merge verification.

Test plan: 22 automated scenarios green (L1 vitest + L3 docker harness); 7 deterministic enforcers green; @review (deepseek-v4-pro) — SHIP, no blocking findings.


robertcsakany:
@coderabbitai full review

robertcsakany:
CodeRabbit round 1 triage — 3 of 8 applied, 5 rejected with reasons:

APPLIED
1. docs/architecture.md — the noSecureRoad block vs endpoint-count empty distinction (committable suggestion taken verbatim).
2. i18n-hu `beolvassad` → `beolvasod` — grammar fix accepted.
3. tests/e2e/pairing-qr.spec.ts F5 cleanup now restores the SAVED `publicBaseUrls` instead of writing `[]`.

REJECTED
4. tasks.md "do not mark deferred manual verification complete" — flipping manual-only rows is this repo's ship-change convention (manifest-aware defer, step 1: manual-only rows are marked done and verified post-merge).
5. GatewayPage section reorder (QR before provider) — the provider-first order is the reviewed design; F3 asserts land+scroll, which works from any scroll position; reordering would churn a reviewed surface for a nit.
6. "State the loopback exception in the pairing note" — prod never emits loopback http into payload `urls[]` (server gate); the same-machine path is covered by the escape-hatch copy; new user-facing strings would spend i18n budget on a test-only case. The exception is documented at the guard.
7. "Use a link, not a button" — the Security link deliberately mirrors the existing `Open Security →` button (Nielsen #4 consistency, ui-plan.md variant A1); the handler needs JS for the land+scroll regardless.
8. "Move the sidecar record into tests/e2e/AGENTS.md" — the per-file sidecar + one-line dir row is this repo's docs convention (`pairing-qr.spec.ts.AGENTS.md` is the direct precedent).

PR #572, opened 2026-08-30
chore(deps): pi 0.84.4, openspec 1.11.0, dependency refresh

robertcsakany:
## What

Routine-looking dependency work that turned up two real bugs. Full suite green: **1520 files, 17236 tests, exit 0**.

## Version pins

| Pin | From | To |
|---|---|---|
| `@earendil-works/pi-coding-agent` (5 governed sites) | 0.84.1 | **0.84.4** |
| `@fission-ai/openspec` (server + extension + guard) | `^1.6.0` | **`1.11.0` exact** |
| biome / playwright / fastify-compress / fastify-static / jest-dom / react-virtual | — | 2.5.11 / 1.62.1 / 9 / 10 / 7 / 3.14 |

`piCompatibility.minimum` stays **0.78.0** — the support floor is an independent policy value and does not track the pinned runtime.

## Two bugs found along the way

**1. Empty OpenSpec panel on any fresh `HOME`.** The openspec CLI prints a telemetry notice *ahead of* its JSON on first run under a given `HOME`. `JSON.parse(stdout)` throws on it, `parseJsonOrNull` returns `null`, and every caller reads that `null` as **"no changes"**. Any new Docker container, new user, or CI runner showed an empty panel on first poll. Fixed by retrying from the first JSON opener; no-opener and trailing-junk still return `null`.

**2. A test that had silently stopped asserting.** The harness gives every test an ephemeral `HOME`, so `openspec-poller-parity` had been hitting its CLI-absent skip branch via that same parse failure — asserting nothing, reporting green. Any real parity regression would have sailed through. It also had a timing cliff of its own: 85 sequential CLI spawns ran **59s against a 60s timeout**, version-independent and worsening with every change added to the repo. Bounded concurrency pool → **9.5s**, and it now genuinely runs all 85.

Also fixes a latent product bug via the new pi events: `session_compact_failed` clears the `compacting` latch, which a failed or aborted compaction previously left stuck, blocking reloads.

## A correction worth recording

I first reported openspec 1.11.0 as a **breaking change** and reverted it. That was wrong — I had probed with positional `openspec status <name>`, but the poller calls `openspec status --change <name> --json`, which 1.11.0 supports fine (identical artifact shape, one additive key, same speed). The revert was undone. The pin is nonetheless **exact rather than a caret**, for a different and real reason: the vendored `.pi/skills/openspec-*` are *generated by* the CLI and carry its version in `generatedBy`, so the two must move together under review. A caret had already let 1.11.0 arrive silently during a routine install — the exact pin is what stops that recurring.

## Deliberately held back

- **`@earendil-works/pi-ai`** stays `^0.75.5`. 0.84 moved `streamSimple`/`getModels`/`getProviders` out of the main entry into `./compat` and emptied `dist/oauth.js`; the model proxy and bridge resolve the hoisted bare import. That's a migration, not a bump.
- **`clear_queue`** not wired: it needs server RPC + protocol + client + TUI changes and contradicts the documented read-only shadow-queue design.
- **Risky majors** (typescript 7, vite 8, electron 44, jsdom 29, pdfjs 6, monaco, sharp, `@vitejs/plugin-react` 6, `@google/genai` 2, …) each want their own change.

## Unrelated failures fixed

Three failures pre-dating this work: a vitest `maxWorkers` grouping abort in `context-budget`, and two **cross-suite `localStorage` leaks** (client vitest shares one localstorage file run-wide) — an i18n suite persisting `'hu'` and `useTheme.test.ts` calling `localStorage.clear()`. Both fixed **at the leaking suite** rather than stubbed in the victims, since the stub-the-victim approach just moves the flake. Plus `context-budget` registered in `knip.json` and the publish allowlist.

## Specs

Four specs contradicted by the above are updated in place (`openspec validate --specs`: 565 passed). Notably `openspec-cli-version-single-source` had the caret as its *stated rule* — that requirement is inverted, not just renumbered, and gains a scenario making a caret count as drift. Two new `server-openspec-polling` requirements cover the preamble tolerance and the parity test's skip/timeout discipline.

## Review notes

- Spec edits were made **directly to main specs**, overriding the usual change/delta-sync convention — explicitly agreed, as items 1-3 are catch-up on already-landed decisions.
- `openspec/groups/groups.json` (dashboard-generated) and a `kb.mjs` file-mode flip were **excluded** as noise.
- Pre-existing and untouched: `openspec validate` fails on `package-browse`, which declares `### Requirement: Install progress feedback` twice.

## Verification

`pnpm test` exit 0 · `tsc --noEmit` clean · `verify-release-deps.mjs` 11/11 · `/api/health` → `minimum 0.78.0, recommended 0.84.4, current 0.84.4`, no skew · build + restart + reload done.


robertcsakany:
Disputed — the premise ("the server cannot identify the blocking prompt") does not hold for the bridge path.

A dashboard-routed prompt is surfaced by the PromptBus protocol (`prompt_request` → interactive UI row → response), which is independent of `ui_prompt_start`/`ui_prompt_end`. In `packages/server/src/session/event-status-extraction.ts` those two events only set/clear `currentTool` (cosmetic status text); they are not the mechanism by which the dashboard learns a prompt is open.

Emitting them from the patched `ctx.ui.*` wrappers would be a behavioural addition to the bridge, out of scope for a dependency-refresh PR, and would double-report status for prompts that already flow through the bus. Filing as a follow-up if the cosmetic `currentTool` flip is wanted for the headless path.

robertcsakany:
Acknowledged but not changed here. The diff already moves this file from `localStorage.clear()` (which wiped concurrent suites' keys — it flaked `useFolderUrgencySort` and `editor-pane/theme-follow`) to touching only the two keys this file owns, so it is strictly less racy than before.

True per-file storage isolation means a vitest-level change (per-file `--localstorage-file` or a serialized pool), which is a config change well outside a dependency-refresh PR and would affect every client suite. Leaving it as a separate follow-up rather than bundling it here.

PR #576, opened 2026-08-31
feat(client): dispatch provider-auth-event on successful credential writes

robertcsakany:
Implements OpenSpec change `dispatch-provider-auth-event`.

## What

Completes the **dispatcher half** of the shipped `provider-auth-event` contract. `useProvidersReady()` has always listened for the event; nothing ever fired it. Now:

- `useProvidersReady.ts` exports `PROVIDER_AUTH_EVENT`; the listener uses the constant.
- `ProviderAuthSection.handleChanged` dispatches it (zero-payload) — the single funnel for all five credential-write paths (API-key save/removal, OAuth sign-in/out, device-code completion); never in `refresh` (a mount is not a write).
- `SettingsPanel` dispatches it on the `/api/providers` PUT success branch only (replace semantics cover add/edit/delete; over-dispatch accepted per spec).

## Verification

- 12 new L1 rows across the three suites (test-plan D1–D6, E1–E5, X1–X4, R1–R2); **D5 verified to fail on a revert** of the two dispatch sites (non-vacuous wiring proof).
- Full suite: 17250 passed; the only reds are (1) `openspec-poller-parity` — **pre-existing on `origin/develop`** (verified on a detached checkout) and (2) a load-sensitive extension timing benchmark that passes in isolation (8/8).
- Live smoke in the isolated docker harness: OAuth sign-out inside the settings RouteBackedOverlay flipped the landing checklist ① done→pending with no reload and no window focus.

## Ship-gate review (model `@review`, review-code rubric)

**VERDICT: pass** — zero blocking findings. Carried suggestion for follow-up: D6's `JSON.stringify(ev)` assertions are vacuous (`detail` is a prototype accessor); the direct `expect(ev.detail ?? null).toBeNull()` is the real no-payload enforcement — the stringify lines could be dropped for clarity in a later cleanup.

Specs synced + change archived (`2026-08-31-dispatch-provider-auth-event`). No QA/manual deferrals.


robertcsakany:
Triage of the two CodeRabbit findings — both disputed, neither applied:

**1. `ApiKeyRow.handleSave` should honor `success: false` (Minor) — dead code; rejected by the reviewed design.**

The endpoint's contract (`packages/server/src/routes/provider-auth-routes.ts:225-241`) has NO body-level failure class: every failure is non-2xx (`400 {error}` for missing provider/key, `500 {error}` on write failure); success is unconditionally `{ ok: true }`. The suggested guard `data?.success === false` can never fire on that body (`success` is not a field; `undefined === false` is `false`) — it is a gate that can never fail, which `proposal.md` explicitly considered and rejected ("checking it would add a gate that can never fail while forcing the dispatch out of the single `handleChanged` funnel"). The spec's body-level-failure scenario targets the one path that HAS a discriminator — `SettingsPanel`'s `data.success` — and is covered by test-plan row X2. Transport failures on this endpoint throw before `onChanged()` and are covered by X1 (non-2xx → no dispatch, error renders). No change.

**2. Move the change directory out of `archive/` (Minor) — documented false positive.**

Archiving completed changes into `openspec/changes/archive/<date>-<name>/` is this repo's ship convention (AGENTS.md, OpenSpec Conventions): `ship-change` moves the completed change there and `scripts/check-conventions.mjs` treats the archive as immutable history. The review asking to move an archived change back is called out in AGENTS.md as a false-positive pattern. The directory stays.

Both findings verified against current code and controlling docs per the review-triage discipline; the auto-apply bar ("clearly-safe, localized fixes") is not met for #1, and #2 asks for a convention violation.

robertcsakany:
@coderabbitai full review

PR #587, opened 2026-08-31
feat(fix-node-family-resolution-gaps): npx managed-runtime slot + rejected-override badge state

robertcsakany:
Implements OpenSpec change `fix-node-family-resolution-gaps` (archived + specs synced).

## What

- **FIX** `npx` strategy chain: `managedRuntimeStrategy("npx")` inserted between `bundled-node` and `managedBin` (`definitions.ts`) — an installed managed Node runtime (`~/.pi-dashboard/node/bin/npx`) is now visible to the whole node/npm/npx family.
- **FIX** `StatusBadge` (`ToolsSection.tsx`): a REJECTED override on an unresolved row now renders a **third badge state** (orange ⊘), distinct from both the plain not-found (red ✕) and the rejected-but-fell-back (amber ⚠) states; tooltip names the rejected path via `rejectedOverrideDetail`, degrading to the reason text when the rejection carries no extractable path. Trail-less payloads no longer throw.
- **ALIGN** `openspec/specs/tool-registry/spec.md`: npm chain corrected (no `managedBin`; win32 `npmCliBesideNode` documented), npx chain gains `managedRuntime`, stale "does not modify the npx registration" paragraph dropped, managed-runtime family-visibility scenario added. DOX rows updated (`definitions.ts`, `ToolsSection.tsx`).
- i18n: `common.overrideRejectedDidNotResolve` added in en / zh-CN / hu.

## Verification

- TDD: 17 automated manifest scenarios authored RED first; every red verified to fail for the right reason; 48/48 touched-suite tests green post-fix.
- Full monorepo suite green (2 failures pass in isolation — port/timeout contention flakes).
- Enforcers green: check-conventions, dox-byte-gate, i18n-lint --strict, i18n-parity, knip-config, knip-ratchet, knip baseline-diff.
- `@review` (isolated): **zero blocking findings** (1 suggestion: tooltip couples to the `invalid: path does not exist:` literal — the degrade path covers a reword; 1 nit: `invalidOverride`/`rejectedOverride` naming overlap).

## QA / manual (deferred to post-merge)

- 8.1: visually confirm the three badge states are tellable apart in Settings → Tools (manual-only per test-plan #F8).

## Reporter follow-up (tasks 9.1–9.3)

To close the loop on the original report (Linux AppImage; `node`/`npm` not found, `npx` = v22): please run and paste:

```bash
curl -s localhost:8000/api/tools | jq '.data.tools[]|select(.name|test("^(node|npm|npx)$"))|{name,ok,path,source,tried}'
cat ~/.pi/dashboard/tool-overrides.json
ls -l ~/.nvm/versions/node/
which -a node npm npx
echo "$APPDIR / $APPIMAGE"
```

Expectations (9.2): this change does **NOT** fix your `npx` = v22 row — your snapshot shows no managed runtime (`node` was not found), so ordinary PATH fallback fully explains it. What it does: a rejected override is now visible on a not-found row, so the next such report arrives with the answer on screen.

Task 9.3 (open, trigger pending evidence): if the `node` row's trail shows `appimage-self-hit:`, that confirms the live Rule-2 candidate named in the proposal's "Not in scope" — a follow-up change scoping `isAppImageSelfHit` to the launcher-name collision it was written for will be opened once this trail arrives.


robertcsakany:
Keeping `[x]` — this is the project's shipping convention, not a completion claim. `ship-change` step 1 mandates: "Flip each deferrable `- [ ]` → `- [x]` (validated post-merge)" for rows mapping to `manual-only` manifest dispositions (here: test-plan #F8, tagged inline `(test-plan: manual-only)`); step 11 then reports "QA/manual tasks were marked done for **post-merge** verification". The task text itself carries the deferral ("**(test-plan: manual-only)**", group header "Manual verification (deferred post-merge)"), and the PR's "QA / manual" section re-states it. Reverting to `[ ]` would re-classify the row as a blocker at the next defer checkpoint — the manifest-aware defer rule is what this checkbox state encodes.

PR #592, opened 2026-08-31
feat(automation): work-source fan-out via schedule.batch

molnar-botond:
## Summary

Adds a generic, fenced **work-source** competing-consumers contract so the automation engine can fan out **one child session per leased work item**, driven by a new `on.kind: schedule.batch` trigger with an `on.source` naming a registered work-source. The engine stays **domain-free** — what an item is lives entirely behind the contract.

OpenSpec change: `automation-work-source-fanout`.

## What's included

- **`shared/work-source.ts`** — `WorkSource` + `LeasedHandle {item, leaseToken, idempotencyKey}`; `ack`/`nack` conditional on the token being current (stale/expired = no-op).
- **`work-source-registry.ts`** — stable, lease-stateful registry keyed by id.
- **`folder-work-source.ts`** — reference source: lease-by-rename into `inflight/<token>/`, visibility-timeout + orphan (restart) reclaim, data-loss-safe return-to-pool, item-derived idempotency key.
- **`schedule-batch-trigger.ts`** — cron timer the engine drains on fire.
- **`automation-schema.ts`** — validate `on.source`; require single `action:`, reject `actions:`/`count` for `schedule.batch`.
- **`engine.ts`** — `startWorkSourceFire`: leases **after** the concurrency policy admits the fire; one child per handle with per-child `${{trigger}}` + injected idempotency key; lease released on **every** finalize path (done→ack, error/stopped/death/spawn-fail→nack; expiry auto-releases); empty vend → no-op, `next()` throw → errored+nothing leased; excess deferred unleased.
- **`resolve-children.ts`** — `settingsDefaultBound` (per-automation → `PI_AUTOMATION_MAX_CONCURRENT_SPAWNS` env → 4).
- **`index.ts`** — build the registry from plugin config; thread the env default.

## Tests

- **26 L1 scenarios** (folder-work-source, schedule-batch-schema, work-source-fanout) — **all green**, plus **6** added from two review rounds (restart reclaim, colliding-name nack, expired-ack fence, stop-abort-rejects, sync-setup-failure, runner-slot safety).
- **L3 e2e F3** (`tests/e2e/automation-fanout.spec.ts`) — a `schedule.batch` fire fans out one child per leased file (3 files → 3 children); **verified green** against the docker harness.
- Full `automation-plugin` package: 316 pass (1 unrelated pre-existing `automation-watcher` env flake).

## Review

Two rounds of isolated `review-code` review (`@propose-review-1`); all blocking lease-lifecycle findings fixed (restart stranding, data-loss on failed return, expired-token fencing, synchronous-setup lease release, runner-slot safety).


molnar-botond:
Thanks for the thorough review — addressed in the latest push (af79c0136).

**Fixed**
- **folder-work-source idempotencyKey** — now `sha256(name+size+mtime)`, so a new file that reuses a leased name gets a distinct key while redelivery of the same file stays stable.
- **Config validation** (index.ts) — `workSources` array-guarded; each entry needs a non-empty `id`/`dir`; non-positive/non-finite `visibilityTimeoutMs` rejected (both at the boundary and in `createFolderWorkSource`); duplicate `dir` rejected (enforces one live source per dir — closes the multi-instance-reclaim concern).
- **Spawn-window abort** (engine.ts) — `.catch()` attached so a rejected abort can't surface as an unhandled rejection after lease release; added a regression test.
- **E2E** — the fan-out poll now binds to this fire's `runId`.
- **Docs** — proposal / work-source spec / test-plan F1 now phrase distinctness as a *concurrently-valid-lease* invariant (at-least-once redelivery after expiry). AGENTS.md rows reordered path-alphabetical.

**Stale (already resolved by scrubbing unrelated artifacts)**
- `.pi/settings.json` and `.pi/skills/manage-flows/SKILL.md` were accidentally swept in by an early `git add -A`; both are unrelated session artifacts and have been removed from the branch entirely (history rewritten), so those three comments no longer apply.

molnar-botond:
Round 3 addressed in ab597a8fe:

- **engine.ts:775 (lease release on unresolved spec)** — the pre-spawn error path now `nack`s every leased handle before settling the parent errored; extended the same guard to the source-unregistered and `next()`-throw paths so no branch strands items until the visibility timeout.
- **index.ts:235 (isolate construction + log rejects)** — `createFolderWorkSource` construction (which does fs I/O via `ensureDirs`/`reclaim`) is now wrapped in try/catch so one unwritable dir can't abort engine init; every rejected config entry is logged with its id via `ctx.logger.warn`.
- **automation-fanout.spec.ts:159 (assert cleanup)** — the cleanup DELETE now parses its body and `expect.soft`-asserts `ok === true` (soft so a cleanup miss is reported without masking a real in-test failure).

The re-posted `.pi/settings.json` and `.pi/skills/manage-flows/SKILL.md` comments remain stale — those files are not in this branch's diff.

PR #598, opened 2026-09-02
feat(automation): add folder-scope contribution axis

molnar-botond:
Implements OpenSpec change `add-automation-folder-scope-contribution`.

## What
Adds a second dynamic contribution axis `automation.folderscope.` (mirroring the existing `automation.action.` axis). A plugin/host publishes `ctx.provide("automation.folderscope.<id>", { base })`; the plugin unions the validated, resolved bases into `folderScopeBases()`, which previously derived folder scopes **only** from live session cwds. This makes a repo with an enabled `automation.yaml` but **no live session** reachable to the engine — scanned, armed, reaped, and watched — without forking the scan→arm→board→toggle machinery. `listScopes()`/`refresh()`/`reapStaleRuns()`/`attachWatchers()` are unchanged.

## Boundary (execution-arming surface)
- Fail-open validation: accepts only a **plain** object (`Object.prototype`/`null` proto — not Date/Map/class/array/null) whose `base` is a non-empty-after-`trim()` string that `path.resolve` accepts. The whole per-entry read is wrapped so a hostile throwing getter/proxy is isolated and cannot abort collection.
- **Warn once per key** (persistent `warnedKeys` Set), never once per read.
- Dedup by resolved path; a base equal to the global home dir is dropped (the `global` scope owns it → no double-arm).
- Nav pins / `knownFolderCwds` are deliberately **not** consumed.
- Boot arm is one-shot; contributions are process-lifetime (host has no `unprovide`). Post-boot zero-session live-add is a documented non-goal.

## Tests
13 automated L1 vitest scenarios (E1–E7, X1–X2, I1–I3, S1) plus 2 hardening tests from local review (non-plain-object rejection; hostile-getter isolation). New `folder-scope-contributions.ts` + collector unit tests; engine/watcher/scanner integration tests.

## Notes for review
- Local independent review (codex gpt-5.6-luna, `review-code` rubric) ran two rounds; both blocking findings (plain-object gate, guarded property access) resolved.
- One pre-existing local test (`attach returns false for a missing automation dir`) is red only on Linux due to `fs.watch` leniency (defers to the async error event) — unrelated to this change; CI env is authoritative.

QA/manual tasks: none — all scenarios are automated.


molnar-botond:
@coderabbitai full review

molnar-botond:
Applied CodeRabbit feedback in d19b222fb:
- Reject non-absolute contributed `base` (spec requires an absolute repo root; a relative base would resolve against the server cwd) + relative-rejection test.
- Strengthened the watcher degrade test to assert the failed base is retried (`badAttachCalls === 2`).

Skipped (with reason): the note to repoint the collector test-file reference in `tasks.md` — that file is under `openspec/changes/archive/` (immutable history per repo conventions) and the reference is a non-functional pointer.

PR #601, opened 2026-09-04
feat(fix-default-model-clobbers-explicit-model): never clobber an explicit --model with the dashboard default

### LINKED ISSUE #595: [BUG FIX] pi-dashboard bridge overwrites subagent model with config default
# [BUG FIX] pi-dashboard bridge overwrites subagent model with config default

**Component:** `pi-agent-dashboard` (`packages/extension/src/bridge.ts`) and `pi-dashboard-extension` (`src/bridge.ts`)

**Type:** Bug fix (behavior parity with pi CLI)

**Status:** Ready for upstream review

---

## 1. Summary

When a pi subagent (spawned by the `pi-subagents` package) runs while the
`pi-agent-dashboard` bridge is loaded, the bridge **overwrites the subagent's
configured model with `config.defaultModel`** ~5 seconds after the subagent
process starts.

Observed on a real subagent session:

```
t+0s    model_change → xai/grok-4.6      (the agent's configured model, from --model)
t+5s    model_change → newapi/glm-5.3-flash   (config.defaultModel — WRONG)
```

The same subagent run under plain `pi` CLI (no dashboard bridge) keeps its
agent model — which is why this regression is dashboard-specific.

---

## 2. Environment

| Component | Version |
|---|---|
| `pi-agent-dashboard` | 0.8.0 |
| `pi-dashboard-extension` | 0.8.0 |
| `pi-subagents` | current (spawns children with `--model <agent-model>`) |
| `pi` | current (`pi-coding-agent`) |

---

## 3. Symptom

1. User configures an agent (e.g. `mini`) with `model: newapi/MiniMax-M3` in
   `~/.pi/agent/agents/mini.md`.
2. Parent session (running under pi-dashboard) calls
   `subagent({ agent: "mini", ... })`.
3. `pi-subagents` spawns a child `pi --mode json --model newapi/MiniMax-M3 ...`
   process. The child is a **brand-new session** (`messageCount === 0`,
   `reason === "startup"`).
4. ~5 seconds later, the dashboard bridge's `applyDefaultModel()` fires
   `pi.setModel(config.defaultModel)` → the child now runs `glm-5.3-flash`
   instead of `MiniMax-M3`.
5. The dashboard session card for the subagent shows the wrong model; the
   subagent's own model verification fails
   (`Expected 'newapi/MiniMax-M3' but observed 'glm-5.3-flash'`).

### Evidence from a failing run (`events.jsonl`)

```
message_start  model=glm-5.3-flash  provider=newapi   ← WRONG (default)
```

`pi-subagents` status record:

```json
"modelAttempts": [{
  "model": "newapi/MiniMax-M3",
  "success": false,
  "error": "model_verification_failed: child reported a different model than
            the launch candidate. Expected 'newapi/MiniMax-M3' but observed
            'glm-5.3-flash'."
}]
```

---

### 4.1 The default-model gate does not distinguish subagent children

`bridge.ts` applies the configured default model at `session_start` for any
**brand-new** session:

```ts
// (from packages/extension/src/bridge.ts, session_start handler)
const entryCount = ctx.sessionManager.buildSessionContext?.()?.messages?.length ?? 0;
const freshConfig = loadConfig();
if (shouldApplyDefaultModel({
  reason: _event?.reason,
  entryCount,
  hasModelRegistry: Boolean(cachedModelRegistry),
  hasDefaultModel: Boolean(freshConfig.defaultModel),
})) {
  pendingDefaultModel = applyDefaultModel();
}
```

`shouldApplyDefaultModel` (in `bridge-default-model-gate.ts`) returns `true`
when:

```ts
reason === "startup" && entryCount === 0
  && hasModelRegistry && hasDefaultModel
```

This is the correct intent for **human-opened** new sessions (fresh dashboard
session should start on the operator's default model). But a **subagent child
is also a brand-new session** (`reason="startup"`, `messageCount=0`), so the
gate fires for it too — even though `pi-subagents` already launched the child
with an explicit `--model <agent-model>` on the command line.

`pi.setModel(config.defaultModel)` then clobbers the child's agent model.

### 4.2 Why pi CLI is unaffected

`pi` CLI does not load the `pi-agent-dashboard` extension, so the bridge
(and `applyDefaultModel`) never runs there. The subagent keeps its
`--model` arg. Hence: **CLI = correct; dashboard = wrong.**

### 4.3 There are two copies of the bridge source

The bridge is shipped in **two packages** with identical `bridge.ts` logic:

1. `@blackbelt-technology/pi-dashboard-extension/src/bridge.ts`
   (standalone package)
2. `@blackbelt-technology/pi-agent-dashboard/packages/extension/src/bridge.ts`
   (monorepo-internal copy)

Subagent child processes actually load copy #2 (confirmed via
`/tmp/jiti/src-bridge.*.mjs` cache provenance: the cache compiled for
`pi-agent-dashboard/packages/extension/src/bridge.ts`). **A fix must be
applied to both copies** or the regression persists for whichever copy a
child loads.

---

### 5.1 Detection signal (already provided by `pi-subagents`)

`pi-subagents` sets `PI_SUBAGENT_CHILD=1` in every child's environment
(`buildPiArgs` → `env[SUBAGENT_CHILD_ENV] = "1"`, `pi-args.ts:850`). The
bridge runs inside the child process, so `process.env.PI_SUBAGENT_CHILD`
is available at `session_start`.

### 5.2 Change 1 — guard the session_start default-model gate

In **both** `bridge.ts` copies, `packages/extension/src/bridge.ts` and
`pi-dashboard-extension/src/bridge.ts`:

```diff
     const entryCount = ctx.sessionManager.buildSessionContext?.()?.messages?.length ?? 0;
     const freshConfig = loadConfig();
-    if (shouldApplyDefaultModel({
+    const isSubagentChild = process.env.PI_SUBAGENT_CHILD === "1";
+    if (!isSubagentChild && shouldApplyDefaultModel({
       reason: _event?.reason,
       entryCount,
       hasModelRegistry: Boolean(cachedModelRegistry),
       hasDefaultModel: Boolean(freshConfig.defaultModel),
     })) {
       pendingDefaultModel = applyDefaultModel();
     }
```

### 5.3 Change 2 — guard the deferred retry path (defense-in-depth)

The same override can re-fire later via `onProviderChanged` → `applyDefaultModel`
retry when `pendingDefaultModel` is set. Guard that path too:

```diff
-      if (pendingDefaultModel) {
+      if (pendingDefaultModel && process.env.PI_SUBAGENT_CHILD !== "1") {
         pendingDefaultModel = applyDefaultModel();
       }
```

### 5.4 Rationale

- The subagent's model is **already authoritative** — `pi-subagents` resolved
  it from the agent definition (`agents/*.md` frontmatter) and passed it as
  `--model`. The dashboard has nothing to add for subagent children.
- Skipping default-model application for `PI_SUBAGENT_CHILD=1` makes the
  dashboard behave exactly like `pi` CLI.
- No behavior change for human-opened sessions, resume/fork/reload
  (those already bypass the gate via `entryCount > 0` / `reason !== "startup"`).

---

## 6. Verification

After applying the fix to **both** bridge copies:

### 6.1 Runtime check (real subagent)

Run a subagent (e.g. agent `mini` with `model: newapi/MiniMax-M3`) with a
sleep task, then inspect the child's event stream:

```bash
# before fix
grep -o '"model":"[^"]*"' /tmp/pi-subagents-uid-1000/async-subagent-runs/<run>/events.jsonl
# → "model":"glm-5.3-flash"   (WRONG — default overwrote agent model)

# after fix
# → "model":"MiniMax-M3"      (CORRECT — agent model preserved)
```

### 6.2 Regression surface (must remain unchanged)

| Scenario | Expected model |
|---|---|
| Fresh human-opened session via dashboard | `config.defaultModel` |
| Resume / fork / reload (`entryCount > 0`) | existing session model |
| Subagent child (`PI_SUBAGENT_CHILD=1`) | agent's `--model` (from `agents/*.md`) |
| pi CLI (no bridge loaded) | agent's `--model` |

---

## 7. Suggested upstream follow-ups

1. **Single source of truth for the bridge.** The duplicate
   `bridge.ts` in two packages is fragile — consider publishing
   `pi-dashboard-extension` and having `pi-agent-dashboard` re-export it
   (it already declares `@blackbelt-technology/pi-dashboard-extension` as a
   dependency), or a shared workspace package.
2. **Document the `PI_SUBAGENT_CHILD` contract** (set by `pi-subagents`
   `buildPiArgs` for every child) so other extensions can rely on it.
3. Optionally: instead of the env check, `pi-subagents` could pass a
   `--model`-authoritative flag through `session_register` (e.g.
   `spawnModelExplicit: true`) — but the env check is simpler and works
   without protocol changes.

---

## 8. Files touched

| File | Change |
|---|---|
| `pi-agent-dashboard/packages/extension/src/bridge.ts` | 2 guards (session_start gate + retry) |
| `pi-dashboard-extension/src/bridge.ts` | 2 guards (same, duplicate copy) |

(No changes to `pi-subagents` are required for this fix — it already sets
`PI_SUBAGENT_CHILD=1` and passes `--model` correctly.)

robertcsakany:
Implements OpenSpec change `fix-default-model-clobbers-explicit-model` (issue #595).

## What
The bridge applied `config.defaultModel` to every brand-new startup session — including sessions launched with an explicit `--model` on the pi argv (~5s deferred clobber). The default-model gate now carries a required `hasExplicitModel` input; `bridge.ts` derives it from `process.argv` via the new pure `hasExplicitModelArg()` (exact-token match, mirroring pi's own `parseArgs` — `cli/args.js:55`). An explicit model now dominates every other gate condition; `defaultThinkingLevel` is intentionally skipped with it (CLI parity). The provider-ready retry needs no separate guard: the gated entry never sets `pendingDefaultModel` for explicit-model sessions.

## Tests
- L1: gate decision-table (E1–E5), apply-mirror with argv injection + pending signal (X1 + control), thinking-level coupling (X2) — 45 tests across the three suites.
- L3 (docker harness): new `tests/e2e/explicit-model-preserved.spec.ts` — I1 launches a detached child pi process in-container with `--model faux/faux-2` and asserts the model stays put across the ≥8s deferred-apply window; I2 proves a plain REST-spawned session still converges to the seeded default `faux/faux-1`. `faux/faux-2` registered in the faux fixture for a non-default explicit target.
- Harness-verified green on a rebuilt image; knip ratchet green on all classes.

## Also in this PR (separate commit)
`chore(knip)`: retires 19 verified-unused exports/types + one duplicate that kept the plain knip-ratchet red on develop (+10/+2/+1 over baseline). Zero external references per knip + repo-wide rg; tsc/build/vitest verified.

## Known pre-existing red
`skill-frontmatter.test.mjs` (video-transcription description budget) fails identically on origin/develop and is why develop's own CI is red (run 33822271219). Not touched by this PR per maintainer decision.

QA/manual tasks (real pi-subagents child check #4.1, runtime regression sweep #4.2) deferred to post-merge verification.

Closes #595


robertcsakany:
@coderabbitai full review

PR #602, opened 2026-09-05
test: make the suite deterministic — fixed-tick guard + waitFor conversions + single worker source

robertcsakany:
## Summary

Implements `make-test-suite-deterministic` (scope-amended — see `SHIP_IT_BLOCKED.md` in the change dir for the approved defect-A drop; already fixed upstream by 71ea6e593).

- **Fixed-tick barriers → waitFor:** the 10 client test files that awaited a bare-resolve `setTimeout` as an assertion barrier now poll (`waitFor` / mock-state) — the rotating local-red class. `PairLanding`'s mock-internal yield is kept and annotated as the guard's per-occurrence opt-out exemplar.
- **Machine enforcement:** `scripts/check-fixed-tick-waits.mjs` (client census, per-occurrence opt-out, stderr naming) + CI step + vitest wrapper (`scripts/__tests__/fixed-tick-waits.test.mjs`, test-plan E10–E14/P3/X5). 483 client test files, 0 violations.
- **maxWorkers single source of truth:** repo-root `vitest.workers.ts` (`PARALLEL_MAX_WORKERS = "50%"`) imported by all 27 parallel vitest configs by relative path (no dependency edge); the 7 serial projects keep explicit `1`. Census guard `scripts/__tests__/vitest-workers.test.mjs` (E15–E17/P1). No effective worker count changed anywhere.
- **F2 behaviour-preservation census:** `fixed-tick-conversion-equivalence.test.ts` pins the 10 converted files' test ids to the merge base — the conversion cannot silently add/drop/rename tests (it correctly caught #597's 11 incoming PathPicker tests during the develop merge).

## Scope amendment & review

- Planning artifacts amended with author approval after filesystem-reality checks invalidated defect A (evidence: `SHIP_IT_BLOCKED.md`).
- Isolated `@review` gate ran with the review-code rubric; blocking finding (mutation-harness residue visible mid-run) resolved by process discipline; minors applied (guard regex widening: two-param/function-form/multiline/sleep-helper; P1 census tightening).
- Keeper E5 one-shot read race absorbed by approved decision (poll-for-child-bytes before purity assert).

## Test evidence

- Change-scoped suites green in **all 8 full-suite runs** on a loaded dev box; standalone guard: 483 files / 0 violations; enforcers green (conventions, dox-byte-gate, i18n ×2, knip-config, knip baseline-diff). Knip ratchet is red +10/+2/+1 **byte-identically on origin/develop** (pre-existing drift, nightly-owned; baseline untouched, ceiling not raised).
- P2 (3-run soak) environment-limited by the same pre-existing rotating population; follow-up filed: `contention-harden-real-process-tests` (documented in the change's design notes + tasks).

See change: make-test-suite-deterministic


robertcsakany:
Review fixes applied (84f6→see push):

1. **keeper label** — design.md now labels keeper E5 as baseline-only casualty (fixed mid-verification, green since).
2. **`.ts`→`.mjs`** — corrected in test-plan + tasks.
3. **E15 omission** — partially applied with evidence: vitest's unset `maxWorkers` default is `numCpus-1` (verified in vitest dist: `getDefaultThreadsCount`), NOT 50% — so adopting the module in `dashboard-plugin-skill` would change its effective worker count (15→8), violating design D3's no-value-change mandate. Documented as an explicit exception in E15; full adoption is a decision for the follow-up.
4. **F2 scope** — reworded to test-id/count preservation only; assertion preservation remains a diff-review criterion (no-weakening guard + @review gate enforced it on this diff).
5. **PathPicker** — both refetch sites now flush the resolved browseDirectory result into component state (`await act(async () => {}`) after the call-count poll, before the Enter/Create-row assertions.

PR #610, opened 2026-09-05
feat(config): add configurable readinessTimeoutMs for bridge cold-start readiness

pikujs:
## Why

The bridge auto-spawn gives the dashboard server a fixed **10 s** cold-start readiness window, hardcoded at the `launchServer` call site (`packages/extension/src/server-launcher.ts` → `healthTimeoutMs: 10_000`). On slow hosts the server's real cold start — dominated by the startup session scan, which grows with the number of sessions under `~/.pi/agent/sessions` — can outlive that window.

Measured on a machine with 229 sessions: **~16.5 s** from spawn to `writePid()`, plus more for health-OK. When the window expired, the bridge reported:

```
Warning: Dashboard server failed to start: readiness timeout
See log: ~/.pi/dashboard/server.log
```

…while the server went on to boot healthily in the background (HTTP 200 seconds later). The warning is misleading — nothing failed — and unactionable, since the window is not configurable; the only user-side remedy is patching `node_modules`. This is the same scenario `fix-bridge-server-start-diagnostics` anticipated ("slow hosts reach `writePid()` but are not health-OK within 2 s"), just at a larger margin.

## What Changes

- **`packages/shared/src/config.ts`** — new `readinessTimeoutMs` field on `DashboardConfig`: a positive finite number is honored; anything else falls back to the 10 s default (the historical hardcoded value, so existing configs see **zero behavior change**). Included in `ensureConfig`'s written defaults for discoverability. Doc comment spells out that the timeout is a readiness *budget*, not a boot deadline: expiry only controls when the bridge surfaces a warning; the spawned server keeps booting regardless.
- **`packages/extension/src/server-launcher.ts`** — `launchServer` forwards `config.readinessTimeoutMs` (10 s fallback for legacy config objects) as `healthTimeoutMs` instead of the hardcoded constant.

**Not in scope:** the standalone CLI path (`packages/server/src/cli.ts`, hardcoded 30 s) — it already tolerates slow starts, and this change's scope is the bridge path that produced the observed failure. It can adopt the same field later if desired.

OpenSpec proposal: `openspec/changes/add-configurable-readiness-timeout/proposal.md` (modifies `shared-config`, `bridge-auto-start-lifecycle`).

## Testing

- `packages/shared/src/__tests__/config.test.ts` — new case: defaults to 10 000; positive values round-trip; `0` / negative / non-numeric fall back to the default.
- `packages/extension/src/__tests__/server-launcher-launch.test.ts` — forwarding pin updated: default 10 s when the field is absent; configured value (60 000) wins when present.
- Both touched test files: **123/123 passing**. Repo-wide `pnpm test`: 18 011 passed; the 5 failures in 4 files reproduce identically on a clean `develop` checkout (real-`fs.watch`/docker/timer-dependent environment tests) and are unrelated to this change. `tsc --noEmit` clean.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -2739,6 +2739,7 @@ Precedence: CLI flags → environment variables → config file (`~/.pi/dashboar
 | `autoStart` | true | Bridge extension auto-starts server if not running |
 | `autoShutdown` | false | Server shuts down after idle period (disabled by default; enable for TUI auto-start scenarios) |
 | `shutdownIdleSeconds` | 300 | Idle timeout before auto-shutdown |
+| `readinessTimeoutMs` | 10000 | Bridge auto-spawn cold-start readiness budget (ms). Expiry surfaces "readiness timeout" warning only; spawned server keeps booting. Raise on slow hosts (large session scan). Positive number clamped to [1000, 600000]; anything else falls back to default. Auto-start lock staleness bound derived from it: `spawnReadinessBudgetMs` = max(3x value, 30000); raising it cannot break the single-flight lock |
 | `spawnStrategy` | `"headless"` | How to spawn new sessions: `"headless"` or `"tmux"` |
 | `tunnel.enabled` | true | Enable Gateway (internal id `tunnel`) for remote access |
 | `tunnel.provider` | — | `"zrok"`\|`"ngrok"`\|`"tailscale"`\|`"zerotier"`. Required when enabled |
--- a/openspec/changes/add-configurable-readiness-timeout/.openspec.yaml
+++ b/openspec/changes/add-configurable-readiness-timeout/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-05
--- a/openspec/changes/add-configurable-readiness-timeout/proposal.md
+++ b/openspec/changes/add-configurable-readiness-timeout/proposal.md
@@ -1,0 +1,30 @@
+## Why
+
+The bridge's auto-spawn gives the dashboard server a fixed 10 s cold-start readiness window, hardcoded at the `launchServer` call site (`packages/extension/src/server-launcher.ts`, `healthTimeoutMs: 10_000`). On slow hosts the server's real cold start can outlive that window — the dominant cost is the startup session scan, which grows with the number of sessions under `~/.pi/agent/sessions`. A measured cold start with 229 sessions took ~16.5 s from spawn to `writePid()`, plus more for health-OK.
+
+When the window expires, the bridge reports `Dashboard server failed to start: readiness timeout` while the server goes on to boot healthily in the background (HTTP 200 seconds later). The warning is misleading — nothing failed — and unactionable: the window is not configurable, so affected users' only remedy is patching `node_modules`. This mirrors the situation the `fix-bridge-server-start-diagnostics` change already anticipated ("slow hosts reach `writePid()` but are not health-OK within 2 s"), just at a larger margin.
+
+## What Changes
+
+- Add `readinessTimeoutMs` to `DashboardConfig` (`packages/shared/src/config.ts`), parsed from `~/.pi/dashboard/config.json` like the sibling numeric fields: a positive finite number is honored, anything else falls back to the 10 s default (the historical hardcoded value, so existing configs see zero behavior change). Included in `ensureConfig`'s written defaults for discoverability.
+- `launchServer` (`packages/extension/src/server-launcher.ts`) forwards `config.readinessTimeoutMs` (with a 10 s fallback for legacy config objects) as `healthTimeoutMs` to the shared `launchDashboardServer` primitive.
+- The timeout is a readiness **budget**, not a boot deadline: expiry only controls how long the bridge waits before surfacing the warning. The spawned server keeps booting either way, so a value below the real cold start produces a spurious error next to a healthy server — the doc comment on the new field says so explicitly.
+
+**Not in scope:** the standalone CLI path (`packages/server/src/cli.ts`, hardcoded 30 s) — it already tolerates slow starts, and this change's scope is the bridge path that produced the observed failure. It can adopt the same field later if wanted.
+
+## Capabilities
+
+### New Capabilities
+
+_(none — this modifies an existing capability)_
+
+### Modified Capabilities
+
+- `shared-config`: schema gains `readinessTimeoutMs` (number, default `10000`); non-positive or non-numeric values fall back to the default.
+- `bridge-auto-start-lifecycle`: the spawn readiness window is taken from `readinessTimeoutMs` instead of a hardcoded constant; semantics of expiry (warning + background boot continues) are unchanged.
+
+## Impact
+
+- **Code**: `packages/shared/src/config.ts` (field + default + parse + `ensureConfig`), `packages/extension/src/server-launcher.ts` (forwarding + doc comment).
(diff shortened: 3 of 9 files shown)


pikujs:
Note for reviewers: this is a first-time fork PR, so the `CI` workflow run is currently in `action_required` state pending workflow approval — it needs a maintainer to click "Approve and run" before checks execute.

Local verification done in lieu of that run: touched test files 123/123 green; repo-wide `pnpm test` 18 011 passed with only environment-dependent failures (real-`fs.watch` / docker / timer tests) that reproduce identically on a clean `develop` checkout; `tsc --noEmit` clean; `pnpm spec:validate` 573/573; `biome check --changed` reports only pre-existing findings present on `develop`.

robertcsakany (changes requested):
## Advisory review (local deep-audit, files outside the diff included)

First: this is a well-built change — the parse mirrors `subagentTickThrottleMs` exactly, the doc comment honestly distinguishes readiness *budget* from boot deadline, both forwarding paths are pinned in tests, and default behavior is genuinely zero-change. The finding below only surfaced by auditing the lock machinery *around* the diff.

### issue(blocking): raising the window past 30 s breaks the auto-start lock invariant

`packages/extension/src/server-launcher.ts` (forwarding) ↔ `packages/shared/src/config.ts:669` ↔ `packages/extension/src/autostart-lock.ts:97,118`

`SPAWN_READINESS_BUDGET_MS` (= `HEALTH_CHECK_TIMEOUT_MS × 3` = 30 s) bounds auto-start **lock staleness**, and is documented as "deliberately LARGER than the health poll". This PR makes the health poll configurable but leaves the budget constant — so the documented invariant (budget > poll) silently inverts for any configured value > 30 s.

Concretely: the lock is `childPid`-less for the entire readiness window (`server-auto-start.ts` records `childPid` only on readiness success — the "Accepted gap" comment), so `isLockStale` falls through to pure age: `now - startedAt > 30_000`. With `readinessTimeoutMs: 60000` (this PR's own test value), at t=30 s a second pi session declares the lock stale, breaks it, and spawns a **competing server** → `PortConflictError`. That converts today's cosmetic warning into a dual-spawn failure — for exactly the slow hosts this change targets.

**Suggested fix (no new config needed):** derive the effective budget from the configured timeout and thread it through the seam that already exists (`server-auto-start.ts` already takes `deps.readinessBudgetMs ?? SPAWN_READINESS_BUDGET_MS`):

```ts
const effectiveBudgetMs = Math.max(3 * readinessTimeoutMs, SPAWN_READINESS_BUDGET_MS);
```

and pass it as `readinessBudgetMs` on the auto-start path (optionally also scale `SERVER_STARTUP_DEADLINE_MS`, which is defined as `budget × 4` for the same anti-drift reason). Alternative: clamp `readinessTimeoutMs ≤ SPAWN_READINESS_BUDGET_MS` in `loadConfig`, but that caps the fix at 30 s, which may not cover the ~16.5 s + health-OK cold start described in the proposal on even slower hosts.

### issue(non-blocking): no upper clamp ⇒ unbounded readiness poll

`readinessTimeoutMs: 1e21` makes `deadline = now() + v === v` in `shared/src/server-launcher.ts:302` — the 300 ms poll and the bridge's launch spinner run for the session's lifetime and `onLaunchEnd` never fires. The doc comment advertises "no upper clamp"; suggest clamping in `loadConfig` (e.g. 10 min).

### issue(non-blocking): `10_000` magic number now exists in three places

`HEALTH_CHECK_TIMEOUT_MS` (config.ts:668) is the existing single source of truth; the diff hardcodes `10_000` in `DEFAULTS`, and in the `??` fallback in `server-launcher.ts`. Reference the constant in both.

### suggestion

- `proposal.md` is missing the `## Discipline Skills` section required by this repo's OpenSpec conventions (`scripts/check-conventions.mjs` Rule 4) — even "none apply" must be stated under the heading.
- The field appears in none of the four config-reference tables (`docs/architecture.md`, `README.md`, `docs/faq.md`, `docs/features.md`) — given the field's purpose is discoverability for slow-host operators, worth adding at least to `docs/architecture.md`.
- Parse accepts any positive finite number: `0.5` reproduces the spurious-timeout failure mode instantly. Consider a sane floor (≥ 1000 ms) and pinning `Infinity`/`NaN`/`null` in the config test.

robertcsakany:
Addressed in 36d8786 — CI green (28m31s), CodeRabbit clean.

**Blocking — raised readiness window inverts the lock invariant**
Fixed as suggested, via the existing `deps.readinessBudgetMs` seam. New exported helper in `packages/shared/src/config.ts`:

```ts
export function spawnReadinessBudgetMs(readinessTimeoutMs?: number): number {
  const poll = /* positive + finite, else HEALTH_CHECK_TIMEOUT_MS */;
  return Math.max(poll * 3, SPAWN_READINESS_BUDGET_MS);
}
```

`server-auto-start.ts` now does `deps.readinessBudgetMs ?? spawnReadinessBudgetMs(config.readinessTimeoutMs)`, and `autoStartServer`'s `config` param gained optional `readinessTimeoutMs`. The ×3 ratio is the existing one, so `budget > poll` holds for **every** configured value; with no config the result is byte-identical to the old constant (30 s), so no existing test moved.

Regression test in `server-auto-start-guarded.test.ts` — a childPid-less lock held 40 s ago by an alive session:
- under `readinessTimeoutMs: 60_000` (budget 180 s) the lock stays live → the second session takes the loser path, `launchServer` **not** called;
- under the default window (budget 30 s) the same lock still expires → spawn proceeds.

Verified red: reverting the budget line to the constant fails exactly the first case and nothing else.

`SERVER_STARTUP_DEADLINE_MS` deliberately stays constant, now stated in the proposal's not-in-scope: it bounds a server killing *itself* on a hung boot, not a spawner waiting, so a raised bridge window cannot turn it into a false positive.

**Non-blocking 1 — no upper clamp**
Clamped into `[READINESS_TIMEOUT_MIN_MS = 1_000, READINESS_TIMEOUT_MAX_MS = 600_000]` (10 min, as suggested). This also covers the parse-accepts-`0.5` suggestion — the floor is the same clamp. Non-finite input is unchanged: `Infinity`/`NaN`/`null` still fall back to the default (all three pinned, plus the raw `1e999` literal a hand-edited config can carry, since JSON has no `Infinity` token).

**Non-blocking 2 — `10_000` in three places**
`DEFAULTS.readinessTimeoutMs` and the `launchServer` `??` fallback now both reference `HEALTH_CHECK_TIMEOUT_MS`; the literal is gone from both.

**Suggestions**
- `proposal.md` gained `## Discipline Skills` (`doubt-driven-review`, `review-code`) — `node scripts/check-conventions.mjs` reports no violations. What-changes / not-in-scope / impact sections updated for the derived budget too.
- `docs/architecture.md` config reference gained the `readinessTimeoutMs` row (default, clamp range, expiry semantics, and the derived lock bound).

**Local verification:** full suite 18 007 passed. The 5 failures are the pre-existing environment-dependent ones (pi version pin, kb eval-guard, published-imports, bus-client type-negative fixture) that fail identically on an untouched checkout. `biome --changed` surfaces only diagnostics already present on the previous head.

One caveat for anyone re-running locally in a git worktree: `tsc --noEmit` resolves `@blackbelt-technology/pi-dashboard-shared` through the **main** checkout's `node_modules` symlink, so it cannot see `readinessTimeoutMs` at all — including the field this PR's first commit added. CI's fresh install is the authoritative typecheck.

PR #612, opened 2026-09-07
fix(kb): share KB stats per folder + card-placement reindex control

robertcsakany:
Implements OpenSpec change `fix-kb-card-refresh-and-shared-stats`.

## Why

`useKbStats(cwd)` held its whole fetch/poll/pending/error state machine **per hook instance**. `FolderKbSection` and `KbSettingsPanel` are two independent consumers, so a reindex triggered from the settings dialog was invisible to the sidebar/worktree-card KB row until a page reload. Separately, the worktree-card placement lost its reindex affordance when slot actions folded into the folder-actions menu — a menu that scope does not have.

## What

**1. Shared per-cwd stats store** (`packages/kb-plugin/src/client/kb-stats-store.ts`, new)

Module-level `Map<cwd, KbStatsStore>`; `useKbStats` becomes a thin `useSyncExternalStore` subscription with the public `UseKbStatsResult` shape unchanged. `POLL_MS`/`MAX_POLL_MISSES`/`REINDEX_GUARD_MS` semantics are preserved, just hoisted to folder scope:

- first subscriber → initial fetch; later subscriber → retained snapshot synchronously + background revalidate **coalesced** with any in-flight fetch;
- last unsubscribe → poll stopped, in-flight fetch **not** aborted, guard left armed and self-clearing `pending` **without** fetching at zero subscribers;
- epoch-ordered fetches — a stale response can never overwrite newer state;
- `loading` toggles on initial/revalidate/refetch epochs only, never on poll ticks;
- snapshot object identity replaced only on a real field change (`useSyncExternalStore` re-render-loop guard).

Deliberate semantic change, spec'd in the delta: `error`/`reindexError` become **shared folder state** — a settings-panel trigger reject now also flips the sidebar/card row to the error state (Retry is one click away there).

**2. Card-placement sibling reindex control** (`FolderKbSection.tsx`)

One compact `folder-kb-card-reindex` icon-button rendered as a **sibling** of `SlotPill`, outside the pill's `role="button"` root (ARIA nesting + the pill's Enter/Space `preventDefault` keyboard trap). State-varying accessible name (Retry / indexing… / Index now / Reindex now), label kept perceivable while disabled via a wrapper `title`, disabled for the whole busy window. Sidebar placement is untouched.

## Tests

- 16 new L1 tests (12 store/hook + 3 card control + 1 miss-run parity) covering the manifest's E1–E6, F1/F3/F4/F5/F7, X1–X4.
- 1 new L3 Playwright scenario (F2, the reported defect): a reindex from the settings panel updates the sidebar row **without a reload**, with the corpus grown out-of-band so the count genuinely changes.
- `resetKbStatsStores()` added to `beforeEach` in every kb-plugin client test file (module singleton).
- Biome warnings in `packages/kb-plugin/src/client` drop 7 → 2.

## Known-unrelated local failures

`scripts/__tests__/knip-harness.test.mjs` (2 tests) fails on this machine **before** this change too — `vitest.workers.ts` is not `COPY`d into the docker image, so knip inside the container cannot load the per-package `vitest.config.ts` files and reports 2 extra unused files. Pre-existing gap from #602; verified by re-running the spec with this change stashed.

QA/manual task 6.1 (visual fit of the card control across themes) deferred to post-merge verification.


robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

robertcsakany:
@coderabbitai full review

PR #613, opened 2026-09-07
feat(asciidoc-support): first-class AsciiDoc in the kb + styled preview

robertcsakany:
Implements OpenSpec change `asciidoc-support`.

## What lands

- **kb AsciiDoc chunker** (`packages/kb/src/adoc-chunker.ts`) — delimited-block-safe section splitting across all 9 block types, document-attribute header extraction, real 1-based line anchors, `xref:`/`<<>>` link extraction feeding the existing file-level graph edges. Total (never throws) and pure (CRLF-normalized). Shares thresholds + id/hash scheme with the markdown chunker for chunk-contract parity, but has its own oversize splitter so a split point can never land inside a delimited block.
- **Indexing pipeline** — all three selection gates widened together (walk regex, `config.ts` `include`/`extensions`, per-extension chunker dispatch). `SCHEMA_VERSION` 2 -> 3 adds `start_line`/`end_line` FTS5 columns after `body` so bm25/snippet column indices stay valid.
- **kb-extension** — Job 1 (reindex) and Job 2 (DOX nudge) predicates decoupled: an `.adoc` edit now runs both; markdown behaviour is unchanged.
- **Preview styling** — new `.asciidoc-body` stylesheet using theme CSS custom properties only, replacing the dead `prose prose-invert` classes on the AsciiDoc and docx html-mode wrappers.

## Migration note

`CREATE VIRTUAL TABLE IF NOT EXISTS` is a no-op on an existing table and FTS5 has no `ALTER TABLE ADD COLUMN`, so a version bump alone would leave pre-existing stores with a 10-column `chunks` table and every insert throwing. `SqliteFtsStore.init()` therefore detects the old column set and rebuilds, clearing `files` so the next incremental walk repopulates. This covers openers with no version gate (the kb-extension reindex path).

## Verification

- 18,138 unit tests pass on the integrated tree; `npm run build` succeeds.
- New Playwright spec `tests/e2e/asciidoc-preview.spec.ts` (test-plan F1-F6) passes against the docker harness with computed-style assertions on heading hierarchy, admonition card, `frame=none`/`grid=none` and `stripes=even` tables, the `#toc` panel, and live theme retargeting.
- All repo enforcers green: `check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity`, `knip-config`, `knip-ratchet` (+ baseline-diff).

> Note: this branch also carries the pre-existing `plan(diagram-rendering)` commit (planning artifacts only, no code) that was committed here before implementation started.


robertcsakany:
Thanks — addressed both findings on this change's code in 0b5c8d197.

**`packages/kb/src/adoc-chunker.ts` — attribute-name canonicalization: applied.**
`parseHeader` now lowercases the attribute NAME (`attributes[a[2].toLowerCase()]`), so `:Tags:` and `:URL-Repo:` reach the indexer's `tags` / `url-repo` keys. Values keep their case, matching asciidoctor. Covered by a new unit test asserting `:Tags: API Docs` → `attributes.tags === "API Docs"`, `attributes.Tags === undefined`, and that it survives into `frontmatter`.

**The `has_tag` array half: deliberately deferred, not overlooked.**
`design.md` D3b for this change states: *"No attribute→tag mapping in v1 (AsciiDoc has no conventional tags attribute; deferrable without spec change)."* Splitting a scalar `:tags: a, b` into an array would add a behaviour the delta specs do not describe, so it stays out of this change by design. With the name now canonicalized, adding it later is a one-line change in the indexer rather than a chunker change.

**`tests/e2e/asciidoc-preview.spec.ts` — no-reload proof: applied.**
Agreed that `toBeVisible()` cannot distinguish an in-place retarget from a reload rendering the same preview. The spec now writes `window.__adocNoReload` before opening the ThemePicker and asserts it still holds a value after the theme flips. Re-verified green against the docker harness (both specs pass).

**The four threads under `openspec/changes/diagram-rendering/**` are out of scope for this PR.**
Those are planning artifacts for a *different, unimplemented* OpenSpec change whose commit predates this branch's work; they carry no code. Editing another change's proposal/design/specs from this PR would silently rewrite its planning decisions. They should be handled when `diagram-rendering` is implemented — the Kroki renderer whitelist, the upstream response-size bound, and the `saveConfig()` round-trip invariant all look like real points to settle there.

PR #616, opened 2026-09-08
plan(detach-automation-goal-from-core): request for validation — 8 open questions

molnar-botond:
## Request for validation, not an implementation

This PR adds **only OpenSpec documents** — no code, no spec deltas. It records a verified problem statement and ends in **eight open questions** that need a second opinion before any code moves. Please answer the questions rather than reviewing an implementation.

`openspec validate detach-automation-goal-from-core` → **valid** (`skip_specs: true`, per the `investigate-*` precedent — the scope is itself Q1).

## The finding

The dashboard already proved a decoupling doctrine — **publish/collect IoC** over `provide`/`consume`/`consumeAll` (`archive/2026-07-01-decouple-automation-action-registry`). **Session ownership is the one axis that never adopted it.**

| | `automation-plugin` | `goal-plugin` |
|---|---|---|
| publish/collect axes used | 3 (`action.`, `folderscope.`, `worksource.`) | **0** |
| product code in core | 1 registry + 1 event arm | **10 files** under `server/src/goal/` |
| server entry | substantial | 138-line shell |

Core also spells plugin words directly — `types.ts:425/435/442`, `session-meta.ts:128/153/160`, and on the **generic plugin API** at `server-context.ts:144` — and makes **lifecycle decisions** from them (`session-meta.ts:211` respawn, `pi-gateway.ts:898` finalize).

Proposed direction: core carries an **opaque `pluginRef`** keyed by the spawn token it already mints; plugins publish their own payload. **The same keys are still emitted** — `.meta.json` stays byte-identical, so there is no migration and no back-compat shim.

## Correction to the record

An earlier framing claimed the automation token-claim fix (`892e5d9d1`) was *deleted by a merge* and its tests were *dead*. **Both are false**, verified:

- `git log -S'consume(cwd, spawnToken)' origin/private/invoicebot` → exactly one commit, the one that **added** it. No deletion commit exists.
- It is alive and wired today: `event-wiring.ts:434`.
- Its tests assert the real property (`"never hands a token-bound stamp to a foreign or tokenless session"`).
- The apparent revert (`ca03b8003`/`d2a2dcb46`, 21s apart) touched only doc artifacts.

Actual situation: **branch divergence** — `develop` still has `consume(cwd)`; `private/invoicebot` has the fix. 114 ahead / 28 behind. Expected steady state under the merge doctrine, not a bug.

## The eight questions

1. **Q1 Scope** — seam only / seam + full goal move / two sequenced changes?
2. **Q2 Child participation** — host-side only, or an env-var ref the child echoes? (the latter must not be scrubbed → re-opens the `fix-spawn-token-env-leak` class)
3. **Q3 Continuity** — keeper respawn wants the ref to propagate; in-process fork wants it *not* to. Plugin policy or blanket core rule?
4. **Q4 cwd-FIFO** — may a cwd match ever assign ownership, or is it classification-only?
5. **Q5 Ordering** — file `token → ref` *before* the spawn await? (also: automation has no spawn-failure rollback where goal does — second latent bug?)
6. **Q6 Lifecycle policy** — a ref slot does not remove `isRecoveryCandidate`/`pi-gateway:898`. Declarative block or hooks?
7. **Q7 Undeletable wiring** — is a same-cwd collision test the right guard?
8. **Q8 Upstream** — is `private/invoicebot` the permanent home, or should this be upstreamable to `develop`?

## Review aids

`design.md` is an **evidence appendix** — every claim has a file:line anchor, plus the full register-path matrix (which `session_register` senders can carry a token) and the exact git commands behind the correction, so nothing needs re-deriving.

### DIFF
--- a/openspec/changes/detach-automation-goal-from-core/.openspec.yaml
+++ b/openspec/changes/detach-automation-goal-from-core/.openspec.yaml
@@ -1,0 +1,3 @@
+schema: spec-driven
+created: 2026-09-08
+skip_specs: true
--- a/openspec/changes/detach-automation-goal-from-core/design.md
+++ b/openspec/changes/detach-automation-goal-from-core/design.md
@@ -1,0 +1,144 @@
+# Design — evidence appendix
+
+> This change is a **request for validation**, not an implementation plan. This
+> document exists so a reviewer can check every claim in `proposal.md` against
+> the tree without re-deriving it. All line numbers are against
+> `private/invoicebot` unless a comparison to `develop` is stated.
+
+## The doctrine being applied
+
+Source: `openspec/changes/archive/2026-07-01-decouple-automation-action-registry/proposal.md`
+
+Distilled into six rules:
+
+1. **Publish/collect, not push-into.** `provide("<owner>.<axis>.<id>", frozen
+   contribution)`; the owner reads with `consumeAll("<owner>.<axis>.")`.
+2. **Owner owns the mechanism; contributor owns the payload.**
+3. **Collect lazily at read time** → load-order independent.
+4. **Built-ins are peers** — the owner self-publishes its defaults through the
+   same door (`ctx.provide(CORE_ACTION_KEY, coreActionContributions())`).
+5. **Neither side references the other** — no `dependsOn`, no imports.
+6. **Boundary-validate each contribution, fail-open** — plain-object check,
+   per-entry try/catch, warn once per key (see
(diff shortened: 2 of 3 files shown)


molnar-botond:
Closing: the private/invoicebot fork is retired. The invoice-bot plugin now runs detached on stock develop; this plan is superseded. The branch is backed up as a git bundle.

PR #619, opened 2026-09-09
fix(openspec-board): gate the board worktree button on folder availability, not liveness

robertcsakany:
Implements OpenSpec change `fix-openspec-board-worktree-button-gating`.

## Why

The OpenSpec board's per-card **New worktree** action was gated on
`sessions.some(s => s.cwd === boardCwd && !!s.gitBranch)`. `gitBranch` is only
populated for the folder-HEAD poll work-set, so an ordinary git repo with no
live session in its cwd rendered **no button at all** — indistinguishable from
a render bug. Three divergent gates existed for the same question (board
`gitBranch`; sidebar inline `some(s => s.isGitRepo !== false)`, which fails
CLOSED on zero sessions; `folderIsGitRepo`, which fails OPEN).

## What

- New pure helper `packages/client/src/lib/git/folder-worktree-availability.ts`
  — `resolveWorktreeAvailability` returns `{ available, reason? }`:
  preference gate first (`undefined` reads as ON, so no cold-load flash),
  then `folderGitMap` as **positive evidence only** (`null` = unknown — it
  conflates non-git, unborn repo, read error and stale cache), then a session
  reporting `isGitRepo === false`, else fail open. Folder/session identity goes
  through the shared `pathKey`.
- Board, sidebar `+ New Worktree` and `folderIsGitRepo` all consume it, so the
  three gates collapse into one rule.
- Board renders the action **always**, `disabled` + explaining `title` when
  unavailable (design D7) — an unavailable state is now directly assertable.
- Sidebar behaviour change (design D6): a pinned git folder with zero sessions
  now shows `+ New Worktree`.
- Repairs `tests/e2e/helpers/openspec-board.ts`, which waited on the
  `folder-open-home-<cwd>` testid deleted by `add-folder-actions-menu` — every
  `openBoard`-based spec was dying there.

## Verification

- Full `npm test` green on the merged tree (18180 passed).
- New L1 suite (14 cases) + rewritten board/sidebar RTL rows; the former
  "button absent" assertions become "present + disabled + reason".
- New L3 spec `tests/e2e/openspec-board-worktree-availability.spec.ts`, proven
  **red against the old gate** and green with the fix in the docker harness.
- Enforcers green: check-conventions, dox-byte-gate, i18n-lint --strict,
  i18n-parity, knip-config, knip-ratchet (+ baseline-diff).
- `@review` pass: no blocking findings; both non-blocking findings fixed
  (thread `folderGitMap` into `folderIsGitRepo`; de-vacuum the E10
  path-normalization assertion).


robertcsakany:
Applied 2 of 3 findings:

- **E11 liveness test** — now pins `isGitRepo:false` on the same-folder session and expects `not-a-git-repo` for both `ended` and `active`, so a resolver that skipped ended sessions fails open and gets caught.
- **tests/e2e/AGENTS.md** — corrected to the per-run `/fixtures/board-wt-<base36 ts>` repo the spec actually creates.

**Disputing the third** (`openspec-board-worktree-availability.spec.ts`, "wait for `session_removed`, not only shutdown"): the concern is that an ended row keeps its `gitBranch` and so could satisfy the legacy gate. That is exactly right — and it is why the spec's *discriminating* half is **Phase A**, which runs against a per-run fixture with provably zero sessions and therefore no `gitBranch` anywhere. Phase A was verified **red** against the old gate in the docker harness (`expect(cold.every(Boolean)).toBe(true)` → received `false`) and green with the fix. Phase B is a convergence check — no post-shutdown assertion *can* discriminate the gates, because the client keeps the ended session with its persisted `gitBranch`; what it pins is that ending a session must not regress availability. A comment recording this now sits above Phase B.

PR #621, opened 2026-09-09
feat(add-git-checkout-root-resolver): one shared git checkout-root resolver

robertcsakany:
Implements OpenSpec change `add-git-checkout-root-resolver`.

## What

Three call sites each re-derived "which repo is this?" by taking `dirname(--git-common-dir)`. That is wrong for a linked worktree (yields the `.git/worktrees/<name>` husk) and for a submodule. This replaces the copies with **one shared resolver** in `packages/shared/src/platform/git.ts` that returns three facts: `thisCheckout`, `isLinkedWorktree`, `mainCheckout`.

Converted consumers (all read-only):
- `vcs-info.detectWorktree` — worktree/main path reporting
- kb-plugin cwd guard — admission anchor
- `session-scanner` — load-time repair of persisted phantom `mainPath` values

## Verification

- Real-repo fixture builder materializes **nine** git states; every probe triple was measured, not reasoned.
- 29 unit scenarios + 3 Playwright E2E scenarios (`tests/e2e/submodule-folder-grouping.spec.ts`) green against the docker harness, twice.
- Full suite: 18241 passed. One unrelated client perf test (`history-gap-trigger` P1) flaked under full-suite load; passes in isolation, and this change touches zero client files.

## Notes

- kb guard probe budget lowered 2000ms → 500ms: the guard went from one synchronous probe to four, so this holds the *previous* 2s worst case without an async refactor.
- Two follow-up changes scaffolded as proposals: `apply-checkout-root-to-worktree-ops`, `widen-containment-to-resolved-checkout`.
- Task 8.2 (manual submodule smoke) marked done — its surface is covered by E2E F1–F3, verified post-merge.


robertcsakany:
Review round 1 addressed. 7 of 9 fixed in 170277c47; 2 disputed with evidence.

**Fixed**

- **Bare hub named `.git` (Major).** Real finding. Rule 2b is now gated on repo-local `core.bare`, via a new `GIT_CONFIG_LOCAL_CORE_BARE` probe under the same `--local` + argv-form constraints as `core.worktree`. Added real-repo fixture **E5b** (`git clone --bare` into `<parent>/.git`, then `worktree add`) asserting `mainCheckout === null`. I verified it **fails closed**: reverting the guard turns E5b red. Spec, delta spec, design and `docs/architecture.md` (incl. the mermaid gate) all updated.
- **`.git` cwd admitted by direct known-folder match.** Fixed: `isAllowedCwd` now rejects a canonical cwd with an exact `.git` segment **before both** admission paths. Added a route test with `<repo>/.git` in the known-folder set asserting 403 with no store opened — also verified to fail closed. Specs updated.
- **Probe budget.** The bareness probe makes it five sync probes, so `timeout` went 500ms → **400ms**, still holding the superseded single-2000ms worst case rather than multiplying it.
- **`process.env.X = undefined`.** Correct and worth fixing. Extracted `restoreEnv(name, saved)` into `git-fixtures.ts` (deletes when unset) and used it in all three suites.
- **`execFileSync("test", …)` canary.** Replaced with `existsSync(canary)` — you're right that an ENOENT lookup failure satisfied `.toThrow()` without ever checking the canary.
- **E2E cleanup fallback.** Real bug. Switched to the bus path (`connectBus` + `shutdownSession`, both now exported from `fixtures.ts`); the bare `request.newContext()` had no baseURL, so every relative POST rejected and the fallback shut down nothing.
- **Full-suite count.** Reconciled. The archived record now states both runs and their differing failures.

**Disputed**

- **`__dirname` in `packages/kb-plugin/vitest.config.ts`.** Not a defect here. `packages/server`, `packages/client` and `packages/extension` are all `"type": "module"` and all use `__dirname` in their vitest configs; Vite loads the config through its own transform, so `__dirname` is defined. All four suites pass, including this one. Diverging from the mirrored configs would be inconsistency without a benefit.

**Carried forward, not fixed here**

- The wider `.git`-segment / containment surface beyond this guard is already scoped as follow-up change `widen-containment-to-resolved-checkout`, alongside `apply-checkout-root-to-worktree-ops` for `resolveMainPath`'s twelve remaining consumers.

Full suite after the fixes: **18244 passed, 0 failed**. `tsc --noEmit`, `lint:e2e`, `biome check --changed` and `check-conventions` all clean.

robertcsakany:
Round 2 addressed in 3a0e7a41c.

**Fixed — both Major findings on the bareness guard were right**

- **`core.bare = yes` was read as NOT bare.** Correct catch. The probe now uses `--type=bool`, so git canonicalizes `yes` / `on` / `1` / `true` to exactly `true`, and the comparison is no longer against raw config text. Regression fixture **E5c** builds a real bare hub at `<parent>/.git` with `core.bare = yes` and asserts `mainCheckout === null`.
- **A failed or timed-out probe took the same fallback.** Also correct, and the sharper of the two. The probe is now three-valued — exported `GitBareness = "not-bare" | "bare" | "unknown"` — and rule 2b requires a **positive** `"not-bare"`. A spawn failure or timeout is `"unknown"` and falls through to rule 3 (`null`), so an unread probe can no longer re-open the fallback the check exists to close. An *unset* key stays `"not-bare"`: that is git's own boolean default and a successful read, not a failure to read. Regression scenario **E5d** covers the unknown probe, with a positive control on the same shape proving the gate still resolves when the answer is real.

I verified both new tests **fail closed** by reverting each guard independently.

**Disputed — `.git`-traversal before canonicalization**

`<repo>/.git/..` is not a traversal; it denotes `<repo>` itself, and `<repo>` is the known folder. Two independent reasons the raw-string check adds nothing:

1. `hasGitPathSegment` calls `normalizePath` *before* splitting into segments, and `normalizePath` collapses `..`. So `hasGitPathSegment("<repo>/.git/..")` already evaluates `<repo>` — the raw and canonical tests return the same answer.
2. `canonPath` additionally realpaths, so a symlink **aimed at** a git dir (`/alias -> /repo/.git`) is caught by the canonical test and would be *missed* by a raw-string test. The canonical form is strictly the stronger of the two.

Nothing under `.git` is served in this path — the request is admitted as the known folder it names. I pinned that behaviour in a test (`admits a known folder spelled with a .git/.. round trip`) with the reasoning in the source comment, so a future "reject the raw string too" change has to justify breaking a legitimate spelling.

**Also disputed — `__dirname` in `packages/kb-plugin/vitest.config.ts`** (re-raised from round 1)

`packages/server`, `packages/client` and `packages/extension` are all `"type": "module"` and all use `__dirname` in their vitest configs. Vite loads the config through its own transform, so `__dirname` is defined; all four suites pass. Diverging from the three mirrored configs would be inconsistency without a benefit.

**Round-1 thread still open on `design.md`:** that is the bare-hub finding, fixed in 170277c47 (guard + fixture E5b) and further hardened here.

Full suite: **18257 passed, 0 failed.** `tsc --noEmit`, `lint:e2e`, `biome check --changed`, `kb dox lint` and `check-conventions` all clean.

robertcsakany:
@coderabbitai full review

robertcsakany:
Round 3 addressed in ffb38bed4.

**Fixed**

- **`name` derived from `cwd` instead of `thisCheckout`** (nitpick, but a real user-visible bug). A session sitting in a subdirectory of a worktree labelled its folder card with the *subdirectory* name. Now `path.basename(roots.thisCheckout ?? cwd)`, with a regression test that puts the session three levels deep and asserts the worktree's own name.
- **Proposal out of step with the fail-closed bareness rule.** Correct — I hardened the code and the specs in round 2 but left `proposal.md` describing the old rule. Prose and flowchart now both carry the `--type=bool` / confirmed-not-bare / unknown-falls-through gate.
- **`git-context` requirement vs the bridge's `.git`-segment omission.** Fixed in both the synced spec and the delta: a resolved main checkout containing a `.git` segment is treated as *no worktree*, with its own scenario. Also corrected the `name` clause to say the worktree root rather than the cwd.
- **Git 2.31.0 floor.** Documented, in the spec and at the recipes themselves: `--path-format=absolute` needs git ≥ 2.31, and below that floor both probes fail, resolution yields `null`, and every consumer degrades to its no-result branch — the card omits `gitWorktree`, the guard rejects. Fail-closed, but worktree admission is unavailable below 2.31.

**Carried to a follow-up, deliberately — "bind `mainCheckout` to the repository"**

This one is right, and it is the sharpest finding in the whole review. I am not fixing it here, for two reasons:

1. **It is not introduced by this change.** The permissive `mainCheckout` anchor predates it; this change converted the *derivation* from `dirname(--git-common-dir)` to the shared resolver and, in every state except a submodule admitted on its own cwd, made the guard **stricter**, not looser.
2. **The real fix is a new probe on an authorization request path** — re-resolving the claimed main checkout and requiring it to point back at the same common dir. That is a security widening/narrowing with its own spec deltas, its own `security-hardening` pass, and its own tests. Landing it as an unreviewed addendum to a display-bug fix is exactly the failure mode this repo's review discipline exists to prevent.

It is now recorded in `openspec/changes/widen-containment-to-resolved-checkout/proposal.md` under *"Also carried forward: bind the kb guard's anchor to the repository"*, alongside the closely-related `core.worktree`-outside-the-repository case that proposal already owned. I explicitly wrote down that the bounded-reach argument (the store opens at the request's own `cwd`) must be **re-tested, not re-assumed**, because `reindexAll` follows a cwd-local `knowledge_base.json` whose `resolvedSources` need not stay under that cwd — which is the strongest version of your point.

**Still disputed** — `__dirname` in `packages/kb-plugin/vitest.config.ts` (three sibling `"type": "module"` packages do the same and all pass) and the `<repo>/.git/..` raw-string check (`hasGitPathSegment` normalizes before splitting, so raw and canonical give the same answer, and `canonPath` additionally catches a symlink aimed at a git dir). Evidence in my previous two comments.

Full suite at the previous commit: 18257 passed, 0 failed. This commit's affected suites: 2103 passed, 0 failed; `tsc --noEmit` and the client build clean.

robertcsakany:
Round 4 addressed in 63e8a3de2. All three fixed.

- **Linked worktree with a null `thisCheckout`.** You're right, and my round-3 fix was half a fix: `?? cwd` reintroduced the subdirectory mislabel for exactly the case that can't be verified. `detectWorktree` now returns `undefined` when `roots.thisCheckout` is null — a linked worktree always *has* a working tree, so null means the `--show-toplevel` probe failed, which is inconclusive rather than "nameless". Regression test drives a deep subdirectory through that verdict.
- **Mixed-probe-form scenario vs the probe contract.** Correct contradiction, and the fix is to say what the scenario actually is. It is a *wiring* guard — unreachable through the canonical wiring precisely because `--path-format=absolute` is required on both probes — so it is now stated as verified with **injected** probes, with an explicit clause that production keeps the flag on both sides. That is also how it is tested.
- **KB guard missing from the follow-up's Impact and tasks.** Added: `widen-containment-to-resolved-checkout` now lists `isAllowedCwd` + its tests in Impact, carries a MODIFIED delta for `kb-plugin-cwd-guard`'s *Git-repo-main admission* requirement, and gains tasks **1.4** (bind `mainCheckout` to the repository — re-resolve and require it to point back at the same common dir), **1.5** (re-test, don't re-assume, the bounded-reach argument, since `reindexAll` follows a cwd-local `knowledge_base.json` whose `resolvedSources` need not stay under that cwd) and **2.2** (route test proving a `core.worktree` aimed at an unrelated known folder is rejected, with a positive control that an honest worktree of a known repo is still admitted).

**Standing disputes, unchanged** — `__dirname` in `packages/kb-plugin/vitest.config.ts` (three sibling `"type": "module"` packages do the same, all pass) and the `<repo>/.git/..` raw-string check (`hasGitPathSegment` normalizes before splitting; `canonPath` additionally catches a symlink aimed at a git dir, which a raw test would miss). Evidence in my earlier comments.

**Standing carry-forward** — "bind `mainCheckout` to the repository" stays out of this PR. It predates this change, the conversion made the guard strictly *stricter* in every state but one, and the real fix adds a probe to an authorization request path. It is now fully specified in the follow-up rather than half-done here.

Full suite: **18259 passed, 0 failed.** `tsc --noEmit`, `check-conventions`, `kb dox lint` and the client build all clean.

robertcsakany:
Round 5 addressed in 779b9ed25 — one finding, fixed.

**Unset `core.bare` was undefined in the proposal.** The code and the specs already mapped absence to not-bare, but `proposal.md` only said "fail-closed", which reads as if absence were `unknown`. It now states the three-valued split explicitly:

- **unset** → `not-bare`. That is a *successful read of git's own boolean default*, not a failure to read, so it takes the fallback.
- **command error / timeout / unparseable** → `unknown`, which never takes the fallback.

Added real-repo fixture **E5e**: a clone with `core.bare` removed via `config --local --unset`, asserting the fallback still resolves to the main checkout. The absence is asserted by the throw, since `git config --get` exits 1 on an unset key. E5e and E5d now pin the two halves of the distinction against each other — unset resolves, unknown does not.

Full suite: **18259 passed.** One unrelated `cors.test.ts` timing assertion flaked under full-suite load and passes in isolation; this change touches no server CORS code. `tsc --noEmit`, `check-conventions` and the client build clean.

---

Convergence across the five rounds: **9 → 4 → 3 → 1 → (this one)** actionable findings, 15 fixed. Two standing disputes with evidence (`__dirname` in `packages/kb-plugin/vitest.config.ts`; the `<repo>/.git/..` raw-string check) and one deliberate carry-forward ("bind `mainCheckout` to the repository", now fully specified as tasks 1.4/1.5/2.2 of `widen-containment-to-resolved-checkout`). Unless the next pass surfaces something new and blocking, I intend to merge.

PR #623, opened 2026-09-11
fix(fix-stuck-streaming-status-latch): reconcile session status against bridge agent liveness

robertcsakany:
Implements OpenSpec change `fix-stuck-streaming-status-latch`.

## Problem

A session card could show `Thinking…` forever. `status: "streaming"` was a **one-way latch** — `agent_end` was the ONLY path back to `idle` in the whole server, and `event_forward` is fire-and-forget (no ack, no replay queue). So a single dropped `agent_end` was unrecoverable without a restart. Bridge WS churn made it common: 839 `connection closed` in one observed server run, 42 `heartbeat timeout but WS still OPEN`.

## Fix

- **One carrier, no new message type.** `SessionHeartbeatMessage` gains an optional `agentRunning: boolean` read from `getBridgeState().isAgentStreaming`. The server reconciles a mismatch on each beat via a new pure `reconcileAgentLiveness(currentStatus, agentRunning)`.
- **The reconnect heal becomes symmetric.** The pre-existing mid-turn synthetic `agent_start` is untouched; the idle case sends one `session_heartbeat{agentRunning:false}` from a new `onPostFlush` hook in `ConnectionManager`. It must be post-flush: `onopen` fires `onReconnect()` *before* draining the buffer, so a heal sent earlier would overtake a buffered real `agent_end` and silently erase the `streaming→idle` unread edge.
- **Never a synthetic `agent_end`** — that event carries run-boundary side effects (unread stamping, card reordering, OpenSpec clearing) a correction must not fire. The unread exemption is by **call path** (the reconcile never enters `isUnreadTrigger`), not by a flag.

Guards: `ended` is inert in both directions (a beat racing teardown must never resurrect); `active` + not-running is inert (no churn per beat); agreement is inert; the reconcile is suppressed during replay and preserves a live `ask_user`. `→ idle` clears `currentTool`, `→ streaming` deliberately does not.

Additive + optional, so old bridge/new server and new bridge/old server both degrade to today's behaviour — no coordinated deploy.

## Verification

- 8 pure-function tests (full `status` × `agentRunning` decision table), 10 server wiring tests, 5 extension tests against the real `ConnectionManager`, 1 new L3 Playwright spec verified against a rebuilt docker harness.
- Full suite: 18284 passed, 0 failed. Client build green. All enforcers green (check-conventions, dox-byte-gate, i18n lint/parity, knip-config, knip-ratchet + baseline-diff).
- Independent `@review` checkpoint: approve, no blocking findings; two nits applied.

One known gap, recorded in `tasks.md`: the two `bridge.ts` lines that *emit* liveness have no automated coverage (no suite mounts `initBridge`), so they are covered by the deferred live-instance check instead.


robertcsakany:
Out of scope for this PR — disputing with evidence.

These files belong to change `add-git-checkout-root-resolver`, not to `fix-stuck-streaming-status-latch`. They appear in this PR's three-dot diff only because this branch was cut from `develop` at `d5c323da0` (`plan(add-git-checkout-root-resolver): ...`), and PR #621 later **squash**-merged that work, rewriting the commit — so the original planning commit is no longer an ancestor of `origin/develop` and its files surface as if added here.

Verification:

```
git log --oneline --all --source -- openspec/changes/add-git-checkout-root-resolver/design.md
d5c323da0  refs/heads/develop  plan(add-git-checkout-root-resolver): ...
```

`develop` already carries that change archived at `openspec/changes/archive/2026-09-09-add-git-checkout-root-resolver/`, plus the `apply-checkout-root-to-worktree-ops` follow-up. Nothing in this PR authors or modifies these files, and squash-merging this PR into `develop` leaves them untouched.

The two in-scope findings were applied in f285086a3.

PR #627, opened 2026-09-11
fix(fix-quota-widget-clipping): budget chat-pane height so bottom rows never clip

robertcsakany:
Implements OpenSpec change `fix-quota-widget-clipping`.

## Problem

The quota bar at the bottom of the chat pane rendered half-height. This is a
**chat-pane height-budget bug, not a quota-plugin bug** — `packages/quota-plugin/`
is untouched.

`split-chat-pane` is `flex-col overflow-hidden`. `ChatView` is the only elastic
row; every other row is floored at min-content by `min-height: auto`. Once
`ChatView` surrenders its grow allocation (measured `23 → 13 → 3 → 0`), the fixed
furniture below overflows and the last child gets guillotined.

An earlier pass diagnosed this as `flex-shrink` distribution and added
`flexShrink: 0` to the widget. Live measurement proved that a **no-op** (`DIFF=0.0`
across a `220 → 140` sweep); it was reverted.

## Fix

| element | change |
| --- | --- |
| `composer-root` | `max-h-[40%]`, `flex flex-col`, overflow **visible** |
| `composer-card` | `min-h-0`, `overflow-y-auto` (the scrollport) |
| `composer-context-strip`, `status-bar`, footer-slot wrapper | `shrink-0` |

**The scrollport must stay on `composer-card`.** `composer-root` is the positioned
containing block for the `/command` and `@file` autocomplete dropdowns, which
render upward outside its box — making it a scrollport clips them to `0px` visible
(measured `269px` dropdown, `272px` clipped). The first implementation did exactly
that; review caught it. A regression test asserts `composer-root` does **not**
carry `overflow-y-auto`.

## Verification

In-browser sweep, real quota widget (not a probe):

| pane | widget visible | clipped | transcript |
| --- | --- | --- | --- |
| 457 | 13/13 | 0 | 260 |
| 220 | 13/13 | 0 | 62 |
| 190 | 13/13 | 0 | 44 |
| 160 | 13/13 | 0 | 26 |

Baseline at `pane=190` was `6/13` visible, `7px` clipped, transcript `0px`.
Mobile A/B at `390×844` confirms the unfixed build clips at every height in the
`170 → 130` band while the fixed build clips none of it until `~145`.

New tests verified **red first** by stashing only the source fix.

## Known tradeoffs (documented, not hidden)

- The composer now caps at `183px` where it previously grew to `262px`; it scrolls
  past the bound instead of growing.
- Below ~`145px` of chat pane the row floors exceed the pane and clipping resumes.
  Unreachable on a real phone (~`400px` pane), but real — recorded in `design.md`
  and the spec scenario is worded to match rather than overstate.

### DIFF
--- a/openspec/changes/archive/2026-09-11-fix-quota-widget-clipping/.openspec.yaml
+++ b/openspec/changes/archive/2026-09-11-fix-quota-widget-clipping/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-11
--- a/openspec/changes/archive/2026-09-11-fix-quota-widget-clipping/design.md
+++ b/openspec/changes/archive/2026-09-11-fix-quota-widget-clipping/design.md
@@ -1,0 +1,98 @@
+# Design — chat-pane height budget
+
+Written after implementation, at the point of archive. The change entered as a
+one-property tweak and did not warrant a design doc; it left as a shared-furniture
+layout contract with two non-obvious constraints that a future editor will
+otherwise re-break. Those constraints are the reason this file exists.
+
+## Where the height goes
+
+`split-chat-pane` is `flex-col overflow-hidden`. Its children, top to bottom:
+
+```mermaid
+flowchart TB
+  pane["split-chat-pane<br/>flex-col, overflow-hidden"]
+  chat["ChatView — flex: 1 1 0%<br/>the only elastic row"]
+  strip["composer-context-strip — shrink-0"]
+  status["status-bar — shrink-0"]
+  comp["composer-root — max-h-[40%], flex-col, overflow VISIBLE"]
+  card["composer-card — min-h-0, overflow-y-auto<br/>(the scrollport)"]
+  foot["content-inline-footer wrapper — shrink-0<br/>(QuotaWidget and friends)"]
+  pane --> chat --> strip --> status --> comp --> card
+  pane --> foot
+```
+
+Only `ChatView` is elastic. Every other row has `min-height: auto`, which floors it
+at min-content — so `flex-shrink: 1` on those rows is decorative. When the pane
+shrinks, `ChatView` surrenders its grow allocation (measured `23 → 13 → 3 → 0`) and
+then has nothing left to give; the furniture below overflows and `overflow-hidden`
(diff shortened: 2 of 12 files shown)


robertcsakany:
@coderabbitai review

robertcsakany:
Valid finding — accepted in part, with the remainder tracked as follow-up.

**Accepted and fixed in f90228a1:** the sentence claiming proportional
below-floor degradation was wrong. Measured A/B at `390x844` shows the shortfall
landing on the bottom-most row, not shared:

```
pane:     150 145 140 135 130
clipped:    0   1   6  11  13
```

The spec now states the actual behavior (transcript holds at its floor, bounded
composer at its minimum, remainder clipped by the pane's `overflow: hidden`) and
explicitly records that no weighted allocation is specified.

**Deferred, deliberately:** implementing a real shrink-weight/lower-bound
allocation plus height-boundary regression cases is new behavior rather than a
correction to this change, and it only affects panes below roughly `145px`. A
stacked mobile split on an `844px` screen gives the chat pane about `400px`, so
that band is not reachable in practice. Tracked as OpenSpec change
`define-chat-pane-below-floor-allocation` (added in this PR), which carries the
measurements and names the missing height-boundary coverage in
`tests/e2e/split-composer-overflow.spec.ts`.

robertcsakany:
@coderabbitai full review

robertcsakany:
Fixed the canonical half in 279d04e8; declining the archived half, with reason.

**Fixed:** the mobile scenario still said "degrade proportionally", contradicting
the bottom-most-clipping rule stated two paragraphs above it in the same
requirement. `grep -n proportional openspec/specs/split-editor-workspace/spec.md`
now returns nothing, so the canonical contract states one rule in all three
places.

**Not changing the two archived files** under
`openspec/changes/archive/2026-09-11-fix-quota-widget-clipping/`. This repo treats
archived changes as immutable history — `scripts/check-conventions.mjs` skips that
tree by design, and an earlier comment in this same review explicitly said "do not
rewrite archived history". The archived proposal is the record of what was
believed at the time the change shipped, including the diagnosis that was later
corrected; editing it would erase the audit trail rather than fix anything.
`openspec/specs/` is the live contract and is now correct.

robertcsakany:
@coderabbitai full review

PR #654, opened 2026-09-14
feat(extract-mcp-client-plugin): extract MCP client config into its own plugin

robertcsakany:
Implements OpenSpec change `extract-mcp-client-plugin`.

## What changed
- New `packages/mcp-client-plugin`: `mcp-client.config` service, REST surface
  (`/api/mcp-client/*`), effective merged view with layer provenance + secret
  redaction, schema-driven server editor + global settings draft source, and
  folder-scoped MCP pill/page.
- `mcp-server-plugin` and `apple-tools` now consume the service (removed the
  local adapter probe + local config writer); `isAllowedCwd`/`canonPath`/
  `mainCheckoutPath` moved to `packages/shared/src/cwd-guard.ts`.

## Verification
- `npm test` green (local: 19,257 passed; 1 pre-existing flaky server test that
  passes in isolation).
- Enforcers green: check-conventions, dox-byte-gate, i18n-lint/parity,
  knip-config, knip-ratchet (+ baseline-diff).
- Docker harness L3: 15/15 change-relevant Playwright specs pass.
- `review-code` + `security-hardening` passes: no blocking findings.

## Deferred
- F30 mockups visual parity — manual post-merge check (test-plan: manual-only).

## Known non-blocking follow-ups (security + review passes)
- `PUT /servers/:name` accepts `unset: ["disabled"]`, bypassing the enable
  re-evaluation that `PUT /servers/:name/disabled` performs (a lower-layer
  disable could remain effective); route `disabled` through `setServerDisabled`.
- `/effective` 500 responses echo internal messages (absolute config paths) to
  authenticated callers; genericize for defense-in-depth.
- `isAllowedCwd` for an unknown cwd runs synchronous `git` probes up to ~2s on
  the request path (inherited from kb-plugin, newly reachable); consider a
  negative-result cache.
- `writeFileAtomic` does not fsync the parent directory after rename
  (durability only).


robertcsakany:
@coderabbitai full review

PR #657, opened 2026-09-14
fix(backpressure-status): reconcile shed session_updated from live state

robertcsakany:
Implements OpenSpec change `fix-backpressure-status-and-subagent-frames`.

## Why

A session running a multi-minute subagent renders as **Idle** while the server knows it is `streaming`. The card never picked a wrong value — it never received the right one.

`session_updated` is transcript-class, so `fanout()` sheds it whenever a browser socket sits above `MAX_WS_BUFFER`. What makes that drop visible *for minutes* is that status is never re-pushed: during a long tool call the parent emits one status frame and then goes quiet, so **a single dropped frame is sufficient**, and it stays stale until a reconnect or a reload. There is no seq, no backfill, no periodic reconcile.

## What changes

**Lever A — status becomes self-healing.**

- `broadcast()` derives `dirtyId = msg.type === "session_updated" ? msg.sessionId : undefined` and threads it into `fanout()`. `fanout` receives only the pre-serialized string and cannot recover the type or id without parsing, so the typed caller has to supply it; every other caller passes `undefined`.
- The shed site records the id in a per-socket `Set<string>` — **ids only, never a queued payload**, so it cannot reach the pending-state byte ceiling or move `stalledSocketsTerminated`.
- The debt owns **its own** 250 ms timer, deliberately not the pending-state one: that timer is created only when a *state* frame defers, and a socket saturated purely by transcript traffic — the incident's exact shape — never creates it.
- On tick, while under threshold, the frame is **rebuilt from `sessionManager.get(id)`**. Nothing stale is ever queued, so two partial `updates` never have to be merged.
- The reconcile send carries `ctx.sessionId`, so a reconcile that is itself shed re-enters the debt at the drop site — eventually-delivered rather than check-once, and idempotent (a `Set`).
- Released on socket `close`, on `error`, and on the stalled-socket `terminate()` path.

Accepted limitation, spec'd rather than footnoted: the re-push carries the **settled** value, so an edge shed entirely inside one flood window is not recovered. Strictly better than today, where the final value is lost too.

**Scope.** `session_updated` only. `session_added` / `session_removed` / `sessions_reordered` stay transcript-class and stay unrecovered — create/delete/reorder are not idempotent re-pushes of one row. No client change: the existing `if (existing)` guard already makes a reconcile for an unknown row a no-op, which is what stops a re-push resurrecting a row a shed `session_removed` deleted.

**Lever B — make the back-pressure claim measurable (telemetry only).**

- `/api/health` gains `droppedFrames.statusReconcileQueued` / `statusReconcileSent`, **beside** the drop counters and never folded into them, so the reconcile cannot mask the shed it recovers from.
- `/api/health` gains `socketBufferOccupancy { max, p95, msAboveThreshold }`. Sampled at the send-decision sites that already read `bufferedAmount` — **not** on a timer. A periodic sampler was implemented first and rejected: it broke the gateway's "zero timers on an unsaturated socket" invariant (caught by the existing `browser-gateway-critical-frames` P3/E9). `msAboveThreshold` includes spans still open at read time, so an in-progress stall does not report `0`.

## Testing

16 L1 scenarios + 2 L3 Playwright specs (test-plan E1–E8, X1–X6, P1–P2, F1–F2).

Real back-pressure is a browser failing to drain its own socket, which Playwright cannot induce and no existing harness lever can force. The L3 pair therefore rides `POST /api/test/force-shed`:

- registered **only** under `PI_E2E_FORCE_SHED=1`, set in `docker/compose.test.yml` and never in a real image — on a production instance the route does not exist (404);
- still `networkGuard`-gated;
- returns its effective state, so a harness booted without the flag fails loudly instead of passing as a silent no-op;
- scenario **X6** pins the production-default inertness at L1.

Both L3 specs verified green against the docker harness across four runs, including a rebuild after the final refactor.

## Review

Reviewed by an isolated `@review` subagent over two rounds. Round 1 returned BLOCK (13 findings); round 2 caught a `tsc`-red duplicate-identifier regression the first round's fixes introduced. All blocking and non-blocking findings addressed — notably: the reconcile no longer counts a shed frame as `sent` (`sendTo` now returns a delivery boolean), `msAboveThreshold` reports in-progress stalls, and the shed predicate is deduplicated into a single `shouldShed()` used by all three call sites. The two behavioural fixes were verified to **fail** against the pre-fix code.


robertcsakany:
All four findings were valid and are fixed in 2aff3e0bc.

1. **`currentTool` omission (browser-gateway.ts)** — confirmed a real bug, not cosmetic. The client merges `{ ...existing, ...msg.updates }`, so an `undefined` dropped by `JSON.stringify` left the *stale tool name* in place; a session whose tool had finished reconciled back to the old tool. Now sends `session.currentTool ?? null`.

2. **Unsettled occupancy span (browser-gateway.ts)** — also real, and worse than a rounding error: sampling is event-driven, so a socket that drained without another send decision kept its span open and the read-time pass added elapsed time on *every* health read, growing without bound. `getSocketBufferOccupancy()` now settles spans whose socket is no longer above the threshold (or is closed); the delete makes it idempotent against a later exit.

Both are pinned by tests that were verified to **fail** against the pre-fix code.

3. **Observation-based duration** — agreed; the spec overstated it. `openspec/specs/ws-frame-delivery-policy/spec.md` and `docs/architecture.md` now state that occupancy is derived from samples taken at send decisions, that a crossing entirely between two decisions is not observed, and that a drained socket must stop accruing — plus a scenario pinning the no-growth-per-read property.

4. **E2E attribution** — good catch. The `ended` spec now polls `statusReconcileQueued` while shed, before releasing, so the post-release delivery is attributable to the reconcile rather than passing vacuously had the server never emitted the update.

PR #662, opened 2026-09-14
fix(ci): allowlist the vendored playwright-core internal aliases

robertcsakany:
## Why

`ci.yml` has been **red on develop** at *Verify published imports are declared* since `add-browser-relay` landed (run 34823973372, commit 69e08a20e), with 5 `undeclared-import` errors. Because that step fails early, **every downstream step is skipped repo-wide** — `pnpm test`, `Build`, `Biome`, `Typecheck`, `lint` — so no PR could be gated on them either. (That is what happened to #659.)

## Root cause

The flagged specifiers are **not dependencies**. `packages/browser-plugin` **vendors** playwright-core's relay, and those vendored files import bare *internal* aliases that do not exist on npm:

```
@isomorphic/manualPromise  @isomorphic/time  @isomorphic/timeoutRunner  @utils/wsServer
```

`tsconfig.base.json` `paths` maps each one to a local shim under `packages/browser-plugin/src/server/relay/vendor/shims/`; vitest resolves them through `resolve.alias`; and the real server enables jiti `tsconfigPaths` for them — `packages/server/bin/pi-dashboard.mjs:125-126` documents exactly this and why (without it the plugin fails to load and the whole relay is dead at runtime).

Declaring them would write **unresolvable dependencies into a published manifest** — strictly worse than the import the check exists to catch.

## Fix

Uses the checker's own **reasoned-allowlist** mechanism (an entry without a `reason` fails the run, so exceptions cannot accumulate silently), matching the existing pi-forms-bpmn spread pattern. 4 entries for `packages/browser-plugin`, each carrying the rationale.

## Verification

- `node scripts/verify-published-imports.mjs` → **`42 non-private workspace(s) checked · 0 error(s) · 0 warning(s)`**, exit 0 (run after a faithful `pnpm install --frozen-lockfile`).
- `scripts/__tests__/verify-published-imports.test.mjs` + `dependency-declarations.test.mjs` → **76 passed / 2 skipped**. The latter is the P1 assertion that the full-repo CLI run exits 0 within its budget — the regression guard for this fix, so no new test was needed.
- `biome lint .` → exit 0, **0 errors** (warnings only).
- `check-conventions --base origin/develop` and `knip-ratchet --check-baseline-diff origin/develop` → pass.
- `dox-byte-gate` → pass (the sidecar row was updated).

## Note

This deliberately does **not** touch `packages/browser-plugin` source, the shims, or the vendored tree — the vendored files and their hash/integrity test are untouched. Only the checker's exception list and its DOX row change.


robertcsakany:
Obsolete — develop already carries both changes verbatim.

- The 4 `packages/browser-plugin` allowlist entries (`@isomorphic/manualPromise`, `@isomorphic/time`, `@isomorphic/timeoutRunner`, `@utils/wsServer`) landed in f2532376e (#663). The rebase conflict is a pure duplicate: the `.map()` block is identical, and develop's reason text is strictly better — it cites the never-edit + per-file SHA-256 integrity rule (`relay/vendor/NOTICE`, `vendor-integrity.test.ts`) that makes the specifiers unrewritable, which this PR's wording omitted.
- The kb engine fingerprint refresh landed in e9264035e (#664); `packages/kb/engine-fingerprint.json` on develop is byte-identical to this branch's version.

This PR's CI failure was never the import check. In run 34826906665 the `Verify published imports are declared` step reported `42 non-private workspace(s) checked · 0 error(s) · 0 warning(s)` and both `verify-published-imports.test.mjs` (56) and `biome-undeclared-dependencies.test.mjs` (14) passed. The only red step is `Run pnpm test`, failing on `packages/server/src/rpc-keeper/__tests__/keeper.test.ts > E5: child-only growth rotates via the interval timer` — a pre-existing develop-wide failure a rebase would inherit, not fix.

Residual tracked separately: `scripts/verify-published-imports.mjs.AGENTS.md` on develop still under-describes develop's own allowlist (no mention of the pi-forms-bpmn nested-library deps or the browser-plugin aliases). Follow-up PR incoming for that row only.

Issue #665, opened 2026-09-14
harden DELETE /api/paired-devices/:id with the authVia marker (device bearer can revoke sibling rows)


robertcsakany:
From change: mcp-legacy-clients-and-token-issuance (D5, "Note /api/pair/approve and DELETE /api/paired-devices/:id keep their weaker networkGuard posture").

## Problem

`DELETE /api/paired-devices/:id` (packages/server/src/routes/pairing-routes.ts) is guarded by `networkGuard` only. `networkGuard` admits ANY `request.isAuthenticated` request — including one authenticated by a valid **paired-device bearer** (bearer-auth.ts sets it). So a paired device (e.g. a phone paired over a tunnel) can revoke its SIBLING rows.

The same weaker posture applies to `POST /api/pair/approve` (its docblock already flags Phase C hardening).

## Why it matters now

The change added the additive `request.authVia: "session" | "device"` marker exactly to make this tightening a one-line change: `operatorGuard` on the new `POST /api/paired-devices` mint route already refuses `authVia === "device"`. `DELETE /api/paired-devices/:id` should do the same — a device credential must not be able to revoke other devices.

## Suggested fix

Swap `networkGuard` for an operator-style guard on `DELETE /api/paired-devices/:id` (admit `authVia === "session"`, `X-Pi-Local-Token`, or genuinely local; keep/consider Host admission), mirroring `createOperatorGuard` in packages/server/src/routes/pairing-routes.ts.

## Related (same class, smaller)

The pairing `approve` path still writes UNBOUNDED labels — `PairingManager.approve` does not validate label length (design D4: "approve is left as is"; risks table: "Pairing path still writes unbounded labels — out of scope"). Consider validating the label (trim + `MAX_DEVICE_LABEL_BYTES`) there too.

robertcsakany:
Planned in OpenSpec change `harden-server-request-surfaces` (`openspec/changes/harden-server-request-surfaces/`), bundled with the other open server-guard follow-ups (#647, #665, #514, and the #637 enforce-default flip).

PR #666, opened 2026-09-14
feat(mcp-legacy-clients-and-token-issuance): dual-era /mcp + direct device-token issuance

robertcsakany:
Implements OpenSpec change `mcp-legacy-clients-and-token-issuance`.

## What

- **Dual-era `/mcp`** (`packages/mcp-server-plugin`): legacy revisions `2025-03-26` / `2025-06-18` / `2025-11-25` negotiate via `initialize` + an opaque, unrecorded `Mcp-Session-Id`; `notifications/*` → `202` empty, `ping` → `{}`, `subscriptions/listen` refused (streaming stays `2026-07-28`-only). Modern era unchanged. Version resolved once per request in `routes.ts` (single resolution site, also closes the pre-existing gap where a `listen` was never version-checked); `dispatchRpc` receives the resolved `{era, version}`. Unknown `initialize` versions negotiate down to `2025-11-25`; repeated/comma-joined version headers → `400 AmbiguousHeader`; no version marker at all → legacy `2025-03-26`. `mcp.json` provisioning stays pinned to `2026-07-28`.
- **Direct device-token issuance** (`packages/server`): `POST /api/paired-devices` mints a device bearer for MCP clients — operator-only (`operatorGuard`: login session via the new additive `authVia` marker ∨ `X-Pi-Local-Token` ∨ genuinely local, AND unconditional Host admission in enforce semantics — never a device bearer, never a trusted-network-only address). Registry rows carry `source: "pairing" | "manual"` (legacy rows read as `"pairing"`); label validated 1..64 UTF-8 bytes; plaintext-once.
- **Settings UI** (`packages/client`): Paired Devices gains "Create token for an MCP client" — label → one-shot token + copy-ready `claude mcp add --transport http … --header "Authorization: Bearer …"` snippet, dismiss clears the token; `manual` badge on minted rows.
- **Docs**: FAQ entry (Claude Code / Cursor connection), `docs/architecture.md` §MCP endpoint dual-era contract, plugin README.

## Verification

- 19,815 unit/integration tests green (`npm test`); one pre-existing flake retried green (DiagnosticsSection clipboard-fallback timing).
- All deterministic enforcers green: `check-conventions`, `dox-byte-gate`, `i18n-lint --strict`, `i18n-parity`, `knip-config`, `knip-ratchet` (+ baseline-diff), `theme-token-guard`.
- Two-round local review gate (`@review`) passed after fixes.
- Docker-harness e2e: new `tests/e2e/mcp-token-settings.spec.ts` (F1/F2/F3 — mint → snippet → dismiss-unrecoverable → revoke → bearer dies on `/mcp`) plus `pairing-qr` and `mcp-client-settings-a11y` regression specs, all green against the harness built from this branch.

## Follow-up

- `DELETE /api/paired-devices/:id` still admits a device bearer via `networkGuard` (plus the unbounded `approve` label) — filed as #665; the new `authVia` marker makes it a one-line tightening.

## Note

The branch carries the `openspec/changes/paginate-mcp-list-sessions/` planning artifacts (planning-only, no code) from a separately-planned OpenSpec change that predates this branch; they ride along in the squash and are tracked independently.


robertcsakany (commented):
Disputing this finding — it re-litigates a deliberate, spec-bound design decision, and its proposed fix contradicts the change's own specification.

**Why the guard intentionally admits `isGenuinelyLocal`:**

1. **D5 (openspec/changes/archive/2026-09-14-mcp-legacy-clients-and-token-issuance/design.md)** explicitly adjudicates this: the mint route admits a genuinely local caller (loopback, no proxy-forwarding headers) in ANY auth mode, because `auth-plugin.ts` already exempts a genuinely-local request from login BEFORE it looks at the cookie — so the loopback browser with OAuth enabled holds no cookie (X7: "Local browser mints with authentication enabled ... no cookie → 200"). Requiring `authVia === "session"` would make that spec scenario unsatisfiable and break the Settings flow for every auth-on local user, which is the primary consumer (D5: "`X-Pi-Local-Token` is NOT made mandatory: the browser cannot present it, and the Settings flow is the primary consumer").

2. **The threat model of an "unmarked reverse tunnel" that satisfies `isGenuinelyLocal`**: such a tunnel must terminate ON the dashboard host and present a loopback peer with no forwarding headers. That requires a process running on the same host — which already holds the local token and the filesystem (including `~/.pi/dashboard/paired-devices.json`, 0600, hash-only). Design D5's risk table records this exact boundary: "A same-host process already holds the local token and the filesystem; no new trust boundary." An `ssh -R` on the dashboard host minting a token is strictly weaker than what that foothold already has.

3. **`isGenuinelyLocal` is not spoofable remotely**: it requires a loopback peer AND zero `x-forwarded-*` headers (`packages/server/src/auth/localhost-guard.ts`), so a DNS-rebinding page or a tunnel-presenting-as-loopback with markers is refused — and the Host-admission gate runs in enforce semantics on this route regardless of the global gate mode (X6, tested). A rebinding attacker's `Host` is never admitted.

4. The threat is real only for a marker-less reverse tunnel whose operator lacks any other host credential — accepted and documented in D5 ("CLI callers use curl ... or the local token"), the same posture as every other genuinely-local route (e.g. `/api/pair/payload`).

No code change: this is the designed trust boundary, pinned by X4/X5/X6/X7 and the e2e. If the maintainers want to close the ssh -R case anyway, it belongs to the Phase-C hardening follow-up (#665), not a drive-by change to this spec-bound decision.

PR #670, opened 2026-09-15
fix(server,client): server-derived host-pressure verdict — kill the false unresponsive badge

robertcsakany:
Implements OpenSpec change `fix-false-unresponsive-badge`.

## The bug

Every live session card rendered `unresponsive · ~24m` while every bridge was healthy.

The client derived the host-pressure verdict from `session.processMetrics.updatedAt` — a field that reaches the browser exactly once, in the connect `sessions_snapshot`, and is then never refreshed. It froze at page-load time, so roughly a minute later every live session read as unresponsive.

## The fix

The **server** owns the last-frame fact, so it derives the verdict (`packages/server/src/session/host-pressure-tracker.ts`) and pushes it on a **state transition only** — a healthy session still costs zero frames, a pressured one costs two (in, and out). The card renders the server's verdict and only ticks the elapsed silence locally.

### Transport, lifecycle and drift (this PR's second half)

- **Shed frames are repaid.** A transition-only signal has no successor frame, so a shed recovery would leave the badge lit until a reconnect. `flushStatusDebt` now rebuilds `hostPressure` from the live row, with the same load-bearing `?? null` clearing semantics as `currentTool`.
- **An OPEN bridge socket is a precondition of the signal.** Carrier loss belongs to the heartbeat/status machinery; reporting it as host pressure would double-badge a disconnect. A verdict already raised on a half-open socket is **retracted**, not merely forgotten.
- **Every session-exit path releases the tracker entry and its timers** — explicit unregister, all three heartbeat/grace/sleep-retry expiries, the ping-timeout reaper, `closeSession()`, and the `/reload` placeholder swap. `hostPressureTrackedCount()` is the leak oracle.
- **An ended session serves no stale verdict** — both `→ ended` seams clear the row, and the field is transient (absent from `sessionToMeta`, so a restart cannot resurrect it).
- **One source of truth for the thresholds and the shape** (`packages/shared/src/host-pressure.ts`) — the card escalates *between* server transitions, so it must use the very numbers the server fires on.

## Testing

43 L1 tests (tracker boundaries, gateway wiring, backpressure reconcile, the 8-cell client decision table, the cost promise, and the five error-handling scenarios), plus two new L3 Playwright scenarios in `tests/e2e/host-pressure-badge.spec.ts` (live raise + recovery with no reload; snapshot parity for a second browser context), verified against the docker harness.

Each new behaviour was proven non-vacuous by toggling the implementation off and observing the specific tests fail.

Task 7.7 (badge legibility across the four themes) is manual-only per `test-plan.md` and deferred to post-merge visual verification.


robertcsakany:
Triage of this round, for the record:

**Applied (3, in this change's scope)**
- `pi-gateway.ts:566` — ownership gate on `noteFrame`. Applied, mirroring the close handler's existing guard. **No test ships for it**: displacement is `held.terminate()`, so the window in which a non-owning socket can still deliver a frame is not reachable through the public socket path. A test written for it passed with the guard **removed**, so it was deleted rather than shipped as false evidence. The guard stays as documented defense-in-depth.
- `pi-gateway.ts:1254` — `closeSession` now retracts on the `clear()` boolean, exactly as suggested.
- `memory-session-manager.ts:427` — agreed on the gap, but routed differently to avoid a layering inversion: the manager stays pure (no tracker reference). `PiGateway.clearHostPressure` is wired to `sessionManager.onEnded` in `event-wiring.ts`, which fires on the exact `→ ended` transition from **both** seams. Covered by `X8` (proven non-vacuous).
- `tasks.md` / `test-plan.md` prose — synchronized; they still claimed sections 6-8 were unwritten while the checklist marked them done.

**Out of scope (3)**
The findings on `openspec/changes/heal-orphaned-tool-cards-on-session-end/**` are spec-quality feedback on a **different, not-yet-implemented** change. Its planning artifacts ride along in this PR only because they were committed to the branch point; they are untouched by this change. They belong to that change's own review, not to this one, and are deliberately left for it.

robertcsakany:
@coderabbitai full review

PR #671, opened 2026-09-15
feat(heal-orphaned-tool-cards-on-session-end): terminate orphaned tool cards + subagents on session end

robertcsakany:
Implements OpenSpec change `heal-orphaned-tool-cards-on-session-end`.

A session that dies mid-run (kill -9, watchdog force-close, grace expiry) left every in-flight tool card and subagent spinning forever — live **and** after a reload. The server now synthesizes the missing terminal events into the session's own stream on `sessionManager.onEnded` (both end seams), and the transcript parser closes an orphaned `toolCall` with the same error shape, so live and replay agree by construction.

**What changed**
- `packages/server/src/session/open-tool-calls.ts` (new) — pure `findOpenToolCalls` / `findOpenSubagents` + the two synthesizers.
- `packages/server/src/event-wiring.ts` — heal on `onEnded`, before `session_updated{ended}`; idempotent via the store; skipped for a relocation (`session.movedTo`); insert-not-broadcast while replaying.
- `packages/shared/src/state-replay.ts` — orphan-close now `isError:true`, `result:"parent session ended"`, `healedBy:"session_ended"` (was a silent empty success).
- `packages/client/src/lib/chat/event-reducer.ts` — the `running`-only guard generalizes to any `healedBy`; the Agent backfill and the `subagent_completed`/`failed` arm refuse to regress a terminal subagent; a healed end merges into existing `toolDetails`.
- `packages/shared/src/config.ts` + `packages/server/src/config-api.ts` + `cli.ts` — `subagentTickThrottleMs` default `0 → 500` with a marker-gated one-shot boot migration (boot path only, never `loadConfig`).
- `packages/shared/src/pi-package-resolver.ts` — memoize `npm root -g` (measured: 7-way child fan-out loop block 5046 ms → 54 ms, 24 spawns → 1).

**Verification**
- Every test-plan scenario folded into suites: `open-tool-calls.test.ts` (E1–E7, P1), `session-end-orphan-heal.test.ts` (X1–X5), `state-replay.test.ts` (X6/X7), `event-reducer.session-ended-heal.test.ts` (F1–F5), `throttle-default-migration.test.ts` (E8–E10, X8), `pi-package-resolver.test.ts` (E11–E14).
- L3 `tests/e2e/session-ended-orphan-heal.spec.ts` (#F6 live heal, #F7 cold re-hydration) — green twice against the docker harness; `default-throttle P3` green on the `PI_SYNTH_AGENT_TICKS=1` arm.
- Local review: isolated `@review` subagent, 2 rounds, ending with zero blocking findings.


robertcsakany:
@coderabbitai full review

robertcsakany:
Thanks — 4 of the 9 applied in 693ba41; the rest are deferred with reasons.

**Applied**
- `event-reducer.ts` — the healed merge now stamps `status: isError ? "error" : "completed"` into the preserved snapshot, and the test asserts it starting from a `status:"running"` snapshot. Good catch: without it the details contradicted `toolStatus`.
- `pi-package-resolver.ts.AGENTS.md` — stale "Read-on-call (no cache)" claim replaced.
- `tests/e2e/*` docs — the rows now describe the implemented scenario (long-running bash call, `force_kill` for #F6 vs raw `kill -9` for #F7, subagent half covered at L1 `#X1`).

**Deferred, with reasons**
- **`open-tool-calls.ts` synthetic `agent_start` boundary** — real and already documented in `design.md` (D2) + the module header. Dropping the turn scope would let a start whose end never reached the store (transport/back-pressure gap) be "healed" with a false error, which is the worse failure; persisting a reconnect discriminator is a bridge-protocol change outside this change's scope. The D7 transcript orphan-close is not turn-scoped, so the next cold hydration still heals such a call.
- **`session-end-orphan-heal.test.ts:146` — broadcast `session_updated{ended}` on the `update()` seam** — that seam's broadcast behaviour is pre-existing (each `update({status:"ended"})` call site broadcasts its own, e.g. `handleForceKill`); routing it through `onEnded` is a lifecycle refactor, not part of this heal. The test now documents the asymmetry explicitly rather than asserting a vacuous ordering.
- **`state-replay.ts` incomplete retained transcripts** — accurate observation, but pre-existing in kind: before this change the same truncated `toolCall` replayed as a *silent empty success*, which is no more truthful. Threading completeness through `replayEntriesAsEvents` touches every transcript consumer and belongs in its own change; the retained read already reports `state: "incomplete"` to its callers.
- **`pi-gateway.ts` / `server.ts` host-pressure findings (3)** — not from this change. They belong to commit 6208203 (`server-derived host-pressure verdict`), which rides along on this branch; they are being tracked separately.

PR #677, opened 2026-09-16
openspec: plan close-registry-frame-shed-gaps, harden-server-request-surfaces, fix-terminals-action-opens-terminal

robertcsakany:
## What

Lands OpenSpec **planning artifacts** for three changes that are not yet implemented:

| Change | Artifacts |
|---|---|
| `close-registry-frame-shed-gaps` | proposal · design · tasks · test-plan · 2 spec deltas |
| `harden-server-request-surfaces` | proposal · design · tasks · test-plan · 4 spec deltas |
| `fix-terminals-action-opens-terminal` | proposal · design · tasks · test-plan · 1 spec delta |

**This PR is 100% markdown — zero source changes.** No behaviour ships. Each change is implemented later in its own worktree.

> Branch naming note: the branch is `os/paginate-mcp-list-sessions`, but that change was already implemented and archived separately (#674). This branch only ever carried the planning work above.

## Review-gate fixes included

A local `@review` pass found three blocking defects in the artifacts; all are fixed here:

1. **`ws-frame-delivery-policy` spec self-contradiction.** The scenario *"Add then remove within one window converges to removed"* demanded `only session_removed` for the same state (`sawAdd` true, record ended) where the requirement text, the sibling *"created and ended inside one flood window"* scenario, and test-plan **E7** — which uses the **same id `s3`** — all require `session_added` + `reconciled:true`. Deleted; coverage preserved by the sibling scenario and E7.

2. **`harden-server-request-surfaces` risk mitigation inverted its own security decision.** The Risks line said `gh` sites *"route through `buildSafeArgv`"* — the opposite of D3, which forbids it because `buildSafeArgv` re-interposes `cmd.exe` and re-exposes `& | % ^`. An implementer following the risk line would have reintroduced the exact injection surface the change removes. Realigned with D3's own *Residual* (ENOENT accepted as fail-closed).

3. **Unresolvable exemplar paths.** `tasks.md`/`test-plan.md` cited `packages/client/src/components/__tests__/useMessageHandler.*`; the files live under `hooks/__tests__/`. A literal `apply` would have seeded nine orphan test files away from the hook suite. All 17 cited paths now resolve on disk.

## Verification

- `check-conventions.mjs --base origin/develop` — clean
- `dox-byte-gate.mjs` · `i18n-lint.mjs --strict` · `i18n-parity.mjs` · `knip-config.mjs` — clean
- `knip-ratchet.mjs --check-baseline-diff origin/develop` — no baseline class raised
- `@review` round 2 — all three blockers RESOLVED, no new blocking findings

### Known pre-existing (not caused here)

`knip-ratchet.mjs` reports `types 194 > baseline 193 (+1)`. This branch's source tree is **byte-identical to `develop`**, so the drift is inherited; `--check-baseline-diff` confirms no ceiling was raised. Left untouched rather than folding an unrelated dead-code fix into a planning PR.

### Non-blocking review findings, deliberately not fixed

Reported for a follow-up rather than widening this PR: coverage-summary arithmetic drift in all three test-plans (level splits disagree with the manifest rows; disposition counts are correct), a stale task cross-ref in `close-registry` design D2, a 16-vs-17 site-count drift in `harden`, and an untested `host-admission` boot-log clause.


robertcsakany:
@coderabbitai review

robertcsakany:
All 9 findings addressed in 69fadf2b4.

**Accepted and fixed (9/9).** Verified each against source before editing rather than taking the finding on trust.

**Design gaps**

- **Debt not cleared on delivery** — confirmed. Added the clear-on-delivery rule to the requirement (explicitly *not* discarding debt recorded during the send), plus the ordering scenario you asked for, test-plan `E21`, and task `3.15b`.
- **`loadConfig()` fallback paths (CWE-16)** — confirmed, and this was the most serious one. `loadConfig` has **three** early returns of `DEFAULTS` (`!existsSync` → missing file, `!raw.trim()` → empty, `catch` → malformed) that never reach `parseHostGateMode`, and `DEFAULTS.hostGate` is `{mode:"report"}`. D4 asserted flipping it was "an inert no-op" — false. A **fresh install with no config.json**, the most common deployment, would have stayed report-only. D4, task 4.12, the `shared-config` delta, test-plan `E33b` and task `4.2b` now flip it; the malformed-JSON path fails closed deliberately.
- **Non-string `label`** — confirmed contradiction. Resolved toward the spec: supplied non-string → `400`; only a genuinely absent key keeps the pending label. Updated design D2, task 2.16, `E24`, and made the `qr-device-pairing` delta state it explicitly.
- **Duplicate-PTY goal vs race** — confirmed. Narrowed the goal to orderings the readiness gate can observe and named the accepted residual (bounded to one extra terminal per entry); closing it needs server-side create idempotency, out of scope here.

**Counts** — I recounted the manifest rows programmatically rather than trusting either reviewer:

| change | level line was | now | rows |
|---|---|---|---|
| close-registry | L1 29 · L3 4 | L1 32 · L3 3 | 36 |
| harden | L1 50 | L1 53 | 56 |
| fix-terminals | L1 14 | L1 15 | 21 |

The `class` and `disposition` lines were already correct in all three; only the level lines were wrong. Totals now reconcile exactly (the new E21/E33b rows are included above). Also corrected the 16→17 site count in tasks 3.12 + proposal Impact, and the `status: null` clause (status always carries the current status; only `currentTool`/`hostPressure` are nullable).

One finding I did **not** apply as stated: the `proposal.md:12` "already-ended `session_added` path" item — the four-path summary is a restatement of detail the requirement already pins, and the proposal deliberately stays at summary level. Happy to add it if you feel the summary misleads.

`openspec validate --strict` passes for all three changes.

robertcsakany:
Pushed `a3e1b1615` for the new finding.

**`tasks.md:32` — Include task 3.15b in the closeout verification range: accepted, fixed.** Good catch, and it was a defect I introduced: adding `3.15b` left it outside 3.16's `Verify 3.1–3.15 green` range, so the new E21 test could have stayed unverified. Range is now `3.1–3.15b`.

The other three open threads are the **original** review comments (17:14) that were already addressed in `69fadf2b4`; they just aren't auto-resolved. For the record:

- **`ws-frame-delivery-policy/spec.md` — clear debt on delivery:** fixed. The requirement now carries the clear-on-delivery rule (and explicitly does not discard debt recorded during the send), with the ordering scenario "A delivered removal clears an older shed-add debt", test-plan `E21`, and tasks `3.15b` + the `3.16` implement text.
- **`harden/proposal.md` — non-string label:** fixed. Supplied non-string → `400` in design D2, task 2.16, `E24`, and the `qr-device-pairing` delta now states it normatively with its own scenario.
- **`close-registry/proposal.md` — already-ended `session_added` path:** the one I'm still declining, as noted above. The proposal's What Changes is a summary; the four-path dispatch is pinned normatively in the requirement and exercised by `E7`. Happy to expand it if you think the summary misleads a reader.

`openspec validate --strict` green on all three changes; `check-conventions` clean.

PR #709, opened 2026-09-19
Redesign: Providers settings page

robertcsakany:
# Redesign: Providers settings page

Branch `os/redesign-providers-settings-page` → `develop`. Single commit `1bea9a063`, 95 files (56 `packages/`, 30 `openspec/`, 9 `tests/e2e/`).

OpenSpec change archived at `openspec/changes/archive/2026-09-19-redesign-providers-settings-page/`; 8 specs updated, 2 new capability specs created (`custom-provider-crud`, `provider-add-flow`).

## Gate results (local)

| Gate | Result |
|---|---|
| Enforcers (conventions, dox-gate, i18n lint+parity, knip config+ratchet+baseline-diff) | All green post-commit |
| Code review | PASS — 2 rounds. Found a 50%-flaky test (`oauthCred()` called twice against a `Date.now()` default); fixed, 0/10 failures after |
| Full suite | 21,358 pass, 6 failures — **0 attributable to this branch** (4 reproduce at stashed HEAD baseline, 1 green post-commit, 1 docker-contention flake passing 9/9 standalone) |
| E2E (L3, docker harness) | 14/14 scenarios green; teardown verified, no containers left |
| Archive sync | Succeeded; specs verified |

### Planning-artifact conformance fixes made during archive (mechanical, no design change)
1. **Renamed requirement** — delta MODIFY'd `Test button on the custom-endpoint surface`, which doesn't exist in the live spec (`…on Add Provider card`). Fixed with a `## RENAMED Requirements` FROM/TO block. `validate --strict` does not catch this; only `archive` does.
2. **Preserved a live scenario the archive would have dropped** — `Save is independent of Test`, adapted to the redesign (asserts the Save Bar must not open) rather than resurrecting the removed provider dirty-state.
3. **Task 10.5 manifest tag** — fold used bold `(**F14**)` instead of `(test-plan: manual-only)`, so `deferDecision` misclassified the manual-only leftover as a blocker. Appended the conforming tag.

## Reported, not fixed (deliberate — out of scope for this PR)

| ID | Finding |
|---|---|
| **F2** | Remove on a `models_json_key` / `runtime` row toasts success while removing nothing |
| **F3** | Empty `baseUrl` is unvalidated in the dialog |
| **F4** | Auth-code refusals surface only in the callback tab |
| **F5–F7** | Nits (see review thread) |

Plus: **conditional `test.skip`** in `tests/e2e/settings-provider-auth-corrupt-authfile.spec.ts` — CI sets `PI_E2E_SEED` but not `PI_TEST_PEERS`, so this spec skips in CI. Tracked, not addressed here.

## Expected CI noise (pre-existing, not introduced here)

`git-worktree-lifecycle-ops`, `session-diff-source` (7.7), `display-fit-perf` X1, `browser-skill-registered` — timing/tarball-sensitive; the repo documents that timing tests are CPU-starved on saturated boxes. If CI is red on exactly these, they are pre-existing and should be reported separately rather than fixed inside this PR.

---
Unrelated, surfaced by the push: GitHub reports 125 pre-existing Dependabot vulnerabilities on the default branch.

## Pre-existing CodeQL alert (not introduced here)

CodeQL reports **1 new critical alert** — `js/request-forgery` (SSRF) at
`packages/server/src/package/provider-probe.ts:266`
(alert [#19](https://github.com/BlackBeltTechnology/pi-agent-dashboard/security/code-scanning/19)).

Verified **not attributable to this PR**:
- `provider-probe.ts` is **not among this PR's 95 changed files**
- `git diff origin/develop...HEAD -- packages/server/src/package/provider-probe.ts` → **empty**
- The flagged `fetch(req.url, …)` sits at line 266 on `origin/develop` too — identical

It surfaces as "new" only because the rebase moved the merge base, so CodeQL re-scanned unchanged code. On the merits it is the custom-provider **Test** button fetching a user-supplied `baseUrl` — the feature's intent. Outbound-request containment is scoped separately under the `add-universal-network-guard` change.


robertcsakany:
## Merging with two red checks — both pre-existing, neither from this branch

### 1. `ci` → `pnpm test` — pre-existing flaky tests

Three runs on this head produced three DIFFERENT timing failures, none touching provider code:

| Run | Failure |
|---|---|
| 1 | `goal-plugin` **E24** — `assignSessionRef` observed `s1` before `s2` (passed on run 2) |
| 2 | `server-real-process` `keeper.test.ts` **E5** — `waitFor timed out after 60000ms` (passed on run 3) |
| 3 | `goal-plugin` **X6** — `pumpUntil` never satisfied |

X6 also fails on **develop itself**:

- develop @ `b6d2a2aa2` (this PR's merge-base) — run 35418883508 — FAIL X6
- develop @ `e342fe292` — run 35466019257 — FAIL X6

`packages/goal-plugin/**` is not in this PR's diff, and `plugin-action-handler.test.ts` passes 8/8 locally across repeated runs.

### 2. `CodeQL` → `js/request-forgery` at `provider-probe.ts:266`

Covered in the PR body: the file is byte-identical to develop and absent from this PR's changed files. It surfaces as "new" only because the rebase moved the merge base.

### Green on this head

`CI-level publish scenarios` (the AGENTS byte-cap gate fixed in `860157f85`), lint, typecheck, Biome, CodeQL actions/python/javascript-typescript, and **CodeRabbit (review completed)**.

All 10 CodeRabbit findings were addressed across `c01759152` and `f8d6067ca` — including the Major silent-data-loss bug (PATCH returning 200 while persisting nothing on a non-object `providers`) and the 0600 credential-file mode. The one finding left unchanged targets the archived proposal's summary bullet; the normative live spec already states the correct rule.

